PR #6, opened 2025-05-02
TEST: Test `propagate_map_queries` pass

mtsokol:
Hi @willow-ahrens,

I migrated `propagate_map_queries` test from my ["draft autoscheduler" PR](https://github.com/pydata/sparse/pull/831).

To make sure the test passes I had to adjust `def make_term` for `Plan`, `Query`, and `MapJoin`.

My impression is that `x.make_term(x.head(), ...)` kind of duplicates information - `x.head()` is more less a `cls` of `x`, right?

I propose:
```python
def make_term(cls, *args):
    return cls(*args)
```
instead and override it (e.g. `Plan`) to make passing arguments more convenient.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ readme = "README.md"
 packages = [{include = "finch", from = "src"}]
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = "^3.11"
 numpy = ">=1.19"
 
 [tool.poetry.group.test.dependencies]
--- a/src/finch/autoschedule/__init__.py
+++ b/src/finch/autoschedule/__init__.py
@@ -1,4 +1,4 @@
-from .finch_logic import (
+from ..finch_logic import (
     Aggregate,
     Alias,
     Deferred,
@@ -15,7 +15,7 @@ from .finch_logic import (
     Table,
 )
 from .optimize import optimize, propagate_map_queries
-from .rewrite_tools import PostOrderDFS, PostWalk, PreWalk
+from ..symbolic import PostOrderDFS, PostWalk, PreWalk
 
 __all__ = [
     "Aggregate",
--- a/src/finch/autoschedule/compiler.py
+++ b/src/finch/autoschedule/compiler.py
@@ -2,7 +2,7 @@ from collections.abc import Hashable
 from textwrap import dedent
 from typing import Any
 
-from .finch_logic import (
+from ..finch_logic import (
     Alias,
     Deferred,
     Field,
--- a/src/finch/autoschedule/executor.py
+++ b/src/finch/autoschedule/executor.py
@@ -1,5 +1,5 @@
 from .compiler import LogicCompiler
-from .rewrite_tools import gensym
+from ..symbolic import gensym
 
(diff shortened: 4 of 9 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 packages = [{include = "finch", from = "src"}]
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = "^3.11"
```
mtsokol, line 10: We use `obj[*idxs]` which was introduced in `3.11` so I think we need to drop `3.10`.

willow-ahrens:
Let's leave make_term as-is. I'd like to be able to change the representation of the term under the hood, so that if we need to make the head an integer/enum later we still can. Could we instead change the tests here?

pyproject.toml
willow-ahrens: I see, yes I changed this in CI but not in poetry

willow-ahrens:
Unfortunately, I don't think we can keep the overrides of make_term for Plan and MapJoin. The make_term function should be a class attribute, and not vary instance to instance. The make_term function is meant for generic traversals over the IR, so if it has different calling conventions for different terms, it defeats the purpose. Make_term is meant to be called when you don't know what the head of the term is yet. You can override direct constructors for terms if you like, but I'd advocate against it, since it's more clear to pattern-match against dataclasses with the same syntax we use to construct them.

willow-ahrens (changes requested):
Okay, I understand what's going on with make_term.

It should hold for all terms that
`x == x.make_term(x.head(), *x.children())`

Currently, this is not true for any of the terms that hold tuples of arguments.

You needed to override make_term to make this work for Plan and Mapjoin since those are the terms that got tested which hold tuples of arguments.

Could you override make_term so that the above invariant holds for all terms?

mtsokol:
The whole problem originates from the fact that the only thing missing from Python `dataclasses` is variable arguments for the constructor. Therefore number of arguments must be always known at `class` definition level. This requires us to keep variable length arguments as a tuple for a few Logic nodes:
```python
class Plan(LogicNode):
    bodies: tuple[LogicNode] = ()
```
and I think this is sufficient for our case, it's just something to keep in mind when rewriting Julia code.
But the current `def head()` implementation is broken due to this reason:
```python
    @classmethod
    def make_term(cls, head, *args):
        return head(*args)
```
When constructing a plan.make_term(plan.head(), plan.arguments()) the `*args` causes the tuple to be applied as multiple parameters instead of one `bodies` tuple. Therefore I overridden those `make_term` that have iterable as an argument and it doesn't break OOP principles - it makes it compatible with tree-like nodes.

mtsokol:
> It should hold for all terms that `x == x.make_term(x.head(), *x.children())`

The current overrides for tree-like nodes (or expr-like nodes here) make sure that this is valid, e.g.:

```python
class MapJoin(LogicNode):
    @classmethod
    def make_term(cls, head, op, *args):
        return head(op, args)
```

I can add a test for this!

willow-ahrens:
Thanks!

PR #7, opened 2025-05-05
ENH: `lift_subqueries` scheduler pass

mtsokol:
This PR:
- introduces `lift_subqueries` default scheduler pass
- adds `pre-commit` setup
- lints existing files according to `ruff` rules

### DIFF
--- a/src/finch/finch_logic/nodes.py
+++ b/src/finch/finch_logic/nodes.py
@@ -1,20 +1,27 @@
 from abc import abstractmethod
-from collections.abc import Iterable
 from dataclasses import dataclass
 from typing import Any
 from ..symbolic import Term
+
 
 @dataclass(eq=True, frozen=True)
 class LogicNode(Term):
     """
     LogicNode
 
-    Represents a Finch Logic IR node. Finch uses a variant of Concrete Field Notation 
+    Represents a Finch Logic IR node. Finch uses a variant of Concrete Field Notation
     as an intermediate representation.
 
-    The LogicNode struct represents many different Finch IR nodes. The nodes are 
+    The LogicNode struct represents many different Finch IR nodes. The nodes are
     differentiated by a `FinchLogic.LogicNodeKind` enum.
     """
+
+    @staticmethod
+    @abstractmethod
+    def is_expr():
+        """Determines if the node is expresion."""
+        ...
+
     @staticmethod
     @abstractmethod
     def is_stateful():
@@ -25,6 +32,10 @@ class LogicNode(Term):
     def head(cls):
         """Returns the head of the node."""
         return cls
+
+    def children(self):
+        """Returns the children of the node."""
+        raise Exception(f"`children` isn't supported for {self.__class__}.")
 
     @classmethod
     def make_term(cls, head, *args):
@@ -40,6 +51,7 @@ class Immediate(LogicNode):
     Attributes:
         val: The literal value.
     """
+
     val: Any
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


src/finch/finch_logic/nodes.py
```diff
@@ -26,6 +33,10 @@ def head(cls):
         """Returns the head of the node."""
         return cls
 
+    def children(self):
+        """Returns the children of the node."""
+        return []
```
mtsokol, line 38: in Finch.jl `arguments(node)` can be also called on non-tree IR instances. I think this should be implemented the same way here: Return empty list by default and `is_expr=True` IRs override it returning actual children.

willow-ahrens: I have mixed thoughts on this. Implementing `children(node)` for literal feels dangerous because `x.make_term(x.head(), x.children())` is not equal to `x`. What do you think Mateusz? @wraith1995 do you have thoughts on this one? Maybe as long as we don't also implement make_term for this IR node type it's okay.

src/finch/finch_logic/nodes.py
```diff
@@ -144,8 +159,9 @@ class Table(LogicNode):
         tns: The tensor object.
         idxs: The fields indexing the tensor.
     """
+
     tns: LogicNode
-    idxs: Iterable[LogicNode]
+    idxs: tuple[LogicNode]
```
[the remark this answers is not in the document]
willow-ahrens: Good call!

src/finch/finch_logic/nodes.py
mtsokol: I'm also hesitant about this, so I followed Finch.jl:
```python
julia> using Finch

julia> sq = Finch.subquery("lhs", "arg")
Finch Logic: (lhs = arg)

julia> imm = Finch.immediate("imm")
Finch Logic: imm

julia> Finch.arguments(sq)
2-element Vector{Finch.FinchLogic.LogicNode}:
 Finch Logic: lhs
 Finch Logic: arg

julia> Finch.arguments(imm)
Finch.FinchLogic.LogicNode[]
```

mtsokol: Now I would prefer to have it as:

```python
    def children(self):
        """Returns the children of the node."""
        raise Exception(f"`children` isn't supported for {self.__class__}")
```
as a base implementation.

wraith1995: I have two thoughts: 

1. Perhaps more thought should be given to how the language interacts with things outside of it - in an ideal world, nothing outside the IR except for constants would be in the IR or some process would remove such references and replace them with something else. Perhaps, this would fix the problem of `make_term`? If you have tensors in the expressions, perhaps they should be pre-processed out before anything else happens.
2. I generally support that IR instances should have a uniform set of methods so I don't think either should not implement anything. But perhaps returning [] is a bad idea - a pythonic way to do this would be to return None and handle the consequences downstream. Alternatively, you could have make_term fail on expressions by raising an exception. Either way, it requires code in various places to catch this.

willow-ahrens: I agree with 1, and I think that from the point of the view of the IR, everything that isn't `is_expr` is basically different kinds of constants. Mateusz, let's go forward with your Error-related approach for now. I think Teo's return-none approach is similar to `if x.is_expr() x.children else None`, which would need an if-branch anyway.

src/finch/finch_logic/nodes.py
```diff
@@ -314,15 +335,16 @@ def children(self):
...
     and returns `rhs`.
 
     Attributes:
         lhs: The left-hand side of the binding.
         rhs: The argument to evaluate.
     """
+
     lhs: LogicNode
-    rhs: LogicNode
+    arg: LogicNode
```
mtsokol, line 347: Renamed it to follow Finch.jl nodes.

src/finch/finch_logic/nodes.py
mtsokol: 👍 I made it an Exception then!

willow-ahrens (approved):
LGTM!

PR #10, opened 2025-05-08
eager methods/function

### LINKED ISSUE #1: Define eager ops
Following the lead of [Finch.jl eager](https://github.com/finch-tensor/Finch.jl/blob/main/src/interface/eager.jl) and [Finch.jl lazy](https://github.com/finch-tensor/Finch.jl/blob/main/src/interface/lazy.jl) operations, we need to define eager versions of the lazy operations, which can almost always be done using the pattern: eager_op(tensor) = compute(lazy_op(lazy(tensor)))

https://github.com/finch-tensor/finch-tensor-lite/blob/main/src/finch/interface/lazy.py

a starter file is given in https://github.com/finch-tensor/finch-tensor-lite/blob/6541c7f597ed2e176713949c429224eef8f2345b/src/finch/interface/tensor.py#L31-L32

Note that namespacing is a little tricky here, we must be careful to define `prod` in terms of `lazy.prod`.

onopre:
Implemented a few eager methods that use lazy tensors. 

Also added few functions that operate based on the type of tensor - lazy or eager.

towards #1.


src/finch/interface/tensor.py
```diff
@@ -20,15 +20,283 @@ def to_numpy(self):
 
     @abstractmethod
     def __add__(self, other):
-        compute(lazy.lazy(self).__add__(other))
+        lazy.compute(lazy.lazy(self).__add__(other))
```
willow-ahrens, line 23: For dunder binary ops (functions that take the "other" argument), we should return a lazy array if other is lazy. This allows lazytensor to propagate through the computation.

willow-ahrens (changes requested):
In addition to the comments, please add tests for 1 or 2 methods which have lazy implementations.

src/finch/interface/tensor.py
```diff
@@ -20,15 +20,283 @@ def to_numpy(self):
...
+    
+    def __abs__(self):
+        """Define absolute value for tensors."""
+        lazy.compute(lazy.lazy(self).__abs__())
+
+    def __invert__(self):
+        """Define bitwise NOT for tensors."""
+        lazy.compute(lazy.lazy(self).__invert__())
+
+    #TODO: should these raise an error? There are few others like that
```
hameerabbasi, line 232: These should succeed only for 0-dimensional arrays and/or scalars, provided they can be safely coerced. i.e. if the array is floating point, only `__float__` and `__complex__` should succeed.

See the type promotion rules here: https://data-apis.org/array-api/latest/API_specification/type_promotion.html

hameerabbasi: See also, e.g. https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__int__.html#array_api.array.__int__

willow-ahrens (changes requested):
This looks good to me. I'm wondering if we should even implement __complex__ in this class. I'll add some tests and make a separate issue for those methods.

willow-ahrens (approved):

willow-ahrens:
just needs to pass tests now

willow-ahrens:
closed in favor of https://github.com/finch-tensor/finch-tensor-lite/pull/24

PR #15, opened 2025-05-09
Requesting review:

siyi-hu:
## Updates
- pre-added all test cases for all interfaces
- finished the draft version of: sum, all, any
- doing tests on: prod, sum, all, any
- currently working on: min, max, mean, std, var, argmin, argmax
- trying to finish all interfaces within the week 5/11

## Question
Current pytest output on prod looks like:
```
***** prod *****
=== Finch ===
LazyTensor(data=Subquery(lhs=Alias(name='#A#41'), arg=Aggregate(op=Immediate(val=<built-in function mul>), init=Immediate(val=np.int64(0)), arg=Relabel(arg=Subquery(lhs=Alias(name='#A#38'), arg=Table(tns=Immediate(val=array([2, 3, 4])), idxs=[Field(name='#i#39')])), idxs=[Field(name='#i#40')]), idxs=[Field(name='#i#40')])), shape=(), fill_value=np.int64(0), element_type=<class 'numpy.int64'>)

=== Numpy ===
24
```

reduce function returns a lazeTensor instead of a scalar value. Am I missing some steps?


src/finch/interface/lazy.py
```diff
@@ -327,7 +327,64 @@ def prod(arr: LazyTensor, dims) -> LazyTensor:
...
-    return reduce(operator.mul, arr, dims, arr.fill_value)
+    return reduce(operator.mul, arr, axis=dims, init=arr.fill_value)
 
 def multiply(x1: LazyTensor, x2: LazyTensor) -> LazyTensor:
-    return elementwise(operator.mul, x1, x2)
\ No newline at end of file
+    return elementwise(operator.mul, x1, x2)
+
+
+def sum(arr: LazyTensor, dims) -> LazyTensor:
```
willow-ahrens, line 336: Does not conform to the function signature for sum: https://data-apis.org/array-api/2021.12/API_specification/generated/array_api.sum.html

tests/test_interface.py
```diff
@@ -16,4 +17,92 @@ def test_matrix_multiplication(a, b):
...
+    "a",
+    [
+        (np.array([2, 3, 4])),
+    ],
+)
+def test_vector_prod(a):
+    print("***** prod *****")
+    
+    la = finch.lazy(a)
+    result = finch.prod(la, -1)
```
willow-ahrens, line 33: you must call `compute` on the resulting lazy tensor to get an actual tensor result

willow-ahrens (changes requested):
A few fixes required

willow-ahrens:
Hi siyi, I've left a short review. Thanks for your work on this! In the future, notice the upper right-hand section of the page where it says "reviewers". This "reviewers" section allows you to request a review from me or any other collaborator.

siyi-hu:
* Thanks for the review comments and I've just committed current state to this PR
* Several TODOs are attached in the sources to indicate my next steps and also for questions

src/finch/interface/lazy.py
```diff
@@ -1,5 +1,7 @@
 import operator
 import builtins
+# TODO: Is numpy allowed within lazy tensor interface?
+import numpy as np
```
willow-ahrens, line 4: numpy is allowed everywhere, I don't see where you're using it though.

src/finch/interface/lazy.py
```diff
@@ -327,7 +337,173 @@ def prod(arr: LazyTensor, dims) -> LazyTensor:
...
+        /,
+        *,
+        axis: int | tuple[int, ...] | None = None,
+        dtype = None,
+        keepdims: bool = False
+    ) -> LazyTensor:
+    """
+    Calculates the sum of input array ``arr`` elements.
+    """
+    return reduce(operator.add, arr, axis=axis, keepdims=keepdims, init=arr.fill_value)
```
willow-ahrens, line 356: You should add `init=None` in the function signature, and pass the value of init to `reduce`, which will query the `initial_value` function in `algebra.py` to figure out what the initial value should be. Do this for all the reduction functions. You'll need to add initial value properties for most of these.

src/finch/interface/lazy.py
```diff
@@ -327,7 +337,173 @@ def prod(arr: LazyTensor, dims) -> LazyTensor:
...
+
+#     # column‑wise minimum
+#     if axis == 0:
+#         return [min(arr[r][c] for r in range(num_rows)) for c in range(num_cols)]
+#     # row‑wise minimum
+
+#     if axis == 1:
+#         return [min(row) for row in arr]
+
+    # TODO: Use operator.min to reduce?
```
willow-ahrens, line 413: yes, use operator.min here. You'll need to define some properties in `algebra.py` for min.

src/finch/interface/lazy.py
```diff
@@ -327,7 +337,173 @@ def prod(arr: LazyTensor, dims) -> LazyTensor:
...
+        *,
+        axis: int | tuple[int, ...] | None = None,
+        keepdims: bool = False
+    ) -> LazyTensor:
+    """
+    Calculates the arithmetic mean of the input array ``arr``.
+    """
+    origin = np.asarray(arr.shape)
+    ele_no = origin[axis]
+    n = lazy(np.full(np.delete(origin, axis), ele_no, dtype=int))
```
willow-ahrens, line 442: I would try to do this without np. I think you can just use a generator and python base functions.

src/finch/interface/lazy.py
```diff
@@ -327,7 +337,173 @@ def prod(arr: LazyTensor, dims) -> LazyTensor:
...
+
+    m = mean(arr, axis=axis, keepdims=keepdims)
+    # TODO: Confirm whether we have broadcasting or reshape interface for lazyTensor
+    # v = elementwise(operator.sub, arr, m)
+    # v2 = elementwise(operator.mul, v, v)
+    # return elementwise(operator.truediv, sum(v, axis=axis, keepdims=keepdims), n)
+
+    return reduce(operator.add, arr, axis=axis, keepdims=keepdims, init=arr.fill_value)
```
willow-ahrens, line 469: could be sum

src/finch/interface/lazy.py
```diff
@@ -327,7 +337,173 @@ def prod(arr: LazyTensor, dims) -> LazyTensor:
...
+        arr: LazyTensor,
+        /,
+        *,
+        axis: int | tuple[int, ...] | None = None,
+        keepdims: bool = False
+    ) -> LazyTensor:
+    """
+    Returns the indices of the minimum values along a specified axis.
+    """
+    return reduce(operator.add, arr, axis=axis, keepdims=keepdims, init=arr.fill_value)
```
willow-ahrens, line 497: see the way we implement in julia: https://github.com/finch-tensor/Finch.jl/blob/c287ea7a920124e4b953aa13ba88a5310e43c1f1/src/interface/lazy.jl#L776-L809

use`range` or `np.ndindex` or `np.ndenumerate` to get lazy indices in python.

src/finch/interface/tensor.py
```diff
@@ -32,3 +32,63 @@ def prod(arr, /, axis=None):
         return lazy.prod(arr, axis=axis)
     else:
         return compute(lazy.prod(lazy.lazy(arr), axis=axis))
+
```
willow-ahrens, line 35: don't worry about this one for now, on thursday we'll have a better model for eager arrays.

tests/test_reduce_api_funcs.py
```diff
@@ -0,0 +1,215 @@
...
+    "a",
+    [
+        (np.array([[2, 3, 4], [5, 6, 7]])),
+    ],
+)
+def test_lazyTensor_prod(a):
+    output_term("")
+    output_term("***** prod *****")
+
+    # TODO: axis=None is not supported well in all interfaces?
```
willow-ahrens, line 26: fix it if it's not, I think this was just an absentminded oversight on my part

tests/test_reduce_api_funcs.py
```diff
@@ -0,0 +1,215 @@
...
+    # TODO: results computed from operator.mul are all 0?
```
willow-ahrens, line 27: oops, maybe the initial value for mul is wrong, see above comments about init

tests/test_reduce_api_funcs.py
```diff
@@ -0,0 +1,215 @@
...
+    # assert_equal(result, expected)
+    output_term("")
+
+    result = finch.compute(finch.prod(finch.lazy(a), axis=1)); output_term(result)
+    expected = np.prod(a, axis=1); output_term(expected)
+    # assert_equal(result, expected)
+    output_term("")
+
+
+@pytest.mark.parametrize(
```
willow-ahrens, line 47: IN general, you can probably group tests into three groups: float arrays, int arrays, and bool arrays, where each group tests all applicable methods for each.

willow-ahrens (changes requested):
great start! thanks for your hard work on this!

willow-ahrens:
There are only two files in the PR now, have you overwritten your changes?

siyi-hu:
Yes, I reverted the changes in lazy.py and __init__.py to the PR#24 version in order to let git solve all the conflicts.
I'll submit the newest modifications on the next push. Also, tensor.py has been deleted so next revision will be 4 files instead of 5 changed I think.

siyi-hu:
Poetry has been updated to 2.0 and looks like I got following errors running with the new version. Any configuration needed?

```
~/repos/finch-tensor-lite$ poetry install

The Poetry configuration is invalid:
  - The fields ['authors', 'description', 'name', 'version'] are required in package mode.
  - Additional properties are not allowed ('requires-poetry' was unexpected)

~/repos/finch-tensor-lite$ poetry --version
Poetry (version 1.8.2)
```

mtsokol:
@siyi-hu Can you try `poetry self update`? There's `Poetry (version 1.8.2)` in what you pasted. After that you should be able to run it again, e.g. `poetry run pytest`.

siyi-hu:
@mtsokol Thanks! ```poetry self update``` should work fine on updating poetry up to 2.0. 
It seems I got environment error when using pip command due to my local ubuntu server 24.04 LTS configurations. 

```
EnvCommandError

  Command ['/usr/bin/python3', '-m', 'pip', 'uninstall', 'certifi', '-y'] errored with the following return code 1
  
  Output:
  error: externally-managed-environment
  ......
```

Any quick advice would be helpful if you've experienced the procedure on fixing externally-managed-environment. I think I'll fix this soon so no worries.

hameerabbasi:
> 1. Any way to indicate the namespace here in elementwise?

You can do `builtins.any` (along with `import builtins`) to get the Python default `any`.

willow-ahrens:
note when implementing `lazy.any`: [`builtins.any(a)`](https://docs.python.org/3/library/functions.html#any) would need to be implemented as `any(a) = reduce(operator.and, map(operator.truth, a))`. You actually should not implement it as `any(a) = reduce(and_test, a)` because `any` returns `True` or `False`, whereas `a and b` returns `b` if `a` is truthy.

src/finch/algebra/operator.py
```diff
@@ -0,0 +1,17 @@
+import operator
+
+
+def and_test(a, b) -> bool:
+    return operator.truth(a) and operator.truth(b)
```
willow-ahrens, line 5: The point of `and_test` was supposed to be the [prefix](https://www.cs.colostate.edu/~cs165/.Fall19/recitations/L15/doc/traversal-order.html) version of `and`, which is [infix](https://www.cs.colostate.edu/~cs165/.Fall19/recitations/L15/doc/traversal-order.html), similar to [Lib/operator.py](https://docs.python.org/3/library/operator.html). Please read all links in this comment completely, let me know if you have questions.

`operator` defines prefix versions of many infix python calls. `operator.and_` is bitwise, (c's `&`).

You should implement `all` using `reduce(operator.and_, map(operator.truth, x))`. Implement `and_test` to be equivalent to `and` (drop `truth` here), and do not use `and_test` in `all`.

src/finch/algebra/operator.py
```diff
@@ -0,0 +1,17 @@
...
+
+def or_test(a, b) -> bool:
+    return operator.truth(a) or operator.truth(b)
+
+
+def not_test(a) -> bool:
+    return not operator.truth(a)
+
+
+# def ifelse:
```
willow-ahrens, line 16: ifelse(c, a, b) = a if c else b

tests/test_reduce_api_funcs.py
willow-ahrens: Group the tests into three groups: bitwise reductions, integer reductions, and float reductions. Use `@parameterize` to iterate over a list of the applicable ops for each group.

willow-ahrens (changes requested):
This is much closer, though I feel perhaps I may have made things complicated talking about `and_test`. you can implement `and_test`, `or_test`, `not_test`, but do not use them for `all` or `any`. Please read the comments on the review and all of the linked documents.

src/finch/algebra/algebra.py
```diff
@@ -210,6 +211,21 @@ def _return_type_closure(a, b):
...
+    min,
+    "__call__",
+    "return_type",
+    lambda op, a, b: query_property(a, "__add__", "return_type", b),
+)
+register_property(
+    max,
+    "__call__",
+    "return_type",
+    lambda op, a, b: query_property(a, "__add__", "return_type", b),
```
willow-ahrens, line 224: why not just add min and max to the list of reflexive operators?

src/finch/interface/lazy.py
```diff
@@ -393,3 +393,79 @@ def positive(x) -> LazyTensor:
...
+    *,
+    axis: int | tuple[int, ...] | None = None,
+    keepdims: bool = False,
+    init=None,
+):
+    """
+    Test whether all elements of input array ``x`` along given axis are True.
+    """
+    if init is None:
+        init = init_value(builtins.all, x)
```
willow-ahrens, line 432: I think you mean to use `operator.and_`

src/finch/interface/lazy.py
```diff
@@ -393,3 +393,79 @@ def positive(x) -> LazyTensor:
...
+    *,
+    axis: int | tuple[int, ...] | None = None,
+    keepdims: bool = False,
+    init=None,
+):
+    """
+    Test whether any element of input array ``x`` along given axis is True.
+    """
+    if init is None:
+        init = init_value(builtins.any, x)
```
willow-ahrens, line 410: I think you mean to use `operator.or_`. Does the array api spec allow an `init` for this function?

src/finch/interface/lazy.py
```diff
@@ -393,3 +393,79 @@ def positive(x) -> LazyTensor:
...
+    *,
+    axis: int | tuple[int, ...] | None = None,
+    keepdims: bool = False,
+    init=None,
+):
+    """
+    Return the minimum of input array ``arr`` along given axis.
+    """
+    if init is None:
+        init = init_value(builtins.min, x)
```
willow-ahrens, line 454: Do you need to pass init here if we already call init_value in reduce?

src/finch/interface/lazy.py
```diff
@@ -393,3 +393,79 @@ def positive(x) -> LazyTensor:
...
+    *,
+    axis: int | tuple[int, ...] | None = None,
+    keepdims: bool = False,
+    init=None,
+):
+    """
+    Return the maximum of input array ``arr`` along given axis.
+    """
+    if init is None:
+        init = init_value(builtins.max, x)
```
willow-ahrens, line 470: Do you need to pass init here if we already call init_value in reduce?

willow-ahrens (changes requested):
Looking better! Smaller comments this time

src/finch/interface/lazy.py
willow-ahrens: This is not resolved, the reduce function already defines an init. just pass init unmodified.

src/finch/interface/lazy.py
willow-ahrens: This is not resolved, the reduce function already defines an init. just pass init unmodified.

src/finch/interface/lazy.py
willow-ahrens: This is not resolved, the reduce function already defines an init. just pass init unmodified.

src/finch/interface/lazy.py
willow-ahrens: This is not resolved, the reduce function already defines an init. just pass init unmodified.

src/finch/algebra/algebra.py
willow-ahrens: Defining `return_type` of max differently than the actual return type of max is incorrect.

willow-ahrens (changes requested):
Okay, things look good, just one final request:

You defined the return type of `max` differently than the way it actually behaves. `max(True, 0) = True`, but `return_type(max, Bool, Int)` is returning `Int`. Please define a new function in `operator.py` called `promote_max`, which promotes numeric types before maxing them. Roughly, the implementation can be:

```
def promote_max(a, b):
    c = return_type(operator.add, a, b)
    max(c(a), c(b))
```

Then, your definition of the return type of `promote_max` can be the same as you have done already, in terms of `__add__`.

You can then define max as `reduce(promote_max, ...)`.

Do the same for min.

willow-ahrens (approved):
I finished up the implementation of promote_type, and added a few new properties to make things work correctly. I think I had something pretty specific in mind. Thanks for your hard work!

PR #17, opened 2025-05-12
Linting fixes

hameerabbasi:
- **Add pre-commit hook for `mypy` to catch typing errors**
- **Linting fixes for entire library.**

### DIFF
--- a/src/finch/autoschedule/compiler.py
+++ b/src/finch/autoschedule/compiler.py
@@ -22,11 +22,7 @@ T = TypeVar("T", bound="LogicNode")
 
 
 def get_or_insert(dictionary: dict[str, T], key: str, default: T) -> T:
-    val = dictionary.get(key)
-    if val is not None:
-        return val
-    dictionary[key] = default
-    return default
+    return dictionary.setdefault(key, default)
 
 
 def get_structure(
--- a/src/finch/autoschedule/compiler.py
+++ b/src/finch/autoschedule/compiler.py
@@ -1,6 +1,7 @@
 from collections.abc import Hashable
 from textwrap import dedent
-from typing import Any
+from typing import Any, TypeVar
+from ..symbolic import Term
 
 from ..finch_logic import (
     Alias,
@@ -17,8 +18,9 @@ from ..finch_logic import (
     Table,
 )
 
+T = TypeVar('T', bound=Term)
 
-def get_or_insert(dictionary: dict[Hashable, Any], key: Hashable, default: Any) -> Any:
+def get_or_insert(dictionary: dict[str, Term], key: str, default: Any) -> Any:
     if key in dictionary:
         return dictionary[key]
     dictionary[key] = default
@@ -26,8 +28,8 @@ def get_or_insert(dictionary: dict[Hashable, Any], key: Hashable, default: Any) 
 
 
 def get_structure(
-    node: LogicNode, fields: dict[str, LogicNode], aliases: dict[str, LogicNode]
-) -> LogicNode:
+    node: T, fields: dict[str, Term], aliases: dict[str, Term]
+) -> T:
     match node:
         case Field(name):
             return get_or_insert(fields, name, Immediate(len(fields) + len(aliases)))
@@ -35,19 +37,20 @@ def get_structure(
             return get_or_insert(aliases, name, Immediate(len(fields) + len(aliases)))
         case Subquery(Alias(name) as lhs, arg):
             if name in aliases:
-                return aliases[name]
-            return Subquery(
+                return aliases[name]  # type: ignore[return-value]
+            return Subquery(  # type: ignore[return-value]
                 get_structure(lhs, fields, aliases), get_structure(arg, fields, aliases)
             )
         case Table(tns, idxs):
-            return Table(
+            return Table(  # type: ignore[return-value]
                 Immediate(type(tns.val)),
                 tuple(get_structure(idx, fields, aliases) for idx in idxs),
             )
-        case any if any.is_tree():
-            return any.from_arguments(
-                *[get_structure(arg, fields, aliases) for arg in any.get_arguments()]
-            )
+        # TODO: `is_tree` isn't defined
+        # case any if any.is_tree():
+        #     return any.from_arguments(
+        #         *[get_structure(arg, fields, aliases) for arg in any.get_arguments()]
+        #     )
         case _:
             return node
 
@@ -81,7 +84,7 @@ def compile_pointwise_logic(ex: LogicNode) -> tuple:
     return (code, ctx.bound_idxs)
 
 
-def compile_logic_constant(ex: LogicNode) -> str:
+def compile_logic_constant(ex: Term) -> str:
     match ex:
         case Immediate(val):
             return val
@@ -107,7 +110,7 @@ class LogicLowerer:
     def __init__(self, mode: str = "fast"):
         self.mode = mode
 
-    def __call__(self, ex: LogicNode):
+    def __call__(self, ex: Term) -> str:
         match ex:
             case Query(Alias(name), Table(tns, _)):
                 return f":({name} = {compile_logic_constant(tns)})"
@@ -153,6 +156,6 @@ class LogicCompiler:
     def __init__(self):
         self.ll = LogicLowerer()
 
-    def __call__(self, prgm):
+    def __call__(self, prgm: Term) -> str:
         # prgm = format_queries(prgm, True)  # noqa: F821
         return self.ll(prgm)
--- a/src/finch/autoschedule/compiler.py
+++ b/src/finch/autoschedule/compiler.py
@@ -25,7 +25,9 @@ def get_or_insert(dictionary: dict[Hashable, Any], key: Hashable, default: Any) 
     return default
 
 
-def get_structure(node: LogicNode, fields: dict[str, LogicNode], aliases: dict[str, LogicNode]) -> LogicNode:
+def get_structure(
+    node: LogicNode, fields: dict[str, LogicNode], aliases: dict[str, LogicNode]
+) -> LogicNode:
     match node:
         case Field(name):
             return get_or_insert(fields, name, Immediate(len(fields) + len(aliases)))
@@ -34,11 +36,18 @@ def get_structure(node: LogicNode, fields: dict[str, LogicNode], aliases: dict[s
         case Subquery(Alias(name) as lhs, arg):
             if name in aliases:
                 return aliases[name]
-            return Subquery(get_structure(lhs, fields, aliases), get_structure(arg, fields, aliases))
+            return Subquery(
+                get_structure(lhs, fields, aliases), get_structure(arg, fields, aliases)
+            )
         case Table(tns, idxs):
-            return Table(Immediate(type(tns.val)), tuple(get_structure(idx, fields, aliases) for idx in idxs))
+            return Table(
+                Immediate(type(tns.val)),
+                tuple(get_structure(idx, fields, aliases) for idx in idxs),
+            )
         case any if any.is_tree():
-            return any.from_arguments(*[get_structure(arg, fields, aliases) for arg in any.get_arguments()])
+            return any.from_arguments(
+                *[get_structure(arg, fields, aliases) for arg in any.get_arguments()]
+            )
         case _:
             return node
 
@@ -53,7 +62,11 @@ class PointwiseLowerer:
                 return f":({val}({','.join([self(arg) for arg in args])}))"
             case Reorder(Relabel(Alias(name), idxs_1), idxs_2):
                 self.bound_idxs.append(idxs_1)
-                return f":({name}[{','.join([idx.name if idx in idxs_2 else 1 for idx in idxs_1])}])"
+                return (
+                    f":({name}"
+                    + ",".join([idx.name if idx in idxs_2 else 1 for idx in idxs_1])
+                    + ")"
+                )
             case Reorder(Immediate(val), _):
                 return val
             case Immediate(val):
@@ -99,10 +112,18 @@ class LogicLowerer:
             case Query(Alias(name), Table(tns, _)):
                 return f":({name} = {compile_logic_constant(tns)})"
 
-            case Query(Alias(_) as lhs, Reformat(tns, Reorder(Relabel(Alias(_) as arg, idxs_1), idxs_2))):
-                loop_idxs = [idx.name for idx in with_subsequence(intersect(idxs_1, idxs_2), idxs_2)]
+            case Query(
+                Alias(_) as lhs,
+                Reformat(tns, Reorder(Relabel(Alias(_) as arg, idxs_1), idxs_2)),
+            ):
+                loop_idxs = [
+                    idx.name
+                    for idx in with_subsequence(intersect(idxs_1, idxs_2), idxs_2)
+                ]
                 lhs_idxs = [idx.name for idx in idxs_2]
-                (rhs, rhs_idxs) = compile_pointwise_logic(Reorder(Relabel(arg, idxs_1), idxs_2))
+                (rhs, rhs_idxs) = compile_pointwise_logic(
+                    Reorder(Relabel(arg, idxs_1), idxs_2)
+                )
                 body = f":({lhs.name}[{','.join(lhs_idxs)}] = {rhs})"
                 for idx in loop_idxs:
                     if Field(idx) in rhs_idxs:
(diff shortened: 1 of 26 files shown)


willow-ahrens (approved):
this is great, thanks! I especially like the match case that was introduced.

mtsokol (changes requested):
Posted some comments. Let's rebase it and solve conflicts and squash to single commit - the diff should be smaller.

src/finch/finch_logic/nodes.py
```diff
@@ -33,12 +57,12 @@ def head(cls):
...
         return cls
 
-    def children(self):
+    def children(self) -> list[Term]:
         """Returns the children of the node."""
         raise Exception(f"`children` isn't supported for {self.__class__}.")
 
     @classmethod
-    def make_term(cls, head, *args):
+    def make_term(cls, head, args):
```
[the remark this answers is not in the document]
willow-ahrens: Ah, I see what happened here. The description of `make_term` in the abstract class disagreed with the usage of make_term, and Hameer chose the abstract class as the source of truth.

src/finch/finch_logic/nodes.py
```diff
@@ -226,10 +256,10 @@ class Aggregate(LogicNode):
...
     """
 
-    op: LogicNode
-    init: LogicNode
-    arg: LogicNode
-    idxs: tuple[LogicNode]
+    op: Immediate
+    init: Immediate
+    arg: Term
+    idxs: tuple[Field, ...]
```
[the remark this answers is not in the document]
willow-ahrens: this is more correct as LogicNode.

src/finch/finch_logic/nodes.py
```diff
@@ -413,24 +445,24 @@ def is_stateful():
 
     def children(self):
         """Returns the children of the node."""
-        return [*self.args]
+        return list(self.args)
```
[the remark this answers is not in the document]
willow-ahrens: agree.

src/finch/finch_logic/nodes.py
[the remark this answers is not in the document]
willow-ahrens: I think we still need it in logicnode class, because postwalk, etc. call it. Maketerm is for IR-generic transformations, transformations we can apply to all IR nodes.

hameerabbasi:
> squash to single commit

This can be done by the squash-merge strategy in GitHub. Also, squashing merge-commits is hard.

mtsokol:
> > squash to single commit
> 
> This can be done by the squash-merge strategy in GitHub. 

Sure!

> Also, squashing merge-commits is hard.

Right, that's why I personally prefer to always `git rebase -i main` over merging `main` into my branch.

willow-ahrens:
Is there a repo-wide setting I need to update here?

mtsokol:
> Is there a repo-wide setting I need to update here?

Let's make this a principle we stick to - I don't think there is.

[EDIT] Looks that here's a branch rule for it:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-linear-history

But some people complain about it: https://github.com/orgs/community/discussions/61107#discussioncomment-6671872 although I'm not sure this applies to us.

hameerabbasi:
> Right, that's why I personally prefer to always `git rebase -i main` over merging `main` into the branch.

I usually prefer that too -- but I do make exceptions for large PRs where one merge conflict is easier to handle than many small ones.

src/finch/symbolic/term.py
```diff
@@ -1,56 +1,65 @@
...
     def is_expr(self) -> bool:
-        """Return True if the term is an expression tree, False otherwise. Must implement children() if True."""
-        pass
+        """
+        Return True if the term is an expression tree, False otherwise. Must implement
+        `children()` if `True`."""
 
     @abstractmethod
-    def make_term(self, head: Any, children: list["Term"]) -> "Term":
+    def make_term(self, *children: Term) -> Self:
```
mtsokol, line 39: Let's keep it `make_term(self, head: Any, children: list[Term])`. In some previous PR we agreed we need `head` to be able to change type of node: https://github.com/finch-tensor/finch-tensor-lite/pull/6#issuecomment-2847143473

src/finch/symbolic/term.py
```diff
@@ -1,56 +1,65 @@
...
+        `x == x.make_term(*x.children())`
         """
-        pass
 
     def __hash__(self) -> int:
         """Return the hash value of the term."""
         if self._hashcache is None:
             self._hashcache = hash(
-                (0x1CA5C2ADCA744860, self.head(), tuple(self.children()))
+                (0x1CA5C2ADCA744860, type(self), tuple(self.children()))
```
mtsokol, line 50: Let's keep head.

src/finch/symbolic/term.py
```diff
@@ -1,56 +1,65 @@
...
+if TYPE_CHECKING:
+    from ..finch_logic import LogicNode
+
 
 class Term(ABC):
     def __init__(self):
         self._hashcache = None  # Private field to cache the hash value
 
     @abstractmethod
-    def head(self) -> Any:
```
mtsokol, line 19: Let's keep it.

src/finch/symbolic/term.py
```diff
@@ -1,56 +1,65 @@
...
-        """Return the head type of the S-expression."""
-        pass
-
-    def children(self) -> list["Term"]:
+    def children(self) -> list[LogicNode]:
```
mtsokol, line 29: This should return Term. Term isn't aware of LogicNode at all.

src/finch/symbolic/rewriters.py
```diff
@@ -169,9 +181,7 @@ def __call__(self, x: Term) -> Term | None:
         if y is not None:
             if y.is_expr():
                 y_args = y.children()
-                return y.make_term(
-                    y.head(), *[default_rewrite(self(arg), arg) for arg in y_args]
-                )
+                return y.make_term(*(default_rewrite(self(arg), arg) for arg in y_args))
```
mtsokol, line 184: Let's keep `head` here and in the rest of the file.

src/finch/finch_logic/nodes.py
```diff
@@ -28,27 +52,25 @@ def is_stateful():
         """Determines if the node is stateful."""
         ...
 
-    @classmethod
-    def head(cls):
-        """Returns the head of the node."""
-        return cls
```
mtsokol, line 34: Let's keep all `head` methods.

src/finch/finch_logic/nodes.py
```diff
@@ -28,27 +52,25 @@ def is_stateful():
...
-    def make_term(cls, head: type, *args: Any) -> "LogicNode":
+    def make_term(cls, *args: Term) -> Self:
         """Creates a term with the given head and arguments."""
-        return head(*args)
+        return cls(*args)
 
 
 @dataclass(eq=True, frozen=True)
-class Immediate(LogicNode):
+class WithFields(LogicNode):
```
mtsokol, line 66: If we want such abstraction then I'd advocate for a class that isn't a dataclass (there's no need for it):

```python
class WithFields:
    def get_fields(self): ...
```

and make nodes here as `class Immediate(LogicNode, WithFields)` and leave `LogicNode` as is. IMO `WithFields(LogicNode)` is confusing. We would create another level of inheritance each time we want a new interface method - `class WithAnotherMethod()`.

src/finch/autoschedule/optimize.py
```diff
@@ -31,19 +34,18 @@ def _lift_subqueries_expr(node: LogicNode, bindings: dict) -> LogicNode:
...
-                any.head(),
-                *map(lambda x: _lift_subqueries_expr(x, bindings), any.children()),
+                *tuple(_lift_subqueries_expr(x, bindings) for x in any.children()),
             )
         case _:
             return node
 
 
-def lift_subqueries(node: LogicNode) -> LogicNode:
+def lift_subqueries(node: Term) -> LogicNode:
```
mtsokol, line 43: Should stay `LogicNode` and in all other places in this file.

mtsokol (changes requested):

src/finch/finch_logic/nodes.py
hameerabbasi: Well, there are two reasons for doing it this way:

1. `WithFields` is a strict subclass of `LogicNode`; everything that has the `get_fields` method is also a `LogicNode` subclass.
2. It helps with typing: If I do `x: WithFields` it works. Python types don't have support for `+`/`&` from some short tests -- which makes writing types that are BOTH a `T` and `U` for some types `T` and `U` really hard.

To alleviate your concern, I've named the class `NodeWithFields`.

src/finch/autoschedule/optimize.py
hameerabbasi: I tried -- that makes the type checker really unhappy. I can try once more though.

hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/bdda4a3d58883b656ba7c5b11fdd0713922797ec.

src/finch/finch_logic/nodes.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/8dd493c266ea74a284e2a4642c88875ac4f107d7.

src/finch/finch_logic/nodes.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/56349b0d36760ac0f6f916d9b851a2d6744caddd.

src/finch/symbolic/rewriters.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/56349b0d36760ac0f6f916d9b851a2d6744caddd

src/finch/symbolic/term.py
hameerabbasi: Done.

src/finch/symbolic/term.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/56349b0d36760ac0f6f916d9b851a2d6744caddd.

src/finch/symbolic/term.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/56349b0d36760ac0f6f916d9b851a2d6744caddd.

src/finch/symbolic/term.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/56349b0d36760ac0f6f916d9b851a2d6744caddd.

src/finch/finch_logic/nodes.py
mtsokol: Answering to points:
1. I think the same can be achieved by having `WithFields` interface class with one method and making `class LogicNode(Term, WithFields)`, WDYT?
2. I think point 1. also solves this. You only write one type annotation `x: LogicNode`.
Basically I can imagine that we might want another method for `LogicNode` and then adding another level of `With<MethodName>` inheritance isn't desirable. 

I just think that `LogicNode` will be used **heavily** across the whole codebase. Having `...WithFields` (10 characters!) everywhere just because one additional method implemented (which is just a one liner) doesn't sound right to me (especially that we might want add more methods in the future).

I would advocate that logic node should be used under a name `LogicNode` and make it a standard here.
What do you think about `class LogicNode(Term, WithFields)`?

hameerabbasi: That's what I wanted to do originally (just putting `get_fields` on `LogicNode` itself), but it turns out the `get_fields` isn't implemented on all `LogicNode`s, but only specific ones. We can, of course, raise a `TypeError` where it isn't implemented; however in a correctly typed program there should never be `TypeError`s. Another option is to say it should actually be implemented on all `LogicNode`s, but it isn't just yet (in which case we raise a `NotImplementedError`) -- in this case, we can put `get_fields` directly on `LogicNode`. However, something tells me that the `get_fields` wouldn't always be implemented on all `LogicNode`s, which is why I chose to do it this way.

mtsokol (changes requested):
@hameerabbasi I finished reviewing it: I'm only concerned about `LogicNode` class and `node.py` changes (I still prefer the have `LogicNode` class used as a main node representation).

The rest of the changes LGMT.

src/finch/finch_logic/nodes.py
mtsokol: How about `class WithFields` defines implements `get_fields` and throws `NotImplementedError` and relevant `LogicNode` classes override it?

hameerabbasi: > How about `class WithFields` defines implements `get_fields` and throws `NotImplementedError` and relevant `LogicNode` classes override it?

It depends: Do the other classes really not implement it YET, or should they not have it at all? cc @willow-ahrens 

If the former, this is a valid approach. If the latter, then raising `NotImplementedError` is incorrect, it should really be a `TypeError`, and we're back to a typing violation.

src/finch/autoschedule/compiler.py
```diff
@@ -9,36 +8,55 @@
...
-def get_or_insert(dictionary: dict[Hashable, Any], key: Hashable, default: Any) -> Any:
-    if key in dictionary:
-        return dictionary[key]
+
+def get_or_insert(dictionary: dict[str, T], key: str, default: T) -> T:
+    val = dictionary.get(key)
+    if val is not None:
+        return val
     dictionary[key] = default
     return default
```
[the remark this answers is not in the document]
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/fefb6016ae81eefd83b609f1cd5acf782ba7ccfb.

src/finch/finch_logic/nodes.py
```diff
@@ -488,23 +432,21 @@ class Query(LogicNode):
...
     def is_stateful():
         """Determines if the node is stateful."""
         return True
 
-    def children(self):
+    def children(self) -> list[LogicNode]:
         """Returns the children of the node."""
         return [self.lhs, self.rhs]
 
+    def get_fields(self) -> list[Field]:
```
hameerabbasi, line 444: @mtsokol I'm not entirely happy about this one (and the ones below). Can this be implemented?

src/finch/finch_logic/nodes.py
hameerabbasi: Done as discussed offline in https://github.com/finch-tensor/finch-tensor-lite/pull/17/commits/36071fd0e6b2ae651d1bae4b5053f931a3f7afc0.

src/finch/finch_logic/nodes.py
mtsokol: I think it can (`Subquery` has it) but let's leave it for now, and decide in the call (at the end it isn't used anywhere probably).

src/finch/symbolic/term.py
```diff
@@ -28,15 +28,9 @@ def head(cls) -> Callable[..., Self]:
         raise NotImplementedError
 
     @abstractmethod
-    def children(self) -> list[Term]:
+    def children(self) -> list[Any]:
```
hameerabbasi, line 31: This was done as some terms, e.g. `Immediate`, `Alias`, ... return `str` or `Any`.

src/finch/finch_logic/nodes.py
hameerabbasi: If it can be implemented, then it's fine to leave it as `raise NotImplementedError` for now.

src/finch/finch_logic/nodes.py
```diff
@@ -87,23 +118,14 @@ class Deferred(LogicNode):
...
-        return False
-
     @staticmethod
     def is_stateful():
         """Determines if the node is stateful."""
         return False
 
-    def children(self):
-        """Returns the children of the node."""
-        return [self.val, self.type_]
```
mtsokol, line 102: LGTM. Last question: Don't we need it?

mtsokol (approved):

src/finch/finch_logic/nodes.py
hameerabbasi: Whoops, I was experimenting with pushing `children` down to `LogicExpression`, but turns out we need it here. Fixed!

mtsokol:
Merged, Thanks @hameerabbasi!

PR #19, opened 2025-05-15
Add finchassembly

willow-ahrens:


src/finch/finchassembly/interpreter.py
```diff
@@ -0,0 +1,142 @@
...
+class TableValue:
+    tns: Any
+    idxs: Iterable[Any]
+
+    def __post_init__(self):
+        if isinstance(self.tns, TableValue):
+            raise ValueError("The tensor (tns) cannot be a TableValue")
+
+
+class FinchLogicInterpreter:
```
hameerabbasi, line 33: This seems like a duplicate of https://github.com/finch-tensor/finch-tensor-lite/blob/c0cf9908abd0affae0d81581c54fe32f05065854/src/finch/finch_logic/interpreter.py#L36

src/finch/util/config.py
```diff
@@ -0,0 +1,68 @@
...
+
+depot_dir = os.path.realpath(os.path.expanduser(os.getenv('FINCH_PATH', os.path.join("~", ".finch"))))
+
+default_config = {
+    "FINCH_CACHE_PATH": os.path.join(depot_dir, "cache"),
+    "FINCH_CACHE_SIZE": 10000,
+    "FINCH_CACHE_ENABLE": True,
+    "FINCH_TMP": os.path.join(depot_dir, "tmp"),
+    "FINCH_LOG_PATH": os.path.join(depot_dir, "log.txt"),
+    "FINCH_CC": "gcc",
```
hameerabbasi, line 30: I'd use `os.environ.get("CC")`. If we intend to compile C++, I'd change this to `FINCH_CXX` and use `os.environ.get("CXX")`.

src/finch/util/cache.py
```diff
@@ -0,0 +1,41 @@
...
+            cache_dir = os.path.join(get_config("FINCH_CACHE_PATH"), domain)
+        else:
+            cache_dir = tempfile.mkdtemp(prefix=os.path.join(get_config("FINCH_TMP"), domain))
+            atexit.register(lambda: shutil.rmtree(cache_dir) if os.path.exists(cache_dir) else None)
+
+        if not os.path.exists(cache_dir):
+            os.makedirs(cache_dir, exist_ok=True)
+
+        def inner(*args):
+            filename = os.path.join(cache_dir, f"{f.__name__}_{uuid.uuid5(finch_uuid, str((f.__name__, args)))}.{ext}")
```
hameerabbasi, line 36: I'd use [`pathlib`](https://docs.python.org/3/library/pathlib.html) for much of the path wrangling.

src/finch/util/config.py
```diff
@@ -0,0 +1,68 @@
...
+- Configuration values are loaded automatically when the module is imported 
+    and can be accessed using the `get_config` function.
+
+Use this module to easily manage and retrieve Finch-specific settings.
+"""
+
+depot_dir = os.path.realpath(os.path.expanduser(os.getenv('FINCH_PATH', os.path.join("~", ".finch"))))
+
+default_config = {
+    "FINCH_CACHE_PATH": os.path.join(depot_dir, "cache"),
```
hameerabbasi, line 25: I'd add the package version here -- cache can easily be poisoned by using the wrong version!

src/finch/finchassembly/nodes.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    AssemblyNode
+
+    Represents a FinchAssembly IR node. FinchAssembly is the final intermediate representation
+    before code generation (translation to the output language). It is a low-level imperative
+    description of the program, with control flow, linear memory regions called "buffers", and
+    explicit memory management.
+    """
+
+    @staticmethod
```
hameerabbasi, line 18: Hmm. Maybe it makes sense to do `class Expression(Term)` and change `LogicExpression` to inherit from both `LogicNode` and `Expression`.

src/finch/finchassembly/nodes.py
```diff
@@ -0,0 +1,373 @@
...
+@dataclass(eq=True, frozen=True)
+class Block(AssemblyNode):
+    """
+    Represents a statement that executes a sequence of statements `bodies...`.
+
+    Attributes:
+        bodies: The sequence of statements to execute.
+    """
+
+    bodies: tuple[AssemblyNode] = ()
```
hameerabbasi, line 355: In general, `tuple[T]` represents a type with a single element of type `T`. If we want an arbitrary number of elements all of type `T`, we should use `tuple[T, ...]`.

Note that `list` is variable-sized by default so for that, `list[T]` suffices.

src/finch/finchassembly/nodes.py
```diff
@@ -0,0 +1,373 @@
...
+class Return(AssemblyNode):
+    """
+    Represents a return statement that returns `arg` from the current function.
+    Halts execution of the function body.
+
+    Attributes:
+        arg: The argument to return.
+    """
+
+    args: tuple[AssemblyNode]
```
hameerabbasi, line 325:
```suggestion
    args: tuple[AssemblyNode, ...]
```

src/finch/finchassembly/nodes.py
hameerabbasi:
```suggestion
    bodies: tuple[AssemblyNode, ...] = ()
```

src/finch/finchassembly/nodes.py
```diff
@@ -0,0 +1,373 @@
...
+class Call(AssemblyNode):
+    """
+    Represents an expression for calling the function `op` on `args...`.
+
+    Attributes:
+        op: The function to call.
+        args: The arguments to call on the function.
+    """
+    op: AssemblyNode
+    args: tuple[AssemblyNode]
```
hameerabbasi, line 129:
```suggestion
    args: tuple[AssemblyNode, ...]
```

hameerabbasi (changes requested):
A first pass at a review.

src/finch/finchassembly/interpreter.py
willow-ahrens: Yes, I should have deleted that when I copied the classes.

src/finch/util/config.py
willow-ahrens: I think we're doing C for now, I'll use that variable, great catch!

src/finch/util/config.py
willow-ahrens: Ah, yes. Version is one of many things we want to add to the cache. We can also use the arguments to the function as cache keys, but I think having the version in there is extra helpful, I'll add it.

src/finch/finchassembly/nodes.py
willow-ahrens: I'd like to leave this alone for now, and have a discussion about it in a separate issue. There are many very fancy ways we can automate IR definitions, but I like that the current design is really simple for now.

src/finch/util/config.py
willow-ahrens: I can't think of a case where we wouldn't want to use the version as a cache key. I'm wondering now if we want to use a hash of the finch-tensor-lite directory as a cache key.

src/finch/util/cache.py
willow-ahrens: I wasn't sure between pathlib and os. I'll switch to pathlib.

src/finch/util/cache.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Args:
+        f: The function to cache.
+        ext: The file extension for the cache file.
+        domain: The domain name for the cache file.
+
+    Returns:
+        A wrapper function that caches the result of the original function.
+    """
+    def decorator(f: Callable) -> Callable:
```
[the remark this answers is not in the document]
willow-ahrens: I made the intentional choice here to have the function responsible for producing the file. This allows us to call system gcc, or pickle, or whatever file format, serialization format, etc. we want. There should definitely be a few convenience functions for the case where we don't want to produce a file by default.

src/finch/util/config.py
```diff
@@ -0,0 +1,68 @@
...
+- Settings are stored in a `config.json` file within the `FINCH_PATH` directory.
+- Values can be set via environment variables, the `config.json` file, 
+    or the `set_config` function.
+- Configuration values are loaded automatically when the module is imported 
+    and can be accessed using the `get_config` function.
+
+Use this module to easily manage and retrieve Finch-specific settings.
+"""
+
+depot_dir = os.path.realpath(os.path.expanduser(os.getenv('FINCH_PATH', os.path.join("~", ".finch"))))
```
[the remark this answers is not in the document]
willow-ahrens: config-path is not in https://repo.anaconda.com/pkgs/main/

src/finch/util/config.py
willow-ahrens: I'm remembering my worry with CC was that sometimes compiler flags don't translate between compilers, so I thought I'd push the user to read the config if they don't want to use gcc and figure out if the flags are okay. That said, I think `-fPIC -O3` works on all linux systems. For windows, we don't need -fpic, and things are different.

src/finch/util/config.py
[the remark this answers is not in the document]
willow-ahrens: What about https://github.com/pytroll/donfig? It's in https://repo.anaconda.com/pkgs/main/ already. It's also more suited to what we're doing here. The only issue with it is it doesn't specify where I should be putting the compiled cache (which probably shouldn't go in ~, but also I'm not sure where to put it).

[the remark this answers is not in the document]
willow-ahrens: there's a really complicated one which I'm somewhat opposed to: https://github.com/alexferl/vyper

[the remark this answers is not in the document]
willow-ahrens: This one is close but doesn't make as much sense: https://github.com/rbgirshick/yacs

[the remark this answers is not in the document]
willow-ahrens: I'll make a separate issue for this. It appears that pydata/sparse does not use config files, only env vars, so this isn't a problem there.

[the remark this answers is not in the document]
willow-ahrens: Unfortunately, we need to store data in this path, so choices like User/Preferences on apple don't make as much sense. We want the finch folder location to be visible, so the user knows where to look for compiled artifacts, etc.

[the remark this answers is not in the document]
willow-ahrens: I added an issue for discussion: https://github.com/finch-tensor/finch-tensor-lite/issues/22

src/finch/util/cache.py
willow-ahrens: done

src/finch/finchassembly/nodes.py
willow-ahrens: Separate issue here: https://github.com/finch-tensor/finch-tensor-lite/issues/23

src/finch/util/config.py
willow-ahrens: Done, I skipped the directory hash but we should consider it.

PR #20, opened 2025-05-15
Revert "Linting fixes"

willow-ahrens:
Reverts finch-tensor/finch-tensor-lite#17


mtsokol:
FWIW For reintroducing it can be divided into "linting" and "features changes"

PR #21, opened 2025-05-15
Linting fixes

hameerabbasi:
* Mypy fixes
* Linting fixes
* Formatting
* `Leaf` and `Expression` types.

### DIFF
--- a/src/finch/finch_logic/nodes.py
+++ b/src/finch/finch_logic/nodes.py
@@ -1,7 +1,33 @@
+from __future__ import annotations
+
 from abc import abstractmethod
+from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any, Iterable
-from ..symbolic import Term
+from typing import Any, Never, Self, TypeVar
+
+from ..symbolic import Expression, Leaf, Term
+
+__all__ = [
+    "LogicNode",
+    "LogicExpression",
+    "Immediate",
+    "Deferred",
+    "Field",
+    "Alias",
+    "Table",
+    "MapJoin",
+    "Aggregate",
+    "Reorder",
+    "Relabel",
+    "Reformat",
+    "Subquery",
+    "Query",
+    "Produces",
+    "Plan",
+]
+
+
+T = TypeVar("T", bound="LogicNode")
 
 
 @dataclass(eq=True, frozen=True)
@@ -18,37 +44,37 @@ class LogicNode(Term):
 
     @staticmethod
     @abstractmethod
-    def is_expr():
-        """Determines if the node is expresion."""
-        ...
-
-    @staticmethod
-    @abstractmethod
     def is_stateful():
         """Determines if the node is stateful."""
-        ...
 
     @classmethod
-    def head(cls):
+    def head(cls) -> type[Self]:
         """Returns the head of the node."""
         return cls
 
-    def children(self):
-        """Returns the children of the node."""
-        raise Exception(f"`children` isn't supported for {self.__class__}.")
-
-    def get_fields(self) -> Iterable["LogicNode"]:
-        """Returns fields of the node."""
-        raise Exception(f"`fields` isn't supported for {self.__class__}.")
-
     @classmethod
-    def make_term(cls, head: type, *args: Any) -> "LogicNode":
+    def make_term(cls, head: Callable[..., Self], *args: Term) -> Self:
         """Creates a term with the given head and arguments."""
         return head(*args)
 
 
 @dataclass(eq=True, frozen=True)
-class Immediate(LogicNode):
+class LogicExpression(LogicNode, Expression):
+    @abstractmethod
+    def get_fields(self) -> list[Field]:
+        """Get this node's fields."""
+
+    @abstractmethod
+    def children(self) -> list[LogicNode]:  # type: ignore[override]
+        """Get this node's children."""
+
+
+@dataclass(eq=True, frozen=True)
+class LogicLeaf(LogicNode, Leaf): ...
+
+
+@dataclass(eq=True, frozen=True)
+class Immediate(LogicLeaf):
     """
     Represents a logical AST expression for the literal value `val`.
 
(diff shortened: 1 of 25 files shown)


hameerabbasi:
One note: You should inherit from `Expression` and `Leaf` instead of implementing `is_expr` yourself.

src/finch/finch_logic/nodes.py
```diff
@@ -451,53 +458,53 @@ class Produces(LogicNode):
...
-        return head(args)
+    def make_term(  # type: ignore[override]
+        cls, head: Callable[[tuple[LogicNode, ...]], Self], *args: LogicNode
+    ) -> Self:
+        return head(tuple(args))
 
 
 @dataclass(eq=True, frozen=True)
-class Plan(LogicNode):
+class Plan(LogicExpression):
```
willow-ahrens, line 483: There are two meanings of the word "expression" that we're working with here:
1. "S-Expression", a program node in an ast with a head and children.
2. "Expression" which is a pure computation that returns a value, as opposed to a "statement" which modifies program state.

I was assuming when you implemented get_fields for LogicExpression that you were referring to meaning 2. Now that i see a Plan is also a LogicExpression, I wonder if you should add "LogicStatement" as well, since Plan, Query, and Produces do not have fields, do not evaluate to tensors, and modify program state.

src/finch/finch_logic/nodes.py
```diff
@@ -1,7 +1,33 @@
+from __future__ import annotations
+
 from abc import abstractmethod
+from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any, Iterable
-from ..symbolic import Term
+from typing import Any, Never, Self, TypeVar
+
+from ..symbolic import Expression, Leaf, Term
```
willow-ahrens, line 8: Symbolic should stay as is, let's handle this with is_expr rather than types.

willow-ahrens: One reason to use is_expr is that we may opt for IR nodes which are all the same python class, represented by an integer enum, for example.

src/finch/interface/fuse.py
```diff
@@ -44,42 +44,55 @@
...
+    if ctx is None:
+        ctx = get_default_scheduler()
     args = [lazy(arg) for arg in args]
     if len(args) == 1:
         return f(args[0])
     return compute(f(*args), ctx=ctx)
 
-def fused(f, /, ctx=get_default_scheduler()):
+
+def fused(f, /, ctx=None):
```
willow-ahrens, line 80: I think this used to be called compiled, and I'm happy to change that back if you like.

src/finch/interface/lazy.py
```diff
@@ -1,79 +1,124 @@
...
+]
+
+
 @dataclass
 class LazyTensor:
     data: LogicNode
-    shape: Tuple
-    fill_value: Any
+    shape: tuple[int, ...]
+    fill_value: Immediate
```
willow-ahrens, line 56: The fill value is inconvenient as an Immediate, then we need to call .val on it every time we want to use it.

src/finch/symbolic/rewriters.py
```diff
@@ -58,24 +76,25 @@ class PreWalk:
...
-            return y
-        if x.is_expr():
-            args = x.children()
-            new_args = list(map(self, args))
-            if not all(arg is None for arg in new_args):
-                return x.make_term(
-                    x.head(),
-                    *map(lambda x1, x2: default_rewrite(x1, x2), new_args, args),
+        match y:
+            case LogicExpression() as expr:
```
willow-ahrens, line 84: Again, let's avoid making this change. Rewriters are intended to work on more than just LogicExpr. let's leave this as it was.

src/finch/symbolic/rewriters.py
```diff
@@ -141,15 +164,12 @@ class Fixpoint:
...
-            while y is not None and x != y:
-                x = y
-                y = self.rw(x)
-            return x
-        else:
-            return None
+        while y is not None and x != y:
+            x = y
+            y = self.rw(x)
+        return y
```
willow-ahrens, line 172: Can we move this back to the old code here? It's hard to do a PR over many files that makes changes to the logic like this. I think this new form may be wrong, but even if it's not, this was only supposed to be a linting pr.

src/finch/symbolic/rewriters.py
```diff
@@ -164,16 +184,17 @@ class Prestep:
...
+        from ..finch_logic import LogicExpression
+
         y = self.rw(x)
-        if y is not None:
-            if y.is_expr():
-                y_args = y.children()
-                return y.make_term(
-                    y.head(), *[default_rewrite(self(arg), arg) for arg in y_args]
+        match y:
+            case LogicExpression() as expr:
```
willow-ahrens, line 192: I think this could go back to the old form as well

willow-ahrens (changes requested):
Hi Hameer, thank you for your contribution, I appreciate the effort in adding a linter, and a lot of these typing and style changes are a great improvement.

1. In general, when we make big PRs that change style and touch every file in the repo, let's keep them style-only, and make logic changes separately. There's a few places where I'm not sure if the new logic is correct (perhaps it was ported from julia, but isn't currently tested, and should be left alone for now until we have tests on it and can verify that the new logic is equivalent)

2. I tried to comment several places where we use is_expr that I think should be reverted. In addition to these, I'd like the changes to term.py and rewriters.py to be entirely style and typing only in this PR.

3. Similarly, I appreciate your changes to the algebra.py lookup function, it's more correct now. I wonder if it should happen in a separate PR where there's more room to comment on it.

4. Is there a style guide you'd like to adopt for this project?

src/finch/symbolic/rewriters.py
willow-ahrens: In particular, logicnodes should import symbolic, not the other way around.

mtsokol:
I'd also advocate to have one PR for ruff linting and formatting, another one for mypy fixes feature changes. This will make it easier to review.

hameerabbasi:
I'll split this PR into two or three separate ones to make things easier to review and reason about.

hameerabbasi:
The first of the PRs is here: #30

Issue #23, opened 2025-05-15
is_expr vs Expression abstract class


willow-ahrens:
In general, it seems that is_expr is an underappreciated design decision. Over the course of Finch.jl, the design of the IR nodes has changed rather dramatically, but the is_expr design always worked. I kept is_expr in the event that we might one day wish to design an IR where the different ir node types are not different python classes, and are instead something more like a c int or enum. Consider the current design of Finch's IR (https://github.com/finch-tensor/Finch.jl/blob/c287ea7a920124e4b953aa13ba88a5310e43c1f1/src/FinchLogic/nodes.jl#L136-L141), where `LogicNodeKind` is an enum. An optimization I'd like to pursue in the future would be to cache all distinct IR nodes (to only allocate them once and compare them with pointer equality), etc. 

If we want to use Expression (I'd prefer ExprTerm actually) as an optional interface for the children and make_term method, let's make the argument that the above two optimizations would be either achievable or not needed. Let's also make the transition to the new style as a separate PR.

willow-ahrens:
On reflection, I think that defining `ExprTerm` and `LeafTerm` would likely lead to nicer code (as we see in Hameer's introduction of match-cases to handle what used to be `is_expr`).

This was the original design in Finch.jl too. For now, I don't think the single-type optimization in https://github.com/finch-tensor/Finch.jl/blob/c287ea7a920124e4b953aa13ba88a5310e43c1f1/src/FinchLogic/nodes.jl#L136-L141 will matter compared to the cost of python function calls. I think the IR node caching is still possible with this approach, so we can go forward with this.

mtsokol:
👍 for `is_tree` rename (this is how Finch.jl names it I think).

Issue #4, opened 2025-04-30
Implement elementwise ops


willow-ahrens:
This includes ` __add__` and friends (https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types and https://docs.python.org/3/reference/datamodel.html#object.__lt__), but also https://data-apis.org/array-api/latest/API_specification/elementwise_functions.html.

We need to define these in https://github.com/finch-tensor/finch-tensor-lite/blob/main/src/finch/interface/lazy.py and https://github.com/finch-tensor/finch-tensor-lite/blob/main/src/finch/interface/eager.py.

We'll also need to define some algebraic properties in https://github.com/finch-tensor/finch-tensor-lite/blob/main/src/finch/algebra/algebra.py to make this work

joaosferreira:
From the Array API, we seem to be missing as well:

- `floor_divide`
- `maximum`
- `minimum`

joaosferreira:
> I think those are aliases of max, min, and fld?

So just do something like `maximum = max`?

Issue #3, opened 2025-04-29
Expand lazy ops to include special reductions


willow-ahrens:
Implement sum, prod, min, max, mean, std, var, argmin, argmax, all, any.

These functions can be defined in terms of functions already implemented, as is done in https://github.com/finch-tensor/Finch.jl/blob/main/src/interface/lazy.jl

The definitions would be similar to https://github.com/finch-tensor/finch-tensor-lite/blob/0aeaa30517692b6a78f6a940af5f844460a91eec/src/finch/interface/lazy.py#L143.

Please include tests and docs

PR #24, opened 2025-05-18
add eager ops

willow-ahrens:
fixes https://github.com/finch-tensor/finch-tensor-lite/issues/1.

I had to make some substantial changes to the structure of the code.

1. Renames `lazy` to `defer`. Since the `finch.interface.eager` and `finch.interface.lazy` namespaces are important to this PR, and we export a `lazy` function from the `lazy` module, it becomes ambiguous which name is being referred to. I wasn't sure whether it was more disruptive to rename the `lazy` function or module, and here I chose the function.
2. Implements the `__array_function__` and `__array_ufunc__` functions for both `LazyTensor` and `EagerTensor`. I'm pretty sure this is required to make our approach work with numpy.
3. Uplifts `fill_value` and `element_type` to be overloadable properties for new classes. The `algebra.py` module is only for when we haven't defined the class ourself.


src/finch/algebra/algebra.py
```diff
@@ -122,9 +130,18 @@ def element_type(arg: Any) -> Type:
...
+    if hasattr(arg, "element_type"):
+        return arg.element_type
+    return query_property(arg, "__self__", "element_type")
+
+
+register_property(
+    np.ndarray,
+    "__self__",
+    "element_type",
+    lambda x: type(np.zeros((), dtype=x.dtype)[()]),
```
hameerabbasi, line 142:
```suggestion
    lambda x: x.dtype.type,
```

src/finch/interface/eager.py
```diff
@@ -0,0 +1,178 @@
...
+
+def expand_dims(
+    x,
+    /,
+    axis: int | tuple[int, ...] = 0,
+):
+    if isinstance(x, lazy.LazyTensor):
+        return lazy.expand_dims(x, axis=axis)
+    else:
+        return compute(lazy.expand_dims(defer(x), axis=axis))
```
hameerabbasi, line 91: A lot of these can be just decorators.

src/finch/interface/eager.py
```diff
@@ -0,0 +1,178 @@
...
+            np.abs: abs,
+            np.positive: positive,
+            np.negative: negative,
+            np.expand_dims: expand_dims,
+            np.squeeze: squeeze,
+            np.prod: prod,
+        }
+        if func in function_map:
+            return function_map[func](*args, **kwargs)
+        return NotImplemented
```
hameerabbasi, line 53:
```suggestion
        if not {LazyTensor}.issuperset(types):
            return NotImplemented
        func = globals().get(func.__name__)
        if func is None:
            return NotImplemented
        return func(*args, **kwargs)
```

src/finch/interface/eager.py
```diff
@@ -0,0 +1,178 @@
...
+            np.subtract: subtract,
+            np.multiply: multiply,
+            np.negative: negative,
+            np.positive: positive,
+            np.absolute: abs,
+            np.abs: abs,
+            # Add more ufuncs as needed
+        }
+        if ufunc in ufunc_map:
+            return ufunc_map[ufunc](*inputs, **kwargs)
```
hameerabbasi, line 72: Similar to above. We'll eventually need to handle `reduce` as well as `accumulate`.

hameerabbasi (changes requested):
A bunch of suggestions.

hameerabbasi:
While having this work with `np.*` is nice, the modern way to do this is [`__array_namespace__`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__array_namespace__.html#array_api.array.__array_namespace__) and friends [[1]](https://data-apis.org/array-api/latest/API_specification/inspection.html#inspection) [[2]](https://data-apis.org/array-api/latest/API_specification/version.html).

willow-ahrens:
> While having this work with `np.*` is nice, the modern way to do this is [`__array_namespace__`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__array_namespace__.html#array_api.array.__array_namespace__) and friends [[1]](https://data-apis.org/array-api/latest/API_specification/inspection.html#inspection) [[2]](https://data-apis.org/array-api/latest/API_specification/version.html).

If we define this, will numpy call __radd__ faithfully as expected, or do we also need __array_ufuncs__ etc.?

src/finch/interface/eager.py
willow-ahrens: I did remember your suggestion on this, but I had a thought: While people are still learning about this system and trying to contribute, I wonder if it's clearer to leave this as-is. I think it will be easier for new python programmers to add code for #2, #3, #4, #23, #25, and #26 if we do metaprogramming-based automation as a separate PR later on in the project.

Instead of decorators, maybe there's an even more concise way to dynamically add methods to the module from a list of function names?

I made an issue to discuss here: https://github.com/finch-tensor/finch-tensor-lite/issues/27

I'd like to merge this PR without metaprogramming first so that people can get started on #2, #3, #4, #23, #25, and #26

hameerabbasi:
> > While having this work with `np.*` is nice, the modern way to do this is [`__array_namespace__`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__array_namespace__.html#array_api.array.__array_namespace__) and friends [[1]](https://data-apis.org/array-api/latest/API_specification/inspection.html#inspection) [[2]](https://data-apis.org/array-api/latest/API_specification/version.html).
> 
> If we define this, will numpy call **radd** faithfully as expected, or do we also need **array_ufuncs** etc.?

`__radd__` is a Python feature -- for `a + b`, if `a.__add__(b)` returns `NotImplemented`, `b.__radd__(a)` is tried next, and if it also returns `NotImplemented`, a `TypeError` is raised. However, any errors raised are propagated.

So if `some_np_ndarray.__add__(some_finch_array)` returns `NotImplemented` (as I expect it would), it'll fall back to `some_finch_array.__radd__(some_np_ndarray)`.

Defining `__array_namespace__` and its companions have no effect on this. In `sparse`, we've done all of the above, so NumPy functions directly work on `sparse` arrays, but also using array API standard works.

willow-ahrens:
They do have an effect on this. By default, when we run `np_array + unrecognized_array`, Numpy does not use `__radd__`, it instead begins executing its own implementation of addition, adding each element of `np_array` to the entire `unrecognized_array`.

willow-ahrens:
I suppose the concise way to state what I just said is that numpy does not raise NotImplemented on unrecognized arguments to `__add__`, it treats them as scalars

hameerabbasi:
> I suppose the concise way to state what I just said is that numpy does not raise NotImplemented on unrecognized arguments to `__add__`, it treats them as scalars

Then you need `__array_ufunc__` IIUC. `__array_function__` won't help there, that's mostly so `np.*` functions work.

willow-ahrens:
do we also want to define __array_function__ though?

src/finch/interface/eager.py
hameerabbasi: How about leaving the top-level module empty? The decorator can have a very simple interface:

```
from ..deferred import __all__

# in `eager.py`
for function_name in deferred.__all__:
  globals()[function_name] = some_decorator(getattr(deferred, function_name))
```

I can work on adding the decorator myself; it shouldn't be particularly difficult. Then one only needs to work on the deferred functions and gets the eager ones "for free".

src/finch/interface/eager.py
willow-ahrens: These are handled differently, under the `__reduce__` method, right?

src/finch/interface/eager.py
willow-ahrens: yes, this makes sense, and I think it will be a nice addition, and I think we can easily explain to new python programmers that they need only define the lazy version. Can we merge this PR first as an agreement of the functionality of what eager should do, then after that you can go forward with a separate PR that automates eager.py?

willow-ahrens:
I think I understand what's needed here, I'll make a few tweaks and re-request review in an hour or two.

src/finch/interface/eager.py
willow-ahrens: also, a small note, I still like the name "lazy" for the module, rather than "deferred". It's just the function that was renamed here.

hameerabbasi: Agreed!

src/finch/interface/eager.py
hameerabbasi: `np.add.reduce` is roughly the same as `sum`, but with different defaults. `np.add.accumulate` is the cumulative sum.

We don't need to support *every single one*, but it'd be nice to unify the interface.

src/finch/interface/overrides.py
```diff
@@ -0,0 +1,59 @@
...
+        ufunc_map = {
+            np.add: operator.add,
+            np.subtract: operator.sub,
+            np.multiply: operator.mul,
+            np.negative: operator.neg,
+            np.positive: operator.pos,
+            np.absolute: operator.abs,
+            np.abs: operator.abs,
+            # Add more ufuncs as needed
+        }
```
hameerabbasi, line 32: This will be constructed each time. Let's move it out into the module and make the variable name all caps to represent a constant.

hameerabbasi (approved):
Last comment -- otherwise LGTM.

Issue #22, opened 2025-05-15
Where to put finch data and configs?


willow-ahrens:
My current approach is to create one folder, `.finch` in the home directory, and put everything in there. config.json, cached compiled objects, temporary build objects, etc. There are of course many ways to do this across different os, etc.

While https://pypi.org/project/config-path/ supports config file paths, it does not give data paths, and the output directory is not visible, but rather hidden in e.g. /User/.Library/Preferences/com.github.finch-tensor.finch/config.json

https://github.com/pytroll/donfig also supports config file paths and automates some of the env var overrides, but similarly does not specify data paths. I like donfig a little better because it chooses a location in which I would look, but I actually like the simplicity of rolling our own, very basic, approach better than depending on another system, and again, it doesn't offer an approach for data.

Is it okay to write to FINCH_PATH, which defaults to an os-specific `~/.finch`, an alright approach? I like that I can explain it easily, and I always know where to look.

willow-ahrens:
Okay, here's my thinking on this:

We currently need two paths for different purposes:
1. We need a list of paths to a config file. One of the paths should be venv-specific, one of the paths should be the repo itself, one of the paths should be whatever [XDG](https://specifications.freedesktop.org/basedir-spec/latest/) does (the "correct" location), and one of the paths should be a very obvious place in the home directory, like `~/.config/finch.json` or `~/.finch/config.json`.
2. We need a single path to which we can write about 300MB of compiled stuff, as well as use the path for staging builds. This path needs to be finch-version-specific (which can be achieved by being venv-specific), otherwise things will be a little weird if you have multiple versions of finch installed as they will invalidate each other's caches. XDG will tell you to put it in some singular, obscure place, and we can add a version hash to that, like `/Library/Preferences/com.finch-tensor.finch/{version}/cache`. It might be nicer though if we could figure out where python is installing stuff, and put it there. I'm pretty sure an appropriate venv-specific location is `sysconfig.get_path("data")`, according to this ["StackOverflow"](https://stackoverflow.com/a/52638888)

@hameerabbasi, what do you think?

hameerabbasi:
Agreed, except that XDG mostly is Linux-specific: macOS applications and Windows don't use them. To be compliant, I'd use the established library here: https://pypi.org/project/platformdirs/

willow-ahrens:
That's correct, I was mostly assuming platformdirs and XDG were synonymous, but I see now they differ slightly.

[...]

PR #28, opened 2025-05-19
simple roundtrip c compilation, caching, configuration

willow-ahrens:
Adds a simple finch config mechanism, as well as caching to files, and c compilation. Not to be taken as the final word on https://github.com/finch-tensor/finch-tensor-lite/issues/22. Finch assembly depends on this PR and will be added in https://github.com/finch-tensor/finch-tensor-lite/pull/29

### DIFF
--- a/src/finch/__init__.py
+++ b/src/finch/__init__.py
@@ -1,4 +1,5 @@
 from .algebra import element_type, fill_value
+from .codegen import CKernel, get_c_function
 from .interface import (
     AbstractEagerTensor,
     LazyTensor,
@@ -43,4 +44,6 @@ __all__ = [
     "LazyTensor",
     "fill_value",
     "element_type",
+    "get_c_function",
+    "CKernel",
 ]
--- a/src/finch/codegen/__init__.py
+++ b/src/finch/codegen/__init__.py
@@ -1,0 +1,3 @@
+from .c import CKernel, get_c_function
+
+__all__ = ["get_c_function", "CKernel"]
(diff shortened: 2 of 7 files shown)


willow-ahrens:
@hameerabbasi, I've requested your review just to double-check that you're okay with these changes for now.

I've addressed all of your comments from the previous review, except for a comment about file caching that I think is not relevant to this functionality, and limited the scope here to just calling the c compiler and caching. Interestingly, the windows runner does not define any c compiler except gcc. (it does not define cc or cl). I updated the finch config defaults accordingly.

I'd like to merge this separately from the FinchAssembly PR, which may be a little bit more involved.

hameerabbasi (commented):
I'd replace the `config.py` file with calls from `donfig`, as you suggested. Otherwise, LGTM.

As concerns Windows and C compilers -- I can take care of that in a separate PR.

willow-ahrens:
Hameer, I've started down the road of using Donfig, and I think it makes things less clear. Can we stick with the current setup for now and discuss alternative config options in #22?

willow-ahrens:
At the very least, getting donfig right is a little bit overboard in terms of the effort I can spend on this at this time.

willow-ahrens:
Okay, I've moved donfig changes to a separate PR, so I think this one is ready to merge, and we can discuss in https://github.com/finch-tensor/finch-tensor-lite/issues/22

hameerabbasi (approved):
LGTM, thanks @willow-ahrens!

PR #30, opened 2025-05-20
Expand `ruff` checks and fix outstanding issues.

hameerabbasi:
@willow-ahrens We should enable https://pre-commit.ci/ on this repo -- this would ensure that these checks/lints/formatting issues continue to pass.

### DIFF
--- a/src/finch/interface/fuse.py
+++ b/src/finch/interface/fuse.py
@@ -1,20 +1,22 @@
-from .lazy import defer
-from ..finch_logic import FinchLogicInterpreter
 from ..finch_logic import (
-    Plan,
-    Query,
-    Produces,
     Alias,
+    FinchLogicInterpreter,
+    Plan,
+    Produces,
+    Query,
 )
 from ..symbolic import gensym
+from .lazy import defer
 
 """
-This module provides functionality for array fusion and computation using lazy evaluation.
+This module provides functionality for array fusion and computation using lazy
+evaluation.
 
 Overview:
 ---------
-Array fusion allows composing multiple array operations into a single kernel, enabling significant
-performance optimizations by letting the compiler optimize the entire operation at once.
+Array fusion allows composing multiple array operations into a single kernel, enabling
+significant performance optimizations by letting the compiler optimize the entire
+operation at once.
 
 Key Functions:
 --------------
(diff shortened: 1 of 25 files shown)


src/finch/symbolic/__init__.py
```diff
@@ -1,14 +1,23 @@
...
+    PreOrderDFS,
+    Term,
+)
 
 __all__ = [
     "PostOrderDFS",
     "PreOrderDFS",
     "Term",
-    "Rewriter",
+    "Rewrite",
```
hameerabbasi, line 18: Typo maybe? There's no symbol named `Rewriter` that I can find in the library.

src/finch/tensor/tensor.py
```diff
@@ -21,10 +24,10 @@ def __add__(self, other):
     def __mul__(self, other):
         pass
 
+
 def fill_value(arg):
-    if isinstance(arg, LazyTensor):
+    if isinstance(arg, AbstractTensor):
         return arg.fill_value
-    elif isinstance(arg, (int, float)):
+    if isinstance(arg, StableNumber):
```
hameerabbasi, line 31: To make this more robust.

src/finch/tensor/tensor.py
```diff
@@ -21,10 +24,10 @@ def __add__(self, other):
     def __mul__(self, other):
         pass
 
+
 def fill_value(arg):
-    if isinstance(arg, LazyTensor):
+    if isinstance(arg, AbstractTensor):
```
hameerabbasi, line 29: `LazyTensor` was not imported here.

hameerabbasi (commented):
Here are the logic changes -- I assume all of them were bugs caught by the linter.

mtsokol (approved):
Only linting changes - LGTM!

src/finch/symbolic/__init__.py
mtsokol: That's right.

src/finch/interface/fuse.py
```diff
@@ -50,75 +54,82 @@
...
+  or `with_scheduler`.
 """
 
 
 def get_default_scheduler():
     return FinchLogicInterpreter()
 
 
-def compute(arg, ctx=get_default_scheduler()):
+def compute(arg, ctx=None):
```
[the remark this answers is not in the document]
mtsokol: We should have `None` here.

src/finch/tensor/tensor.py
willow-ahrens: I think fill_value is defined in algebra.py, and this definition is extra and should be moved or migrated. Sorry for the dead code

PR #31, opened 2025-05-20
Test `with_subsequence`, `is_subsequence`, and `intersect`

mtsokol:
Hi @willow-ahrens,

There's something that still bugs me about `with_subsequence` and `issubsequence`.

Here I simplified `with_subsequence` removing one `enumerate`, as:
```python
indices = [idx for idx, _ in enumerate(sth)]
for idx, i in enumerate(indices):
    ...
```
will always be the same as:
```python
indices = [idx for idx, _ in enumerate(sth)]
for idx, in indices:
    i = idx
    ...
```

`enumerate()` here always starts at 0 so there's no possibility that `idx` will differ from `i`.

Another thing is that, in my opinion `with_subsequence(x1, x2) == x2` regardless of inputs. We have `if val in x1` where `val` comes from `x2` at `idx`, and then we insert it at `idx` to `res`, so it eventually ends up identical as `x2` (`val`s are instances of dataclasses, mostly `Field`s I guess).

A test case where `with_subsequence(x1, x2) != x2` would help me understand it better, whether I'm missing something.

---

Same goes for `!issubsequence(intersect(idxs_1, idxs_2), idxs_2)`. I see that it's present in one of the rules in `push_fields` - I didn't port it yet, because I can't think of an example where it evaluates to true.

### DIFF
--- a/src/finch/autoschedule/_utils.py
+++ b/src/finch/autoschedule/_utils.py
@@ -1,0 +1,14 @@
+def intersect(x1: tuple, x2: tuple) -> tuple:
+    return tuple(x for x in x1 if x in x2)
+
+
+def is_subsequence(x1: tuple, x2: tuple) -> bool:
+    return x1 == tuple(x for x in x2 if x in x1)
+
+
+def with_subsequence(x1: tuple, x2: tuple) -> tuple:
+    res = list(x2)
+    indices = [idx for idx, val in enumerate(x2) if val in x1]
+    for idx, i in enumerate(indices):
+        res[i] = x1[idx]
+    return tuple(res)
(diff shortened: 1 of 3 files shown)


willow-ahrens:
I don't think this function was ported correctly, the julia version rearranges the elements of one sequence within another, so with_subsequence([1,2,3,4], [4,2]) should return [1,4,3,2]. Perhaps we can rewrite based on the julia version?

mtsokol:
Ah, now it makes sense - I'll update it here then.

mtsokol:
@willow-ahrens Sorry for the noise (here and in other issues) with `is_subsequence` and `with_subsequence`. The port to Python here is all correct, I added a few simple tests here - I got the exact purpose of it now.

willow-ahrens:
No worries, thanks for asking! Sorry I wasn't able to comment further, our timezones aren't aligned. Sounds like this one is cleared up, thanks!

PR #35, opened 2025-05-20
Wma/try donfig

### LINKED ISSUE #22: Where to put finch data and configs?

willow-ahrens:
A PR that shows what it looks like to use donfig for configuration. fixes #22

### DIFF
--- a/src/finch/util/config.py
+++ b/src/finch/util/config.py
@@ -4,6 +4,7 @@ import sys
 import sysconfig
 from importlib.metadata import version
 from pathlib import Path
+
 import donfig  # type: ignore
 
 """
--- a/src/finch/util/config.py
+++ b/src/finch/util/config.py
@@ -57,14 +57,6 @@ def get_version():
     """
     Get the version of Finch.
     """
-    pyproject_path = Path(__file__).parent.parent.parent.parent / "pyproject.toml"
-    if not pyproject_path.exists():
-        raise FileNotFoundError("pyproject.toml not found.")
+    from importlib.metadata import version
 
-    with pyproject_path.open("rb") as f:
-        pyproject_data = tomllib.load(f)
-
-    try:
-        return pyproject_data["project"]["version"]
-    except KeyError:
-        raise ValueError("Version not found in pyproject.toml.") from None
+    return version("finch-tensor-lite")
--- a/src/finch/util/config.py
+++ b/src/finch/util/config.py
@@ -1,0 +1,116 @@
+import json
+import os
+import shutil
+import sys
+import sysconfig
+import tomllib
+from pathlib import Path
+
+"""
+Finch Configuration Module
+
+This module manages configuration settings for the Finch application.
+Finch stores its settings and data in the `FINCH_PATH` directory, which
+defaults to `~/.finch` but can be customized using the `FINCH_PATH`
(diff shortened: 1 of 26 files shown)


src/finch/util/config.py
```diff
@@ -0,0 +1,70 @@
+import os
+import shutil
+import sys
+import sysconfig
```
hameerabbasi, line 4: [`sysconfig`](https://docs.python.org/3/library/sysconfig.html) is meant for querying how Python was built, which is usually very different from the system it's run on.

src/finch/util/config.py
```diff
@@ -0,0 +1,70 @@
...
+    if not pyproject_path.exists():
+        raise FileNotFoundError("pyproject.toml not found.")
+
+    with pyproject_path.open("rb") as f:
+        pyproject_data = tomllib.load(f)
+
+    try:
+        return pyproject_data["project"]["version"]
+    except KeyError:
+        raise ValueError("Version not found in pyproject.toml.") from None
```
hameerabbasi, line 70:
```suggestion
    from importlib.metadata import version

    return version("finch-tensor-lite")
```

hameerabbasi (changes requested):
LGTM other than the two comments.

src/finch/util/config.py
willow-ahrens: I see. I was hoping that `sysconfig.get_path("data")` would allow me to have different Finch caches in each virtualenv.

src/finch/util/config.py
willow-ahrens: Is it wrong that I liked using tomllib? It felt very direct.

hameerabbasi: `pyproject.toml` might not be present in a wheel or user installation of Finch.

src/finch/util/config.py
hameerabbasi: That's fine, my comment was more about `CC` and `CXX`.

hameerabbasi: I don't know of a standard way.

willow-ahrens: for now, I've removed the calls to sysconfig for shared_library_suffix and cc, in favor of more generic defaults. We'll fix it in a separate PR, as you mentioned.

willow-ahrens:
merge based on https://github.com/finch-tensor/finch-tensor-lite/issues/22#issuecomment-2898106784

PR #36, opened 2025-05-21
Introduce type checking

hameerabbasi:
- **Add CI job for `mypy` and update GitHub CI dependencies.**
- **WIP: Fix `mypy` type-checking failures.**

### DIFF
--- a/src/finch/autoschedule/compiler.py
+++ b/src/finch/autoschedule/compiler.py
@@ -1,7 +1,6 @@
-from collections.abc import Hashable
 from textwrap import dedent
-from typing import Any
 
+from ..algebra import fill_value
 from ..finch_logic import (
     Alias,
     Deferred,
@@ -18,11 +17,10 @@ from ..finch_logic import (
 )
 
 
-def get_or_insert(dictionary: dict[Hashable, Any], key: Hashable, default: Any) -> Any:
-    if key in dictionary:
-        return dictionary[key]
-    dictionary[key] = default
-    return default
+def get_or_insert(
+    dictionary: dict[str, LogicNode], key: str, default: LogicNode
+) -> LogicNode:
+    return dictionary.setdefault(key, default)
 
 
 def get_structure(
@@ -44,9 +42,10 @@ def get_structure(
                 Immediate(type(tns.val)),
                 tuple(get_structure(idx, fields, aliases) for idx in idxs),
             )
-        case any if any.is_tree():
-            return any.from_arguments(
-                *[get_structure(arg, fields, aliases) for arg in any.get_arguments()]
+        case any if any.is_expr():
+            return any.make_term(
+                any.head(),
+                *[get_structure(arg, fields, aliases) for arg in any.children()],
             )
         case _:
             return node
@@ -134,7 +133,7 @@ class LogicLowerer:
                     quote
                         {lhs.name} = {compile_logic_constant(tns)}
                         @finch mode = {self.mode} begin
-                            {lhs.name} .= {tns.fill_value}
+                            {lhs.name} .= {fill_value(tns)}
                             {body}
                             return {lhs.name}
                         end
--- a/src/finch/autoschedule/optimize.py
+++ b/src/finch/autoschedule/optimize.py
@@ -38,7 +38,9 @@ def isolate_aggregates(root: LogicNode) -> LogicNode:
                 name = Alias(gensym("A"))
                 return Subquery(name, agg)
 
-    return PostWalk(rule_0)(root)
+    ret = PostWalk(rule_0)(root)
+    assert ret is not None
+    return ret
 
 
 def isolate_reformats(root: LogicNode) -> LogicNode:
@@ -48,7 +50,9 @@ def isolate_reformats(root: LogicNode) -> LogicNode:
                 name = Alias(gensym("A"))
                 return Subquery(name, ref)
 
-    return PostWalk(rule_0)(root)
+    ret = PostWalk(rule_0)(root)
+    assert ret is not None
+    return ret
 
 
 def isolate_tables(root: LogicNode) -> LogicNode:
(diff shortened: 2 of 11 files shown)


src/finch/symbolic/term.py
```diff
@@ -49,8 +49,10 @@ def __hash__(self) -> int:
             )
         return self._hashcache
 
-    def __eq__(self, other: "Term") -> bool:
-        self.head() == other.head() and self.children() == other.children()
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Term):
```
hameerabbasi, line 53: `other` should be an `object`; and if it's incompatible, `NotImplemented` should be returned here.

hameerabbasi:
There's one failure I can't figure out how to address cleanly, [reproduced CI logs](https://github.com/finch-tensor/finch-tensor-lite/actions/runs/15154408097/job/42606335183?pr=36#step:6:14), points to [this line](https://github.com/finch-tensor/finch-tensor-lite/blob/c91d03200fe33854067758707c0a63dfe5a47d11/src/finch/autoschedule/compiler.py#L135)

cc @mtsokol clearly no `LogicNode` has a `fill_value`, so something is wrong here. Do you have a proposed fix?

hameerabbasi:
All logic changes not related to type checking will be pointed out inline.

src/finch/autoschedule/compiler.py
```diff
@@ -44,9 +41,10 @@ def get_structure(
                 Immediate(type(tns.val)),
                 tuple(get_structure(idx, fields, aliases) for idx in idxs),
             )
-        case any if any.is_tree():
-            return any.from_arguments(
-                *[get_structure(arg, fields, aliases) for arg in any.get_arguments()]
+        case any if any.is_expr():
+            return any.make_term(
+                any.head(),
+                *[get_structure(arg, fields, aliases) for arg in any.children()],
```
hameerabbasi, line 47: This was not caught during renames.

src/finch/autoschedule/optimize.py
```diff
@@ -58,12 +62,14 @@ def rule_0(node):
...
+    ret = PostWalk(rule_0)(root)
+    assert ret is not None
+    return ret
 
 
 def pretty_labels(root: LogicNode) -> LogicNode:
-    fields = {}
-    aliases = {}
+    fields: dict[Field, Field] = {}
+    aliases: dict[Alias, LogicNode] = {}
```
mtsokol, line 72: I think it can also be `dict[Alias, Alias]`

src/finch/finch_logic/nodes.py
```diff
@@ -170,8 +170,8 @@ class Table(LogicNode):
         idxs: The fields indexing the tensor.
     """
 
-    tns: LogicNode
-    idxs: tuple[LogicNode]
+    tns: Immediate
```
mtsokol, line 173: Answering your question about missing `fill_value`. If we make this field e.g. `AbstractTensor` from `tensor/tensor.py` and make sure `AbstractTensor` has abstract `fill_value` property, then it should work.

src/finch/symbolic/term.py
mtsokol: Sounds good.

mtsokol (changes requested):

mtsokol:
> cc @mtsokol clearly no `LogicNode` has a `fill_value`, so something is wrong here. Do you have a proposed fix?

I made a suggestion in the review. How about making `tns: AbstractTensor` in the `Table` class? (and making sure `AbstractTensor` has `fill_value` property)

hameerabbasi:
@mtsokol I found another way, does the force-push here look good? https://github.com/finch-tensor/finch-tensor-lite/compare/c91d03200fe33854067758707c0a63dfe5a47d11..01aeb2572f66e5e5beae51b12119d49006f44342

src/finch/autoschedule/optimize.py
hameerabbasi: Done in the [force-pushed commit](https://github.com/finch-tensor/finch-tensor-lite/compare/01aeb2572f66e5e5beae51b12119d49006f44342..3baabd6a6d5be47927341235037bbfc1ae39772a).

hameerabbasi:
Changing `tns` to be typed as `AbstractTensor` leads to other issues:

```
src/finch/autoschedule/compiler.py:42: error: Argument 1 to "Table" has incompatible type "Immediate"; expected "AbstractTensor"  [arg-type]
src/finch/autoschedule/compiler.py:42: error: "AbstractTensor" has no attribute "val"  [attr-defined]
src/finch/autoschedule/compiler.py:111: error: Argument 1 to "compile_logic_constant" has incompatible type "AbstractTensor"; expected "LogicNode"  [arg-type]
src/finch/autoschedule/compiler.py:134: error: Argument 1 to "compile_logic_constant" has incompatible type "AbstractTensor"; expected "LogicNode"  [arg-type]
src/finch/interface/lazy.py:93: error: Argument 1 to "Table" has incompatible type "Immediate"; expected "AbstractTensor"  [arg-type]
src/finch/autoschedule/executor.py:7: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
tests/test_scheduler.py:155: error: Argument 1 to "Reformat" has incompatible type "Immediate"; expected "AbstractTensor"  [arg-type]
tests/test_scheduler.py:156: error: Argument 1 to "Table" has incompatible type "Immediate"; expected "AbstractTensor"  [arg-type]
```

mtsokol:
> @mtsokol I found another way, does the force-push here look good? https://github.com/finch-tensor/finch-tensor-lite/compare/c91d03200fe33854067758707c0a63dfe5a47d11..01aeb2572f66e5e5beae51b12119d49006f44342

Right, calling `fill_value(tns)` instead also works IMO.

mtsokol (approved):

willow-ahrens:
I would have liked to review this, it happened overnight while I was asleep. 😴 I'll take a look and make a separate pr.

hameerabbasi:
@willow-ahrens TBH it would've been a boring review, there are very little logic changes, just making the type checker happy.

The actual PR is on its way. 😄

PR #39, opened 2025-05-22
Add `lift_fields`, and three `propagate_*` pases

mtsokol:
This PR introduces four new passes:

- `lift_fields`
- `propagate_copy_queries`
- `propagate_into_reformats`
- `propagate_transpose_queries`


willow-ahrens (approved):
This looks great, I checked it against the julia version and it's quite similar. Thanks Mateusz!

PR #40, opened 2025-05-22
Move optional deps to project table.

hameerabbasi:


willow-ahrens (approved):
Thanks! I migrated to project table but forgot about the optional deps. This will become even more important once have optional compiler deps.

PR #42, opened 2025-05-25
Use `__mro__` for querying instead of repeated calls to `T.__base__`.

### LINKED ISSUE #41: `query_property` should use `__mro__`
As @hameerabbasi pointed out, `query_property` should use `__mro__` instead of repeated calls to `__base__` when following the class hierarchy to determine the value of properties. See https://github.com/finch-tensor/finch-tensor-lite/blob/4ae03f8ccfe61102fb096092d069a584ac3d56ec/src/finch/algebra/algebra.py#L57-L86.

See (https://docs.python.org/3/howto/mro.html) for more on how method resolution works in python.

If I understand correctly, I think this should be possible by iterating the result of a single call to `__mro__`:

```
>>> O = object
>>> 
>>> class F(O): pass
... 
>>> class E(O): pass
... 
>>> class D(O): pass
... 
>>> class C(D,F): pass
... 
>>> class B(E,D): pass
... 
>>> class A(B,C): pass
... 
>>> A.__mro__
(<class '__main__.A'>, <class '__main__.B'>, <class '__main__.E'>, <class '__main__.C'>, <class '__main__.D'>, <class '__main__.F'>, <class 'object'>)
>>> 
```

hameerabbasi:
Also optimizes the dictionary by performing one query instead of two.

Closes #41

### DIFF
--- a/src/finch/algebra/algebra.py
+++ b/src/finch/algebra/algebra.py
@@ -70,19 +70,20 @@ def query_property(obj: type | Hashable, attr: str, prop: str, *args) -> Any:
     Raises:
         NotImplementedError: If the property is not implemented for the given type.
     """
-    T: type | Hashable
-    if isinstance(obj, type):
-        T = obj
-    else:
-        if isinstance(obj, Hashable) and (obj, attr, prop) in _properties:
-            return _properties[(obj, attr, prop)](obj, *args)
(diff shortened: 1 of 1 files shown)


willow-ahrens (approved):
This looks great! Thanks!

PR #45, opened 2025-05-27
ENH: E2E scheduler test, `normalize_names` and `flatten_plans` passes

mtsokol:
Hi @willow-ahrens,

I think this one should be more interesting to review! Here we have:

- supplemented `optimize()` function - now it matches Julia's `optimize`. Only four passes still missing,
- an E2E test which takes a simple `matmul` plan, passes it to `optimize(plan)` and then runs it with the Interpreter,
- `normalize_names` pass,
- `flatten_plans` pass.

In the next PR I want to ship the remaining passes.


willow-ahrens:
We should not broadcast length-1 dimensions in the interpreter, right? Correct me if I'm wrong, but I thought https://github.com/finch-tensor/Finch.jl/pull/750 was supposed to make broadcasts explicit.

src/finch/finch_logic/interpreter.py
```diff
@@ -72,8 +72,11 @@ def __call__(self, node):
             for arg in args:
                 for idx, dim in zip(arg.idxs, arg.tns.shape, strict=True):
                     if idx in dims:
-                        if dims[idx] != dim:
-                            raise ValueError("Dimensions mismatched in map")
+                        if dims[idx] != dim and dim != 1:
+                            if dims[idx] == 1:  # broadcasting of 1-length dims
```
willow-ahrens, line 76: I believe this behavior is supposed to be avoided, it breaks the "explicit broadcast" semantics of FinchLogic.

src/finch/autoschedule/optimize.py
```diff
@@ -374,6 +399,58 @@ def propagate_transpose_queries(root):
     return _propagate_transpose_queries(root, bindings={})
 
 
+def normalize_names(root):
+    return _normalize_names(root, counts={})
+
+
+def freshen(counts: dict[str, int], symbol: str) -> str:
```
willow-ahrens, line 406: see my comments, could we add the implementation from https://github.com/finch-tensor/finch-tensor-lite/pull/46?

willow-ahrens: We could add it in the same location as #46, in symbolic/environment.py

tests/test_scheduler.py
```diff
@@ -464,3 +472,80 @@ def test_lift_fields():
...
+    )
+
+    result = normalize_names(plan)
+    assert result == expected
+
+
+class HashableEagerTensor(TestEagerTensor):
+    # TODO: @mtsokol This is dreadful but it's only for testing
+    #       purposes. The plan that we pass to scheduler needs
+    #       be hashable, as well arrays that are within.
```
willow-ahrens, line 511: Do we think that we should set the hash for Immediate to use the id of the object?

willow-ahrens (changes requested):
Hi Mateusz, My biggest comment is that this appears to introduce the same bug we fixed in https://github.com/finch-tensor/Finch.jl/pull/750. I think the appropriate way to fix this problem is to instead find where a field is getting assigned to a singleton dimension and then drop that field with a reorder statement. The fix might be in one of the passes, but it could also be in one of the interface functions, introducing a field for a length-1 dimension when it shouldn't.

willow-ahrens:
I see. Looking at what happened to the plan, it looks like a problem with the push-fields rule. in https://github.com/finch-tensor/finch-tensor-lite/blob/3d56b756e3b8c73e779162a9f8ca14c893f534a3/src/finch/autoschedule/optimize.py#L268-L276, instead of `mj.get_fields`, it should be `arg.get_fields`. Sorry I missed that in the last review!

src/finch/autoschedule/optimize.py
willow-ahrens: I think what I meant was to copy the file in symbolic/environment.py from that PR to this one, and `from ..symbolic import Namespace` here.

tests/test_scheduler.py
mtsokol: I think that makes sense.

src/finch/autoschedule/optimize.py
mtsokol: Done!

PR #51, opened 2025-05-29
ENH: `propagate_map_queries_backward` pass

mtsokol:
`propagate_map_queries_backward` pass with a test, and a few algebra rules.


willow-ahrens (approved):
Optional fix, we can make an issue for it

PR #56, opened 2025-06-02
Convert `interpreter.py` to use a `match`-`case` statement.

hameerabbasi:


willow-ahrens (approved):
I'm okay with this, I'm assuming the only change is the match-case. Let me know if anything else is different

PR #58, opened 2025-06-02
add finch assembly, interpreter, c compilation

willow-ahrens:
A minimum viable product demonstrating FinchAssembly, our C-like, backend-agnostic, IR.
fixes https://github.com/finch-tensor/finch-tensor-lite/issues/67


mtsokol:
@willow-ahrens I started reviewing it: of these 2k LOC quite a large portion is just a boilerplate - so it's less complicated than I anticipated.
And with `compiler(verbose=True)` I don't need the reference output tests that I asked you about. Running it verbose myself is enough.

src/finch/finch_assembly/nodes.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+@dataclass(eq=True, frozen=True)
+class Assign(AssemblyTree):
+    """
+    Represents a logical AST statement that evaluates `rhs`, binding the result
+    to `lhs`.
+
+    Attributes:
+        type: The type of the new binding.
```
mtsokol, line 90: I think the `type` is missing below. Or should it disappear from docstring?

src/finch/finch_assembly/nodes.py
```diff
@@ -0,0 +1,360 @@
...
+class Variable(AssemblyExpression):
+    """
+    Represents a logical AST expression for a variable named `name`.
+
+    Attributes:
+        name: The name of the variable.
+    """
+
+    name: str
+    type: Any
```
mtsokol, line 76: There's no mention in the docstring.

src/finch/finch_assembly/nodes.py
```diff
@@ -0,0 +1,360 @@
...
+
+@dataclass(eq=True, frozen=True)
+class Function(AssemblyTree):
+    """
+    Represents a logical AST statement that defines a function `fun` on the
+    arguments `args...`.
+
+    Attributes:
+        name: The name of the function to define as a variable typed with the
+        return type of this function.
```
mtsokol, line 274: nitpick: indentation.

mtsokol (changes requested):

src/finch/codegen/c.py
```diff
@@ -41,41 +51,647 @@ def create_shared_lib(filename, c_code, cc, cflags):
...
+        return blk
+
+    def subblock(self):
+        blk = super().block()
+        blk.indent = self.indent + 1
+        blk.tab = self.tab
+        blk.headers = self.headers
+        blk._headerset = self._headerset
+        blk.bindings = self.bindings.scope()
+        return blk
```
mtsokol, line 529: How about merging it into `def block(self, level: int = 0) -> "CContext":`?

mtsokol (changes requested):

src/finch/codegen/c.py
mtsokol: And we can have an alias:
```py
def subblock(self):
    return self.block(level=1)
```

src/finch/finch_assembly/interpreter.py
```diff
@@ -0,0 +1,245 @@
...
+                        f"but expected type {ret_t}."
+                    )
+
+                self.bindings[func_n] = my_func
+                return None
+            case asm.Return(value):
+                self.ret.append(self(value))
+                return None
+            case asm.Break():
+                self.loop.append([])
```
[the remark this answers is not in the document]
willow-ahrens: We need to pass a reference to a mutable object in the same context as the loop was defined. When we make a subblock, modifying exit_loop in the subblock would not mutate the boolean field in the parent.

mtsokol (commented):
Apart from the previous comments LGMT.

I think Numba backend with NumPy arrays holding data should be fairly straightforward. I'll start working on it tomorrow.

pyproject.toml
wraith1995, line 1: Also, I think now might be the moment to start with mypy here - a lot of the code now is not hard to type.

wraith1995 (commented):
Mainly python advice ATM.

pyproject.toml
willow-ahrens: I think we do run mypy, but a lot of files are still untyped

tests/test_assembly_interpreter.py
```diff
@@ -0,0 +1,80 @@
...
+)
+def test_dot_product(a, b):
+    # Simple dot product using numpy for expected result
+    c = asm.Variable("c", np.float64)
+    i = asm.Variable("i", np.int64)
+    ab = NumpyBuffer(a)
+    bb = NumpyBuffer(b)
+    ab_v = asm.Variable("a", ab.get_format())
+    bb_v = asm.Variable("b", bb.get_format())
+    mod = AssemblyInterpreter()(
```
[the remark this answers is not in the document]
willow-ahrens: I know it's waseful, but I want people to have an example of getting the format from a buffer and using as a type during compilation.

src/finch/codegen/c.py
```diff
@@ -41,41 +51,647 @@ def create_shared_lib(filename, c_code, cc, cflags):
...
 class CKernel:
     """
     A class to represent a C kernel.
     """
 
-    def __init__(self, function_name, c_code):
-        self.function_name = function_name
-        self.c_code = c_code
-        self.c_function = get_c_function(function_name, c_code)
+    def __init__(self, c_function, ret_type, argtypes):
```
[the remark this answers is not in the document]
willow-ahrens: i'm not very moved to make these dataclasses, the don't feel very "data" to me I guess? I guess I'm not sure that it makes things clearer to fit these into dataclasses

src/finch/finch_assembly/interpreter.py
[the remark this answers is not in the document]
willow-ahrens: I added a HaltState struct

mtsokol (approved):
I went through it once more I agree with all the changes - approved!

PR #59, opened 2025-06-02
4 implement elementwise ops

yusheng036:
Added bitwise functions, truediv, floordiv, mod and pow. I have also added all of the functions above to interface.__init__.py, interface.overrides.py and finch.__init__.py. I have added test in test_algebra.py and test_interface.py to reflect the changes made in lazy.py and eager.py.

```
from finch import floordiv 
import numpy

a = numpy.array([5, 7])
b = numpy.array([2, 3])

floordiv(a, b)

```

Output: 
2, 2

### DIFF
--- a/tests/test_algebra.py
+++ b/tests/test_algebra.py
@@ -1,12 +1,32 @@
 import math
 import operator
 
-from finch.algebra import is_annihilator, is_distributive, is_identity
-
+from finch.algebra import is_annihilator, is_distributive, is_identity, is_associative
 
 def test_algebra_selected():
     assert is_distributive(operator.mul, operator.add)
+    assert is_distributive(operator.mul, operator.sub)
+    assert is_distributive(operator.and_, operator.or_)
+    assert is_distributive(operator.and_, operator.xor)
+    assert is_distributive(operator.or_, operator.and_)
     assert is_annihilator(operator.add, math.inf)
     assert is_annihilator(operator.mul, 0)
+    assert is_annihilator(operator.or_, 1)
+    assert is_annihilator(operator.and_, 0)
     assert is_identity(operator.add, 0)
     assert is_identity(operator.mul, 1)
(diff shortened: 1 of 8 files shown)


willow-ahrens (approved):
This looks great to me! I don't have any big changes to request, I am curious what others think.

willow-ahrens:
@hameerabbasi, I approved this and it auto-merged, but if you have time to quickly look over I would appreciate your comments.

willow-ahrens (commented):
I have one last comment: The array API specifies the name `bitwise_left_shift`, not `bitwise_lshift`. I think you may have picked up `lshift` from the dunder method name. I'll file an issue to fix.

tests/test_algebra.py
```diff
@@ -1,12 +1,30 @@
...
 def test_algebra_selected():
     assert is_distributive(operator.mul, operator.add)
+    assert is_distributive(operator.mul, operator.sub)
+    assert is_distributive(operator.and_, operator.or_)
+    assert is_distributive(operator.and_, operator.xor)
+    assert is_distributive(operator.or_, operator.and_)
     assert is_annihilator(operator.add, math.inf)
     assert is_annihilator(operator.mul, 0)
+    assert is_annihilator(operator.or_, 1)
+    assert is_annihilator(operator.and_, 0)
```
hameerabbasi, line 16: Perhaps these should be `True` and `False`?

tests/test_algebra.py
```diff
@@ -1,12 +1,30 @@
...
     assert is_identity(operator.add, 0)
     assert is_identity(operator.mul, 1)
+    assert is_identity(operator.or_, 0)
+    assert is_identity(operator.and_, 1)
```
hameerabbasi, line 20: Same here.

hameerabbasi (commented):
Two minor comments other than what @willow-ahrens already observed.

PR #69, opened 2025-06-02
Implementation of linalg function. Resolves #2

### LINKED ISSUE #2: Define lazy tensordot, lazy outer, lazy stacked matmul
Spec for matmul: https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.matmul.html
Spec for tensordot: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.tensordot.html
Spec for outer: https://data-apis.org/array-api/2022.12/extensions/generated/array_api.linalg.outer.html

These functions can be defined in terms of functions already implemented in https://github.com/finch-tensor/finch-tensor-lite/blob/main/src/finch/interface/lazy.py, using expand_dims, maybe permutedims (I suspect it's not needed), broadcast, and reduce. 

The definition would be similar to https://github.com/finch-tensor/finch-tensor-lite/blob/0aeaa30517692b6a78f6a940af5f844460a91eec/src/finch/interface/lazy.py#L143, but perhaps more complex as it would need to satisfy array API requirements.

Osmansiddiquer:
This pull request introduces two new mathematical operations, `matmul` (matrix multiplication) and `matrix_transpose` (matrix transposition), to the Finch library. These operations are implemented for both eager and lazy evaluation contexts, with comprehensive testing added to ensure correctness. Additionally, some imports in `fuse.py` were reorganized for clarity so that module docstring is at the top.

### New Mathematical Operations:

* **`matmul` (Matrix Multiplication):**
  - Added a `matmul` function to `src/finch/interface/eager.py` for eager evaluation, which computes the matrix product, supporting both `EagerTensor` and `LazyTensor` inputs.
  - Added a `matmul` function to `src/finch/interface/lazy.py` for lazy evaluation, including support for 1D, 2D, and batched matrix multiplication while handling broadcasting and dimension mismatch errors.
  - Included `matmul` in the public API by updating `src/finch/__init__.py` and `src/finch/interface/__init__.py`. [[1]](diffhunk://#diff-152d80afa48138af730ca8d974bf44f722cf6f6788682fad3c171c2e57dcb1c5R20-R21) [[2]](diffhunk://#diff-152d80afa48138af730ca8d974bf44f722cf6f6788682fad3c171c2e57dcb1c5R60-R61) [[3]](diffhunk://#diff-a90f9b16bd7c4c416df304a4b182460f440be350231ea390f16a3978c121871bR13-R14) [[4]](diffhunk://#diff-a90f9b16bd7c4c416df304a4b182460f440be350231ea390f16a3978c121871bR53-R54)

* **`matrix_transpose` (Matrix Transposition):**
  - Added a `matrix_transpose` function to `src/finch/interface/eager.py` for eager evaluation, which computes the transpose of a matrix or stack of matrices.
  - Added a `matrix_transpose` function to `src/finch/interface/lazy.py` for lazy evaluation, with validation for input dimensionality.
  - Included `matrix_transpose` in the public API by updating `src/finch/__init__.py` and `src/finch/interface/__init__.py`. [[1]](diffhunk://#diff-152d80afa48138af730ca8d974bf44f722cf6f6788682fad3c171c2e57dcb1c5R20-R21) [[2]](diffhunk://#diff-152d80afa48138af730ca8d974bf44f722cf6f6788682fad3c171c2e57dcb1c5R60-R61) [[3]](diffhunk://#diff-a90f9b16bd7c4c416df304a4b182460f440be350231ea390f16a3978c121871bR13-R14) [[4]](diffhunk://#diff-a90f9b16bd7c4c416df304a4b182460f440be350231ea390f16a3978c121871bR53-R54)

### Testing:
* All tests were conducted with eager, deferred, and plain numpy tensors
* Added parameterized tests for `matmul` to validate various cases, including 1D, 2D, batched, and broadcasting scenarios, as well as error handling for mismatched dimensions.
* Added parameterized tests for `matrix_transpose` to validate correctness across different input shapes and evaluation contexts.

### Future Work:
* Implementation of tensordot and vecdot remains


README.md
```diff
@@ -77,3 +77,11 @@ tests:
 ```bash
 poetry run pytest
 ```
+
+#### Optional Static Type Checking
+
+The CI/CD pipeline will run mypy to check for type errors, so you don't need to run it locally, but if you want to check types locally, you can run:
```
willow-ahrens, line 83: More precisely, `pytest` will run `mypy`, so you don't need to run `mypy` separately. it's not a distinction between local and CI.

willow-ahrens (commented):
Only a small change about the wording of the readme. Looking good so far!

src/finch/interface/lazy.py
```diff
@@ -563,3 +664,124 @@ def mod(x1, x2) -> LazyTensor:
...
+
+def conjugate_op(x):
+    if hasattr(x, "conjugate"):
+        return x.conjugate()
+    return x
+
+
+# register the conjugate operation return type. The conjugate operation
+# preserves the element type of the input tensor.
+register_property(
```
Osmansiddiquer, line 677: I am not sure if this is the write way to implement the element wise conjugate and especially the register_property. This is needed for the vecdot operation,

willow-ahrens: This looks correct. Let's move it to algebra/operator.py, with the other standard operators we need to introduce.

willow-ahrens: A more precise implementation might define it in terms of 
`query_property(x, "conjugate", "return_type")`, because we ultimately call the `conjugate()` method of `x`. However, I think you're correct that we should instead blanket restrict `conjugate_op` to return the same datatype as it's input because that's the spec in array_api.

src/finch/interface/lazy.py
```diff
@@ -563,3 +664,124 @@ def mod(x1, x2) -> LazyTensor:
...
+    x2p = permute_dims(x2, tuple(newaxes_b))
+
+    # Expand x1p and x2p so that their contracted axes align for broadcasting
+    # x1p: shape (..., K1, K2, ..., Kn)
+    # x2p: shape (K1, K2, ..., Kn, ...)
+    # We want to multiply so that contracted axes are aligned
+
+    # For x1p, add len(notin_b) singleton dims at the end
+    for _ in range(len(notin_b)):
+        x1p = expand_dims(x1p, axis=-1)
```
willow-ahrens, line 754: call expand_dims with axis=(-1, -2, -3), etc. It will generate a more succinct expression here.

src/finch/interface/lazy.py
```diff
@@ -563,3 +664,124 @@ def mod(x1, x2) -> LazyTensor:
...
+    # For x2p, add len(notin_a) singleton dims at the front
+    for _ in range(len(notin_a)):
+        x2p = expand_dims(x2p, axis=0)
```
willow-ahrens, line 757: Same here, calling expand_dims once with multiple dims is preferable.

src/finch/interface/lazy.py
```diff
@@ -563,3 +664,124 @@ def mod(x1, x2) -> LazyTensor:
...
+    if x1.shape[axis] != x2.shape[axis]:
+        raise ValueError(
+            "Shapes are not compatible for vector dot product along the specified axis"
+        )
+
+    return reduce(
+        operator.add,
+        multiply(conjugate(x1), x2),
+        axis=axis,
+    )
```
willow-ahrens, line 787: Solid implementation of vecdot.

PR #75, opened 2025-06-04
ENH: `concordize` pass

mtsokol:
`concordize` pass.


willow-ahrens (approved):
This looks correct to me. There's a difference in how we handle nested plans, but we don't have a good story about what a program even means if it has nested plans.

PR #77, opened 2025-06-05
Add `If` and `IfElse` assembly nodes

mtsokol:


willow-ahrens (approved):
Looks good to me! I will also add tests for the c backend.

PR #84, opened 2025-06-07
Wma/finch notation

### LINKED ISSUE #79: Implement FinchNotation

### LINKED ISSUE #81: Implement NotationInterpreter

willow-ahrens:
add finch notation. fixes #79 fixes #81. I suppose I should elaborate on this PR a bit for whoever feels moved to review it: this PR adds FinchNotation to the repo, but it's a little bit simplified from Finch.jl. Because Finch.jl implemented FinchNotation as a frontend, it made a lot of very complex assumptions about e.g. automatically determining tensor modes, automatically making canonical expressions, automatically adding wrapper arrays. We have a reduced time budget, and I think it will help the clarity of this repo if we made a more constrained (but equally expressible) FinchNotation. The intent here is that only developers, not users, will be writing FinchNotation in this repo. If we feel so moved, at some point we can make a nice frontend to this, but for now, we'll stick with the simpler version and get this done in the next two months.


src/finch/finch_notation/nodes.py
```diff
@@ -0,0 +1,396 @@
...
+@dataclass(eq=True, frozen=True)
+class Update(NotationTree):
+    op: NotationNode
+
+    def children(self):
+        return [self.op]
+
+
+@dataclass(eq=True, frozen=True)
+class Increment(NotationTree):
```
mtsokol, line 135: Is there a better name than `Increment`? I think that `lhs = lhs - rhs` also counts as an increment now.

mtsokol (changes requested):
It's more a bunch of questions to understand it rather than a review.

PR #86, opened 2025-06-09
ENH: Numba backend

mtsokol:
Hi @willow-ahrens,

This PR introduces Numba backend.

### DIFF
--- a/tests/test_codegen.py
+++ b/tests/test_codegen.py
@@ -122,8 +122,7 @@ def test_codegen(compiler, buffer):
     )
     mod = compiler(prgm)
     f = mod.test_function
-
-    result = f(buf)
+    f(buf)
     result = buf.arr
     expected = np.array([1, 2, 3, 2, 3, 4], dtype=np.float64)
     assert_equal(result, expected)


willow-ahrens (changes requested):
NumbaCompiler, NumbaContext, etc. look good to me. I'm requesting that we not reorganize the abstract type hierarchy for the buffers. The buffers were designed so that we could define buffer interfaces for each backend. For example, I was thinking we would just define a `NumbaBuffer` interface similar to `CBuffer` interface, so that we could have `class NumpyBuffer(CBuffer, NumbaBuffer)` implement both. I would really like to avoid two different classes with the same name. In the same way that `NumpyBuffer` currently implements `serialize_to_c` which returns `NumpyCBuffer`, I was thinking we would eventually implement `serialize_to_numba` which returns `NumpyNumbaBuffer` or perhaps just `ndarray`. The two classes can coexist, and the mechanism by which we switch from one to another was meant to be the `serialize_to` call, which each backend would require their own version of. 

Is there a reason you needed to require two separate buffers from the start? I liked the idea of having one buffer type because it allows us to have e.g. a `CSCTensorFormat(NumpyBufferFormat)` which we can easily compile to either backend, or maybe both if we're debugging. I think it's less confusing to just have multiple interfaces on the same buffer object.

src/finch/codegen/numpy_buffer.py
```diff
@@ -25,7 +26,17 @@ class NumpyCBuffer(ctypes.Structure):
...
 
-class NumpyBuffer(Buffer, CArgument):
+class NumbaArgument(ABC):
+    @abstractmethod
+    def serialize_to_numba(self):
+        """
+        Return a Numba-compatible object to be used in place of this argument
+        for the Numba backend.
+        """
+        ...
```
willow-ahrens, line 36: Should this be in numba_backend.py?

src/finch/codegen/numpy_buffer.py
```diff
@@ -70,8 +81,17 @@ def deserialize_from_c(self, c_buffer):
         Update this buffer based on how the C call modified the NumpyCBuffer structure.
         """
 
+    def serialize_to_numba(self):
+        return self.arr
```
willow-ahrens, line 85: I'm concerned that if the program calls resize on the result of serialize_to_numba, that the new array won't be passed back into this numpy buffer type. It's unfortunate, but `np.resize` returns a new object, so the reference in this numpy buffer class will have gone stale.

src/finch/codegen/numpy_buffer.py
```diff
@@ -70,8 +81,17 @@ def deserialize_from_c(self, c_buffer):
...
+
+
+class NumbaBufferFormat:
+    @staticmethod
+    def full_name():
+        return "numpy.ndarray"
```
willow-ahrens, line 91: is this still needed? If it's needed, does it make sense to give it a more numba-related name? I'm guessing the purpose is to emit the name of the class returned by `serialize_to_numba`?

willow-ahrens (changes requested):
I have a few clarifying questions/comments. My biggest concern is that the numpyBuffer class will not be updated with the reference to the result of `np.resize`.

src/finch/codegen/numpy_buffer.py
willow-ahrens: that's fine, but we need to update the reference to the outer numpy buffer. We might instead pass a reference to the numpy array, and update the reference when needed?

src/finch/codegen/numpy_buffer.py
mtsokol: Correct, it's needed for code generation to have `a: numpy.ndarray`. I renamed it to `def numba_name()` and moved the definition of the class to `numba_backend.py`.

src/finch/codegen/numpy_buffer.py
mtsokol: So calling `np.resize(arr, size)` in the Numba backend should modify an instance of NumpyBuffer that this `arr` came from? Like `numpy_buffer_resize_callback`?

IIUC I'm not sure how this can't be achieved - We would need some kind of a callback to NumpyBuffer object but we can't pass custom classes to jited functions.

willow-ahrens: can we pass NamedTuples to jited functions?

mtsokol: @willow-ahrens The last commit is my proposition for supporting "in-place" resize for Numba, but it also impacts C backend.

Let's wrap the NumPy array in the `NumpyBuffer` in a list and keep it as `self.arr_ref = [arr]`. This way in Numba I access it with `arr_ref[0]` and can update the reference to it.
For C backend it means that `arr_ref` is in the `NumpyCBuffer` but nothing else much.

WDYT? Is it too much of a hacky approach or still reasonable?

src/finch/codegen/numpy_buffer.py
```diff
@@ -36,31 +36,32 @@ def __init__(self, arr: np.ndarray):
...
-        return self.arr.size
+        return self.arr_ref[0].size
 
     def load(self, index: int):
-        return self.arr[index]
+        return self.arr_ref[0][index]
 
     def store(self, index: int, value):
-        self.arr[index] = value
+        self.arr_ref[0][index] = value
```
willow-ahrens, line 53: Instead of messing with the NumpyCBuffer, could you make a numba-focused buffer type like NumbaBuffer to do this transform on?

willow-ahrens:
For both of these buffer types, I'll need to eventually add some sort of compiler-unpacking code to make it so that we don't need a double-pointer every time we want to access the array.

tests/test_codegen.py
```diff
@@ -103,25 +121,20 @@ def test_codegen():
...
-    ctx = CCompiler()
-    mod = ctx(prgm)
+    mod = compiler(prgm)
     f = mod.test_function
 
-    a = np.array([1, 2, 3], dtype=np.float64)
-    b = finch.NumpyBuffer(a)
-    f(b)
-    result = b.arr
+    result = f(buf)
```
mtsokol, line 127: Just `f(buf)` should be here.

willow-ahrens (changes requested):
Sorry for the back and forth on this. I added my suggestions, I hope they make sense to you. The idea is to not need separate buffer types for different backends, and instead to just require additional interfaces. My only issue right now is that there's still one failure related to resize. working on it.

PR #88, opened 2025-06-09
Changes user-facing assertions to proper errors

### LINKED ISSUE #83: Do not use assert for user-facing errors
quoting @Osmansiddiquer: 
>  we shouldn't use assert used for user-facing validation since relying on assert statements means all checks vanish under python -O. which we might need to do for our production builds. We can use asserts for debugging and sanity checks, however.

This is very true. I'd like to fix this in the codebase and add something to the Readme about a style guide on this point.

Osmansiddiquer:
Resolves #83 .

- Adds a ./src/ readme that serves as a primer into the code-base. The aim is to document any design principles and serve as a style guide for our code.
- Fixes some of the user-side assertions in the interface. For now, I have changed all the assertions in the interface (lazy.py)

_This has been merged with os/linalg-funcs so only merge this in after merging #69_

**For the reviewer:** We use a lot of assertions in the code and it is hard to tell which ones may end up floating up to be user-facing. The issue is that it is hard to identify, with proper messaging and Exception types, what error should be shown to the users at lower levels. One way to go about this would be to introduce an assert utility function that can be used for assertions that we would want to persist in the production builds (these will be internal, and effort must be made that they are not exposed to the user but will at-least be understandable/debuggable by the maintainers in case they do show up). python asserts can then be used for sanity-checks that can be optimized away without harm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,3 +80,12 @@ tests:
 ```bash
 poetry run pytest
 ```
+
+#### Optional Static Type Checking
+
+The pytest will run mypy to check for type errors, so you shouldn't need to run it manually.
+In case you do need to run mypy manually, you can do so with:
+
+```bash
+poetry run mypy ./src/
+```
--- a/src/finch/README.md
+++ b/src/finch/README.md
@@ -1,0 +1,45 @@
+# Finch Tensor: Developer README
+
+> **Note:** This README is AI-generated and may contain errors or omissions. If you find any inaccuracies or improvements, please update this file for the benefit of all contributors.
+
+## Overview
+
+This directory contains the implementation for Finch.py, a tensor compiler and runtime designed for extensibility and integration with other frameworks (such as NumPy). The codebase is organized to support both eager and lazy tensor computation, algebraic rewrites, and custom user extensions.
+
+## Directory Structure
+
+- `algebra/` — Algebraic properties, type promotion, and operator logic. Use this to define, register, and query custom properties for classes and operators.
+- `interface/` — User-facing tensor APIs, including lazy and eager tensor implementations. The aim is to provide a consistent interface for tensor operations according to the [array API specification](https://data-apis.org/array-api/latest/).
+- `symbolic/` — Symbolic utilities for tensor computation.
+- `finch_logic/` — Internal logic node representations and manipulation.
+- `overrides.py` — Mechanisms for user and library overrides.
+- `__init__.py` — Module initialization and exports.
+
+## Contributing
+
+- **If you find an error or unclear section in this README or the code, please fix it or open an issue.**
+- Follow the code style and conventions used throughout the project.
+- Add or update docstrings for all public functions and classes.
+- To to write thorough tests for new features and bug fixes in the `tests/` directory.
+
+## Assertions and Validation
+
+- **Do not use `assert` statements for user-facing validation.**
+    - Relying on `assert` for input validation or error checking is unsafe, as all `assert` statements are removed when Python is run with the `-O` (optimize) flag, which may be used for production builds.
+    - Use explicit error handling (e.g., `if ...: raise ValueError(...)`) for all user-facing checks and input validation. For user facing APIs, you should raise exceptions according to the [array API specification](https://data-apis.org/array-api/latest/).
+- `assert` statements **may** be used for internal debugging, invariants, and sanity checks that are not critical for production correctness.
+
+## Testing
+
+- Tests are located in the `tests/` directory at the project root.
+- Use `poetry run pytest` to run the test suite.
+- Add tests for all new features and bug fixes.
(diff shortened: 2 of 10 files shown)


src/finch/README.md
```diff
@@ -0,0 +1,45 @@
+# Finch Tensor: Developer README
+
+> **Note:** This README is AI-generated and may contain errors or omissions. If you find any inaccuracies or improvements, please update this file for the benefit of all contributors.
```
willow-ahrens, line 3: Please read the readme yourself and correct any AI errors. Although AI is a helpful coding tool, I'd appreciate if everyone would personally stand by the code they are asking others to review. Perhaps you could add a statement to this effect in the readme too.

willow-ahrens (changes requested):
Instead of putting this in src/finch/Readme, I think this makes more sense as a CONTRIBUTING.md at the root of the repo. Additionally, I would appreciate if you would personally read and correct anything AI generated. Thanks!

willow-ahrens:
this PR only changes one assert statement. is that really the only place we used assert and not a user-facing error?

Osmansiddiquer:
> this PR only changes one assert statement. is that really the only place we used assert when we shouldn't?

Most of these were addressed in the #69 . No file in the interface currently has an assert in it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,102 @@
...
+- `__init__.py` — Module initialization and exports for the Finch package.
+
+## Testing
+
+- Tests are located in the `tests/` directory at the project root.
+- Add tests for all new features and bug fixes. Write thorough tests for new features and bug fixes in the `tests/` directory.
+
+## Documentation
+
+- All public APIs should have clear docstrings following the NumPy style.
```
willow-ahrens, line 98: are we really following numpy style? I thought we were doing google style for a while.

README.md
```diff
@@ -1,91 +1,37 @@
...
+- **Flexible Execution:** Effortlessly switch between eager (immediate) and lazy (deferred, fused) evaluation to balance simplicity and performance.
+- **Advanced Optimization:** Benefit from algebraic and symbolic rewrites, constant folding, and other compiler-level optimizations—no manual tuning required.
+- **Rich Array Support:** Work seamlessly with dense, sparse, and structured arrays, including custom background (zero) values and user-defined formats.
+- **Interoperable:** Integrates smoothly with NumPy and the broader Python scientific ecosystem.
+- **Modern API:** Follows the [Array API specification](https://data-apis.org/array-api/latest/) for consistency and compatibility.
 
 ## Installation
 
-`finch-tensor-lite` is available on PyPi, and can be installed with pip:
+Install from PyPI:
```
willow-ahrens, line 16: I liked the previous sentence, why did this change?

README.md
```diff
@@ -1,91 +1,37 @@
...
-
-### Packaging
-
-Finch uses [poetry](https://python-poetry.org/) for packaging.
-
-To install for development, clone the repository and run:
+For development (with tests and pre-commit hooks):
 ```bash
+git clone https://github.com/finch-tensor/finch-tensor-lite.git
+cd finch-tensor-lite
```
willow-ahrens, line 24: It's kindof confusing to tell users to run pre-commit hooks. I think we need just the regular install instructions here, and dev install instructions go in contributing.md.

willow-ahrens (changes requested):
I've requested a lot of changes. In general, I'd prefer if we just added a statement about asserts to the existing readme, and move contributing-related sections to the contributing guide, rather than filling in readme sections with AI.

README.md
```diff
@@ -13,79 +9,19 @@ The source code for `finch-tensor-lite` is available on GitHub at [https://githu
...
-```bash
-poetry run pre-commit run -a
-```
-
-### Testing
+## Usage
 
-Finch uses [pytest](https://docs.pytest.org/en/latest/) for testing. To run the
-tests:
+```python
```
Osmansiddiquer, line 14: I tried to write a succinct example that demonstrates what finch does. This is not in the original readme. Let me know if this is needed, or if you'd like a different example.

Osmansiddiquer:
This PR has become more about readme than fixing the issue of user-side assertions 😅. I have changed the title and the description to reflect that. It no longer addresses #83 . We still need to figure out what to do with lower level assert statements that may float up to the user-level

Osmansiddiquer:
Internal asserts will be removed if we make an optimized build for production. This will cause erroneous values to propagate downwards which may lead to poorly described exceptions from the backend we use, or incorrect garbage results at worse. Rule of thumb is to generally have well-defined pre-conditions for all functions and throwing relevant exceptions if they fail instead of using assert. https://github.com/pydantic/pydantic/discussions/4575

PR #89, opened 2025-06-09
Wma/levels

willow-ahrens:
A proof of concept for level formats in finch-tensor-lite. I've taken special care to make sure that levels have "format" factory constructors.

working towards https://github.com/finch-tensor/finch-tensor-lite/issues/80


hameerabbasi:
> While this is still a draft, I've ended up throwing a lot of `@property` decorators into this and I wonder if this is the "pythonic" way to do things. I'm open to feedback on this design before things get too baked in.

Yes it is; as long as the access is `O(1)` should be good.

But for level formats what you probably want are dataclasses; they provide an easy match interface, immutability and other niceties.

src/finch/tensor/level/element_level.py
```diff
@@ -0,0 +1,108 @@
+from typing import Any
+
+import numpy as np
+
+from ...codegen import NumpyBufferFormat
+from ..tensor import Level, LevelFormat
+
+
+class ElementLevelFormat(LevelFormat):
```
hameerabbasi, line 9: Is there a reason this isn't a `dataclass`?

src/finch/codegen/numpy_buffer.py
```diff
@@ -90,7 +90,7 @@ class NumpyBufferFormat(CBufferFormat, NumbaBufferFormat):
     """
 
     def __init__(self, dtype: type):
-        self._dtype = dtype
+        self._dtype = np.dtype(dtype).type
 
     def __eq__(self, other):
         if not isinstance(other, type(self)):
```
hameerabbasi, line 96: Maybe `type(other) is type(self)`?

Types are singletons, so comparing them with equality is inefficient; and this statement accepts subclasses with `other`

src/finch/tensor/level/dense_level.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __call__(self, shape):
+        """
+        Creates an instance of DenseLevel with the given format.
+        Args:
+            fmt: The format to be used for the level.
+        Returns:
+            An instance of DenseLevel.
+        """
+        lvl = self.lvl(shape[1:])
```
hameerabbasi, line 27: Let's be explicit here and say `shape=shape[1:]`?

src/finch/tensor/level/dense_level.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        self.fmt = fmt
+        self.lvl = lvl
+        self.dimension = dimension
+
+    @property
+    def shape(self):
+        return (self.dimension, *self.lvl.shape)
+
+    def get_format(self):
```
hameerabbasi, line 87: Maybe also a `@property`?

src/finch/tensor/tensor.py
```diff
@@ -1 +1,215 @@
...
+    @abstractmethod
+    def buffer_format(self):
+        """
+        Returns the format of the buffer used for the fibers.
+        This is typically a NumpyBufferFormat or similar.
+        """
+        ...
+
+
+class Level(Formattable, ABC):
```
hameerabbasi, line 63: Maybe a `dataclass` with inheritance and a `fmt` member?

hameerabbasi (changes requested):
The overarching design suggestion here is to use more dataclasses. If hash caching is a concern; we can define a `dataclass_hash` function which forces a `_hash` onto a dataclass and returns that if it's present.

Then in each place it's used; we can do the following to avoid repetition:

```python

def dataclass_hash(x: ...) -> int:
    """ Get members of dataclass and force-add a _hash if not already there """

@dataclass(frozen=True, eq=True)
class SomeDataclass:
    ...

    __hash__ = dataclass_hash

```

This should get rid of a lot of boilerplate code in this PR.

willow-ahrens:
> > While this is still a draft, I've ended up throwing a lot of `@property` decorators into this and I wonder if this is the "pythonic" way to do things. I'm open to feedback on this design before things get too baked in.
> 
> Yes it is; as long as the access is `O(1)` should be good. Even `O(ndim)` is okay imho.
> 
> But for level formats what you probably want are dataclasses; they provide an easy match interface, immutability and other niceties.

My confusion is that a dataclass with an 'element_type' attribute doesn't seem to count as an implementation of the '@property @abstractmethod def element_type' abstract method.

hameerabbasi:
Ah; yes. That would have to be added on each base class. Alternatively, it could be added directly on the super dataclass; and it's inherited in all child dataclasses

willow-ahrens:
Honestly it's so much more painful to use a dataclass here if I have to define setters and getters for the variable that have the same name. Consider:
```py
@dataclass
class ElementLevelFormat(LevelFormat):
    fill_value: Any
    element_type: Optional[type | Format] = None
    position_type: Optional[type | Format] = None
    buffer_format: Any = NumpyBufferFormat
    val_format: Any = None

    def __post_init__(self):
        if self.element_type is None:
            self.element_type = get_format(self.fill_value)
        if self.val_format is None:
            self.val_format = self.buffer_format(self.element_type)
        if self.position_type is None:
            self.position_type = self.buffer_format.length_type()
        self.element_type = algebra.element_type(self.val_format)
        self.fill_value = self.element_type(self.fill_value)

    def __call__(self, shape):
        """
        Creates an instance of ElementLevel with the given format.
        Args:
            fmt: The format to be used for the level.
        Returns:
            An instance of ElementLevel.
        """
        return ElementLevel(self)

    @property
    def ndims(self):
        return 0

    @property
    def fill_value(self):
        return self.__dict__['fill_value']
```
the above doesn't work because I also need to define a setter for fill_value. Do we really think this is nicer than just defining a class with an init method?

mtsokol:
IMO having a simple class is better than a `@dataclass` and a puzzling `__post_init__`+setters.

I think dataclasses are useful for reducing boilerplate, but only when the use case doesn't exceed capabilities of a dataclass.

src/finch/codegen/numpy_buffer.py
hameerabbasi: Do the types need to be exactly the same for equality to hold? Or can `instance_of_parent == instance_of_subclass` be true as well?

If the former, we should really do `type(self) is not type(other)`.

If the latter, we should do `isinstance(other, CommonParentClass)` to make sure everything stays symmetrical. With the current code; `instance_of_subclass == instance_of_parent` will never be true.

mtsokol (approved):

mtsokol:
LGTM, I put one comment where one line can be reverted to pre-Numba PR version.

hameerabbasi (approved):
LGTM.

Issue #92, opened 2025-06-10
RFC: style for abstract attributes


willow-ahrens:
Okay, throughout the repo, we should choose between

1. get_* methods.
    a. Advantages: really simple to understand and implement. Property can be named `foo`, and the accessor named `get_foo`
    b. Disadvantages: A little complex to enforce when we should use `x.foo` and `x.get_foo()`.
2. `@property @abstractmethod foo`
    a. Advantages: Very clear, automatically enforced.
    b. Disadvantages: Makes it hard to create a dataclass with a member named `foo`, because the member is not an `@property` accessor, and then to define the accessor we need to create a variable with another name called `foo_` for the purposes of creating an accessor
3. simply document that classes should have an attribute and trust implementers to do the right thing.
    a. Advantages: Very flexible in terms of implementation, `@property` vs. normal attribute. Can always just do `x.foo`, no ambiguity
    b. Disadvantages: Property not enforced automatically on instantiate of the class

I have tried 3. in https://github.com/finch-tensor/finch-tensor-lite/pull/89, and I like it. I dislike 2. because it means that the actual attribute names sometimes need to be different than the @property method name, which is a headache.

willow-ahrens:
Note that 1 will disagree with numpy style, etc for many attributes like `shape` or `dtype`.

willow-ahrens:
Some opinions of random people online:

https://stackoverflow.com/questions/74200729/is-there-a-way-to-make-an-inherited-abstract-property-a-required-constructor-arg
  - argues for our option 3, since abstract properties don't mean what we want. Does point out that attributes with defaults or perhaps slots=true might do what we want

https://florimond.dev/en/posts/2018/10/reconciling-dataclasses-and-properties-in-python
  - Describes the consequences of option 2. I don't expect new users to have an enjoyable time getting this right.

hameerabbasi:
Right -- I did discover the consequences of using 3 practically in `sparse`; in the legacy backend. Lots of people directly reached in, and modified attributes, particularly `fill_value` and constituent arrays, then raised issues about how it broke in various ways due to the introduction of inconsistencies.

We'll have to mark all such issues as wontfix and point to the docs.

hameerabbasi:
Or we can do it inside `__init__` of the parent class which is usually but not necessarily called from child classes.

willow-ahrens:
we might add __init__subclass and wrap the init method of that class to also validate args.

willow-ahrens:
[...]

PR #95, opened 2025-06-10
add abstract attributes

willow-ahrens:
A way to enforce that base classes have the required attributes when they are constructed.

fixes https://github.com/finch-tensor/finch-tensor-lite/issues/92

### DIFF
--- a/src/finch/util/abs_attr.py
+++ b/src/finch/util/abs_attr.py
@@ -1,0 +1,66 @@
+class AbsAttr():
+    """
+    A class to represent abstract required attributes.
+
+    ```
+        # Example usage:
+        class foo(AbsAttr):
+            @classmethod
+            def __attrs__(cls):
+                return ["foo"]
+
+        class bar(foo):
+            def __init__(self, qux):
+                self.foo = qux
+
+        class zoo(foo):
+            def __init__(self, qux):
+                self.goo = qux
+
+        bar(42)
+        zoo(42) # This will raise an error because 'foo' is not defined in 'zoo'
+    ```
+    """
+
+    @classmethod
+    def __attrs__(cls):
+        """
+        Returns the attributes that are required for a subclass.
+        This method is used to define the attributes that must be present
+        in a subclass, ensuring that the class is properly configured.
+        """
+        return []
+
+    @classmethod
+    def __init_subclass__(cls, **kwargs):
+        """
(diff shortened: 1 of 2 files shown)


src/finch/util/abs_attr.py
```diff
@@ -0,0 +1,66 @@
...
+                for attr in base.__attrs__():
+                    # Only set if not already set, so the most base class wins
+                    if attr not in attrs:
+                        attrs[attr] = base
+        cls.__init__ = make_validate_attrs_init(cls, cls.__init__, attrs)
+
+    
+def make_validate_attrs_init(cls, init, attrs):
+    def __validate_attrs_init__(self, *args, **kwargs):
+        init(self, *args, **kwargs)
```
mtsokol, line 54: I think this can go after the docstring.

src/finch/util/abs_attr.py
```diff
@@ -0,0 +1,66 @@
+class AbsAttr():
```
mtsokol, line 1: I think in the comment it was `ReqAttrs`?

mtsokol (approved):
I reviewed it in your comment (thought that GitHub lost it 😄) and looks great!

src/finch/util/abs_attr.py
willow-ahrens: yes, I renamed it

src/finch/util/abs_attr.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init_subclass__(cls, **kwargs):
+        """
+        Initializes the subclass and collects the required attributes.
+        This method is called when a subclass is created, allowing it to
+        gather the attributes that must be present in a Finch model.
+        """
+        super().__init_subclass__(**kwargs)
+        attrs = {}
+        for base in reversed(cls.__mro__):
+            if hasattr(base, "__attrs__"):
```
hameerabbasi, line 44: I'd change this to `isinstance(base, AbsAttr)`

hameerabbasi (changes requested):
LGTM. Just the one nit.

willow-ahrens:
I'm abandoning this because it doesn't work with dataclasses and it differs from python community consensus. It's a tragedy that python dataclasses don't work with properties, maybe one day someone can fix that. for now, I'll use private attributes

PR #96, opened 2025-06-10
Add lazy interfaces mean, std, var

siyi-hu:
Add lazy statistical interfaces mean, std, var. Towards #3

### DIFF
--- a/src/finch/algebra/algebra.py
+++ b/src/finch/algebra/algebra.py
@@ -81,14 +81,16 @@ def query_property(obj: type | Hashable, attr: str, prop: str, *args) -> Any:
         AttributeError: If the property is not implemented for the given type.
     """
     if not isinstance(obj, type):
+        # Only catch TypeError for hashability check
         try:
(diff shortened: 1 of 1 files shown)


willow-ahrens (changes requested):
I think this may be missing std?

willow-ahrens:
It looks like this PR now depends on https://github.com/finch-tensor/finch-tensor-lite/issues/101

siyi-hu:
Got following assertion error when std function dealing with floating inputs ```a = np.array([[0, 0.618, 0, 0.001], [0, 0.01, 0, 0]])```. Any advice about the precision problem?

```
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference among violations: 6.9388939e-18
E           Max relative difference among violations: 1.66915753e-16
E            ACTUAL: array(0.041571)
E            DESIRED: array(0.041571)
```

PR #100, opened 2025-06-12
Wma/fix92

willow-ahrens:
Fixes 92 by embracing boilerplate and just doing the messy canonical python `@property @abstractmethod` approach

fixes https://github.com/finch-tensor/finch-tensor-lite/issues/92


hameerabbasi (approved):
LGTM; thanks for making this (rather painful) change.

PR #103, opened 2025-06-13
Manipulation functions (#26)

### LINKED ISSUE #26: Implement easy manipulation functions
Some functions in https://data-apis.org/array-api/2024.12/API_specification/manipulation_functions.html can be implemented using existing functions.

Consider the function `first(a, b) = a`

`broadcast_to(x, shape)` can be implemented as `elementwise(first, x, np,ndindex(shape))`

`broadcast` can be implemented with `broadcast_to`.

`moveaxis` can be implemented with `permute_dims`

Osmansiddiquer:
This branch aims to resolve #26 . Currently, I am trying to implement broadcast_to using elementwise(). This seems to work well
except for cases where 0-dimesnional arrays are involved. I create an empty tensor with the required shape: https://github.com/finch-tensor/finch-tensor-lite/blob/1b2f0bcc35405d6e461bce6952d1006cce695792/src/finch/interface/lazy.py#L873-L882
However, when I run the test using:
`poetry run pytest -k broadcast -vv -x`

The interpreter throws the error:
` ValueError: undefined tensor alias Alias(name='#A#63')`


willow-ahrens:
perhaps you could print the plan before it executes?

src/finch/interface/lazy.py
```diff
@@ -857,6 +898,80 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+    Args:
+        tensor: The lazy tensor to broadcast.
+        shape: The target shape to broadcast to.
+
+    Returns:
+        A new lazy tensor with the specified shape.
+    """
+    from .eager import EagerTensor
+
+    class NoneArray(EagerTensor):
```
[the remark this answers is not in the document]
Osmansiddiquer: it leads to circular references since eager.py imports everything from lazy.py

src/finch/interface/lazy.py
```diff
@@ -857,6 +898,80 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+
+        @property
+        def ndim(self):
+            return len(self.shape)
+
+        def __getitem__(self, idxs):
+            return None
+
+    if not is_broadcastable_directional(tensor.shape, shape):
+        # If the tensor is already broadcastable to the shape, return it as is
```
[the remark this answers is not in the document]
Osmansiddiquer: you're right. This is from a previous code. I generally clean these up in my last commit for the PR.

src/finch/interface/eager.py
```diff
@@ -14,9 +14,9 @@ def override_module(self):
 
     @property
     @abstractmethod
-    def ndim(self):
+    def ndim(self) -> int:
         """Number of dimensions of the tensor."""
-        ...
+        raise NotImplementedError("Subclasses must implement ndim.")
```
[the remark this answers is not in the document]
willow-ahrens: is this an issue of pylance not following the decorator correctly?

src/finch/interface/lazy.py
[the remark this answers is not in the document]
willow-ahrens: I see. So we can remove it?

src/finch/interface/lazy.py
```diff
@@ -457,6 +459,42 @@ def reduce(
     return LazyTensor(identify(data), shape, init, dtype)
 
 
+def _broadcast_shape(*args) -> tuple[int, ...]:
+    """
+    Computes the broadcasted shape for the given LazyTensor arguments,
+    following arrray_api broadcasting rules.
+    Raises ValueError if shapes are not broadcastable.
+    """
+    # Only support two arguments for now, as in the algorithm
```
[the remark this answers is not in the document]
willow-ahrens: I see. This is fine, since it's more correct.

src/finch/interface/lazy.py
```diff
@@ -457,6 +459,42 @@ def reduce(
     return LazyTensor(identify(data), shape, init, dtype)
 
 
+def _broadcast_shape(*args) -> tuple[int, ...]:
+    """
+    Computes the broadcasted shape for the given LazyTensor arguments,
+    following arrray_api broadcasting rules.
```
willow-ahrens, line 465: arrray -> array

src/finch/interface/lazy.py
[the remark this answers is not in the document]
willow-ahrens: Yes, I'll recommend then that for now, we can just define NoneArray without inheriting from EagerArray. Thus, NoneArray won't have any of the fancy operations defined on it, but it will be able to be used in Lazy expressions as a utility for us.

src/finch/interface/lazy.py
```diff
@@ -857,6 +898,80 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+            return None
+
+    if not is_broadcastable_directional(tensor.shape, shape):
+        # If the tensor is already broadcastable to the shape, return it as is
+        raise ValueError(
+            f"Tensor with shape {tensor.shape} is not broadcastable "
+            f"to the shape {shape}"
+        )
+    # elementwise does not support zero-dimensional tensors, so we
+    # handle that case separately
```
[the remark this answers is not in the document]
willow-ahrens: I don't think we necessarily need to return an immediate from elementwise. Since you've fixed the broadcasting, we can return a LazyTensor with a `MapJoin` over the empty arrays, since the dimensions would match.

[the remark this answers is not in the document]
willow-ahrens: I don't think we need a special case for that, what error have you hit in elementwise?

tests/test_interface.py
```diff
@@ -330,14 +330,17 @@ def test_matmul(a, b, a_wrap, b_wrap):
 
     result = finch.matmul(wa, wb)
     result_with_op = wa @ wb  # make sure the operator overload works too
+    result_with_np = np.matmul(wa, wb)
```
willow-ahrens, line 333: isn't `result_with_np` already computed with `expected`? I'm confused by this.

tests/test_interface.py
```diff
@@ -632,3 +635,115 @@ def test_scalar_coerce(x, func):
...
+        # 1-dim can be broadcast to 0-dim: (3,1) → (3, 0)
+        (np.array([1, 2, 3]).reshape(-1, 1), (3, 0)),
+        # 0-dim can be prepended: (3, 2) → (0, 3, 2)
+        (np.arange(6).reshape(3, 2), (0, 3, 2)),
+        # ——— INVALID CASES ———
+        # mismatched non-1 dim at end
+        (np.array([1, 2, 3]), (2, 2)),
+        # mismatched non-1 dim in middle
+        (np.ones((2, 3, 4)), (2, 5, 4)),
+        # boradcast on right side
```
willow-ahrens, line 670: boradcast -> broadcast

src/finch/interface/lazy.py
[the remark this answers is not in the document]
Osmansiddiquer: You're right, I hadn't tested it after fixing broadcasting logic. It works with 0-dims now

src/finch/interface/eager.py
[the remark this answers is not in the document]
willow-ahrens: Maybe the right approach is to return Any here

Osmansiddiquer:
This PR got delayed a lot so I had to squash with the main that was a lot of commits ahead. I have double checked that it doesn't overwrite any new changes. Other than that, the flatten function in lazy.py is stub for now. We might need an issue to track that

Osmansiddiquer:
I've added conftest.py. We can now declare global fixtures

src/finch/interface/lazy.py
```diff
@@ -532,6 +535,23 @@ def reduce(
...
+
+    Parameters:
+    --------------
+    - *args: Variable number of tuples representing shapes of LazyTensors.
+
+    Returns:
+    --------------
+    tuple[int, ...]: The broadcasted shape as a tuple of integers.
+    """
+    return np.broadcast_shapes(*args)
```
willow-ahrens, line 552: solid, but I'm worried that it converts the types of dimensions to np.int64. It's kindof critical that Finch is able to support infinitely large dimensions. Could you use the version of this you had last time where you followed array api exactly? I thought it was good, just needed a few tweaks

src/finch/interface/lazy.py
```diff
@@ -932,6 +952,197 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+
+def broadcast_arrays(*arrays: LazyTensor) -> tuple[LazyTensor, ...]:
+    """
+    Broadcasts one or more arrays against one another.
+    """
+    shape = _broadcast_shape(*(array.shape for array in arrays))
+    return tuple(broadcast_to(arr, shape) for arr in arrays)
+
+
+def concat(arrays: tuple | list, /, axis: int | None = 0) -> LazyTensor:
```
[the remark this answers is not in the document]
willow-ahrens: The reason i'm now pushing for this is that we have a stronger story for implementation this way with looplets, later this summer

src/finch/interface/lazy.py
```diff
@@ -932,6 +952,197 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+
+        # Initialize or accumulate result
+        result = placed_perm if result is None else result + placed_perm
+
+        cumulative_offset += size
+    assert result is not None
+    return result
+
+
+def flatten(x):
```
willow-ahrens, line 1073: Similar to Concat, could you define `SplitDimsTensor` and `CombineDimsTensor`? Then you can define flatten as a `CombineDimsTensor`, and we have a good shot of supporting reshape.

willow-ahrens: you don't need to support reshape now, i'd like to handle that myself.

src/finch/algebra/operator.py
```diff
@@ -116,3 +116,66 @@ def overwrite(x, y):
...
+    lambda op, a, condition: a,
+)
+
+
+def logical_or(a, b):
+    """
+    Returns the first truthy value between a and b.
+    If both are falsy, returns b.
+    """
+    return a or b
```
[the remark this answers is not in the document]
Osmansiddiquer: It's used in the concat function.
```python
placed = reduce(
            logical_or,
            elementwise(choose_if, permuted_tensor, I_i),
            axis=-2,
        )
```
But with the new implementation, we don't need it any longer

src/finch/interface/lazy.py
Osmansiddiquer: good catch.

src/finch/interface/lazy.py
```diff
@@ -959,6 +1002,419 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+    def __getitem__(self, idxs):
+        return None
+
+    @property
+    def shape(self):
+        return self._shape
+
+    @property
+    def format(self):
+        return LazyTensorFormat(
```
willow-ahrens, line 1019: make a separate format called "NoneTensorFormat".

src/finch/interface/lazy.py
```diff
@@ -959,6 +1002,419 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+        t = self.tensors[tn]
+        shifted_idx = idxs[self.concat_axis] - self.ps_sizes[tn]
+        result = t[
+            idxs[: self.concat_axis] + (shifted_idx,) + idxs[self.concat_axis + 1 :]
+        ]
+        return self.format.element_type(result)
+
+    @property
+    def format(self):
+        return LazyTensorFormat(
```
willow-ahrens, line 1118: make a separate format called "ConcatTensorFormat". It should contain  the formats of the tensors being concatted

src/finch/interface/lazy.py
```diff
@@ -959,6 +1002,419 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+            multiplier *= self.split_shape[i]
+
+        # Reconstruct the original indices
+        original_idxs = idxs[: self.axis] + (linear_idx,) + idxs[split_end:]
+
+        return self.tensor[original_idxs]
+
+    @property
+    def format(self):
+        return LazyTensorFormat(
```
willow-ahrens, line 1219: similarly for this tensor, make a `SplitTensorFormat`. It would be great if all of these new formats inherited from a shared `WrapperTensorFormat`.

src/finch/interface/lazy.py
```diff
@@ -959,6 +1002,419 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+
+    @property
+    def shape(self):
+        return self._shape
+
+    def asarray(self):
+        return self
+
+
+class CombineDimsTensor(Tensor):
```
willow-ahrens, line 1233: Same for this class, it needs its own format.

willow-ahrens (changes requested):
Really small request this time to simply add formats for each of the new tensor classes, so that we might be able to tell which kind of wrapper tensor it is based on the format, and so that we can eventually provide specialized implementations.

src/finch/interface/lazy.py
```diff
@@ -559,6 +564,44 @@ def reduce(
     return LazyTensor(identify(data), shape, init, dtype)
 
 
+def _broadcast_shape(*args: tuple[int, ...]) -> tuple[int, ...]:
```
willow-ahrens, line 567: We don't know if shape is int or not.

willow-ahrens (changes requested):
another quick ask: Don't type shapes as int

src/finch/interface/lazy.py
willow-ahrens: Bump

willow-ahrens (changes requested):

src/finch/interface/lazy.py
willow-ahrens: let's just type it as `tuple`

willow-ahrens (changes requested):
Last request: For the actuall wrapper formats, include the formats that they wrap, and compute their properties from the child format on-the-fly

src/finch/interface/lazy.py
```diff
@@ -959,6 +1003,525 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+    def element_type(self):
+        return self._element_type
+
+    @property
+    def shape_type(self):
+        return self._shape_type
+
+
+@dataclass(frozen=True)
+class WrapperTensorFormat(TensorFormat):
```
Osmansiddiquer, line 1031: shape types are hard to compute from just the wrapper tensor formats because we don't know the final shape of the end result. In cases like concat, the type of concat_axis shape can be expected to change to accommodate a larger size whereas in cases like split_dims, the target axis is split into smaller shape.

src/finch/interface/lazy.py
```diff
@@ -959,6 +1003,525 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+
+@dataclass(frozen=True)
+class ConcatTensorFormat(WrapperTensorFormat):
+    """
+    Tensor format for concatenated tensors.
+    Takes in a tuple of constituent formats, the shape type, and the concatenation axis.
+    Shape type is needed as it cannot be computed just from the constituent formats
+    """
+
+    _shape_type: tuple
```
Osmansiddiquer, line 1121: shape type must be computed and passed in by the ConcatTensor since only it knows about the shape of the concatenation

PR #105, opened 2025-06-13
Added Trigo functions, and bitwise_inverse,  fixed mateuz suggestion, changed overlay in lazy.py

yusheng036:
Added all of the trigonometric functions. I have tried implementing Willow suggestion in slack for the hyperbolic functions. Although it fulfilled some fill value issues, there were some edge cases that pytest wasnt passing so I went ahead with replacing all hyperbolic functions with numpy instead. There was RuntimeWarning in the implementation of numpy but the implementation was successful. I have also implemented Mateuz suggestion for the change in test_algebra.py to change values to boolean. I have also implemented bitwise_inverse that was supposed to be a part of my previous pr but I just included into the current one. I changed the overlay in lazy.py as addressed in previous meetings.

```
from finch import atan2
import numpy as np

y = np.array([ 1.0, 0.0, -1.0])
x = np.array([ 1.0, 1.0,  1.0])
print("atan2:", atan2(y, x))

```
Output:
atan2: [ 0.78539816  0.         -0.78539816]


src/finch/algebra/algebra.py
```diff
@@ -636,3 +635,23 @@ def sum_init_value(t):
...
 register_property(min, "__call__", "init_value", lambda op, arg: type_max(arg))
 register_property(max, "__call__", "init_value", lambda op, arg: type_min(arg))
+
+for trig_op in (
+    math.sin,
+    math.cos,
+    math.tan,
+    math.sinh,
+    math.cosh,
+    math.tanh,
```
willow-ahrens, line 645: I think we can use solely numpy, for consistency.

willow-ahrens: That said, I think it's okay to define properties for math functions we don't use, someone may want them later.

src/finch/interface/lazy.py
```diff
@@ -808,3 +856,55 @@ def vecdot(x1, x2, /, *, axis=-1) -> LazyTensor:
...
+def cosh(x) -> LazyTensor:
+    return elementwise(math.cosh, defer(x))
+
+
+def tan(x) -> LazyTensor:
+    return elementwise(math.tan, defer(x))
+
+
+def tanh(x) -> LazyTensor:
+    return elementwise(math.tanh, defer(x))
```
willow-ahrens, line 882: Let's use numpy for the above as well

willow-ahrens (changes requested):
This looks good! Only minor requests.

willow-ahrens (approved):
Thanks for another great PR!

PR #107, opened 2025-06-13
Add finch.asarray interface

### LINKED ISSUE #101: define finch.as_array to define promotion of scalar numbers to arrays
any time we expect an array argument in finch, we should query an as_array property which either wraps numbers in a Scalar wrapper type or returns arrays unmodified.

siyi-hu:
Add finch.asarray interface and corresponding support in defer to handle scalar type arguments. Fixes #101


src/finch/interface/lazy.py
```diff
@@ -179,6 +181,25 @@ def __bool__(self) -> bool:
...
+    Parameters:
+    ------------
+    - arr: The input array to be converted into a LazyTensor.
+
+    Returns:
+    ------------
+    out: array
+        an output array converted from input parameter ``arr``.
+    """
+    if isinstance(arr, Number | str | bytes | bytearray | memoryview):
```
willow-ahrens, line 197: Instead of an `if isinstance`, could we do a two step process where we first check for an `asarray` attribute, and if it doesn't exist we use `query_property(arr, "asarray", "__attr__")`?

Then we could do `register_property(Number, "asarray", "__attr__", lambda x: np.asarray(x))`.

src/finch/interface/lazy.py
```diff
@@ -179,6 +181,25 @@ def __bool__(self) -> bool:
...
+        return np.asarray(arr)
```
willow-ahrens, line 198: Instead of using `np.asarray`, could we define a class`finch.Scalar`?

it would look like `
@dataclass
class Scalar(EagerTensor):
    val:Any

    @property
    def ndims(self):
        return 0
    
    @property
    def element_type(self):
        return type(self.val)

willow-ahrens (changes requested):
A few comments: looks good so far!

src/finch/interface/lazy.py
```diff
@@ -194,6 +215,7 @@ def defer(arr) -> LazyTensor:
     """
     if isinstance(arr, LazyTensor):
         return arr
+    arr = asarray(arr)
```
willow-ahrens, line 218: @hameerabbasi do you see any reason to call asarray separately from defer?

hameerabbasi: I if asarray doesn't materialise; then no.

siyi-hu: I added this line because I think definitions like ```defer(1)```, ```defer(2.0)``` should be directly supported. This may not be necessary if we use ```finch.Scalar```.

willow-ahrens: Based on hameer's comment, I think this line is great as is. Let's leave it

src/finch/interface/lazy.py
siyi-hu: Tried to write several patterns but still not managed to understand where should we put Scalar type. More advice please?

willow-ahrens: Sure! The idea of the Scalar type is to implement `Scalar` as a 0-dimensional tensor type. So that when we call `asarray(1)` we get `scalar(1)` rather than `np.ndarray`, which might not be handled as efficiently.

hameerabbasi: Small note: `asarray` requires things to follow the standard interface of an array even if what we return is zero-dim. Keeping things synced across scalar and array may be tough; so instead what I would do is generate different/specialised code for 0D tensors instead of defining a different type with different semantics.

hameerabbasi: Small note: `asarray` requires things to follow the standard interface of an array even if what we return is zero-dim. Keeping things synced across scalar and array may be tough; so instead what I would do is generate different/specialised code for 0D tensors instead of defining a different type with different semantics.

willow-ahrens: of course, if we want to use the finch compiler stuff, we'll need to also define `unfurl, declare, freeze, and thaw`

src/finch/algebra/algebra.py
```diff
@@ -655,3 +674,9 @@ def sum_init_value(t):
     )
 
 register_property(np.atan2, "__call__", "return_type", lambda op, a, b: float)
+
+for t in StableNumber.__args__:
+    register_property(t, "asarray", "__attr__", lambda x: np.asarray(x))
```
willow-ahrens, line 679: wrap it in a `Scalar` instead

src/finch/algebra/algebra.py
```diff
@@ -655,3 +674,9 @@ def sum_init_value(t):
...
+
+register_property(str, "asarray", "__attr__", lambda x: np.asarray(x))
```
willow-ahrens, line 681: same here

src/finch/algebra/algebra.py
```diff
@@ -655,3 +674,9 @@ def sum_init_value(t):
...
+register_property(bytes, "asarray", "__attr__", lambda x: np.asarray(x))
```
willow-ahrens, line 682: same here

src/finch/interface/eager.py
```diff
@@ -184,6 +188,36 @@ def __bool__(self):
...
+        return type(self.val)
+
+    def __getitem__(self, idx):
+        return self.val
+
+    def asarray(self):
+        return np.asarray(self.val)
+
+
+def asarray(arr):
```
willow-ahrens, line 214: why are there two instances of asarray. use the earlier version of asarray in algebra

src/finch/algebra/algebra.py
willow-ahrens, line 682: you should define asarray for numpy arrays, eagertensor, and lazytensor

willow-ahrens (changes requested):
I'm not sure why you defined two versions of asarray

src/finch/interface/eager.py
siyi-hu: Moved asarray interface from algebra to both lazy and eager layer. Property registrations also performed in eager.

src/finch/interface/eager.py
```diff
@@ -184,6 +189,42 @@ def __bool__(self):
...
+        return type(self.val)
+
+    def __getitem__(self, idx):
+        return self.val
+
+
+register_property(Scalar, "asarray", "__attr__", lambda x: np.asarray(x))
+register_property(np.ndarray, "asarray", "__attr__", lambda x: x)
+register_property(EagerTensor, "asarray", "__attr__", lambda x: x)
+register_property(lazy.LazyTensor, "asarray", "__attr__", lambda x: x)
```
willow-ahrens, line 218: move this to lazy

src/finch/interface/eager.py
```diff
@@ -184,6 +189,42 @@ def __bool__(self):
...
+
+
+def asarray(arg):
```
willow-ahrens, line 221: no, delete this and import from lazy.

src/finch/interface/eager.py
```diff
@@ -184,6 +189,42 @@ def __bool__(self):
...
+    @property
+    def element_type(self):
+        return type(self.val)
+
+    def __getitem__(self, idx):
+        return self.val
+
+
+register_property(Scalar, "asarray", "__attr__", lambda x: np.asarray(x))
+register_property(np.ndarray, "asarray", "__attr__", lambda x: x)
```
willow-ahrens, line 216: move this line to lazy

src/finch/interface/eager.py
willow-ahrens, line 218: move this line to lazy

src/finch/interface/eager.py
```diff
@@ -184,6 +189,42 @@ def __bool__(self):
...
+
+    @property
+    def element_type(self):
+        return type(self.val)
+
+    def __getitem__(self, idx):
+        return self.val
+
+
+register_property(Scalar, "asarray", "__attr__", lambda x: np.asarray(x))
```
willow-ahrens, line 215: move this line to lazy

src/finch/interface/eager.py
willow-ahrens, line 221: delete this definition. instead, write
```
register_property(object, "asarray", "__attr__", lambda x: Scalar(x))
```

src/finch/interface/__init__.py
```diff
@@ -1,11 +1,13 @@
 from .eager import (
     EagerTensor,
+    Scalar,
     abs,
     acos,
     acosh,
     add,
     all,
     any,
+    asarray,
```
willow-ahrens, line 10: import this from lazy

willow-ahrens (changes requested):

willow-ahrens (approved):
LGTM

PR #113, opened 2025-06-24
Getting rid of some warnings

nmmull:
Some minor changes to get rid of warnings when running the test suite. Not sure everything here is reasonable, thought I'd check with a draft PR.

The thing I'm most unsure about is suppressing runtime warnings from numpy when testing, but this feels reasonable to me. My primary uncertainty: do you want to handle these warnings and/or test the behavior of the warnings themselves?

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,6 +3,7 @@ addopts = --cov-report term-missing --cov-report html --cov-report=xml --cov-rep
 filterwarnings =
     ignore::PendingDeprecationWarning
 testpaths =
-    finch
+    src/finch
+    tests
 junit_family=xunit2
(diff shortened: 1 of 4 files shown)


willow-ahrens (approved):
I'm okay ignoring these warnings for now, and making a github issue noting that we should check for the warnings in the future. It's the kind of correctness that is good to have, but I don't think it's make-or-break for our efforts right now. Eventually, we may be able to correctly handle e.g. NaN, Inf, etc. with strong guarantees, and this kind of thing would be part of that effort.

willow-ahrens:
https://github.com/finch-tensor/finch-tensor-lite/issues/115

PR #114, opened 2025-06-25
Wma/symbolic

### LINKED ISSUE #60: add struct unpacking for finch assembly
We need a way to unpack structs onto the stack in Finch Assembly when it's safe to assume we can make a copy of the fields.

Two Proposals:

1. We add two new nodes: a `Stack` variable namespace to signify the names of the unpacked vars, as well as `Unpack` operator. When we use the stack variable as anything other than a buffer, we would automatically repack into the original variable and return the value.
2. We just add a simple `GetData` operator which strips the pointer out of a buffer. This is less compelling as its more limited, but it could be simpler to conceptualize.

willow-ahrens:
fixes #60 and serves as a template for NotationCompiler.

My main question: Do we want to force buffers to be symbolic, rather than allowing them not to be? This would mean that buffers must always be unpacked before use, and might make buffer implementation easier to read. Currently, we allow buffers to be variables, or literals.

I'd appreciate people's reaction to the design, and a little feedback would be helpful as I make similar decisions in the FinchNotation compiler.

### DIFF
--- a/src/finch/codegen/c.py
+++ b/src/finch/codegen/c.py
@@ -588,9 +588,9 @@ class CContext(Context):
             case asm.Call(f, args):
                 assert isinstance(f, asm.Literal)
                 return c_function_call(f.val, self, *args)
-            #case asm.Slot(var_n, var_t) as ref:
+            # case asm.Slot(var_n, var_t) as ref:
             #    return self(self.deref(ref))
-            #case asm.Symbolic(obj, var_t) as ref:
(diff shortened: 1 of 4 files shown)


mtsokol:
I would advocate for a simplicity therefore my take is that forcing buffers to be only symbolic and making `unpack`ing and `repack`ing obligatory is a better option.
From reviewer's perspective it was easier to me to understand #116 and has more natural/predictable code changes.

PR #116, opened 2025-06-25
Wma/force symbolic

willow-ahrens:
This is similar to https://github.com/finch-tensor/finch-tensor-lite/pull/114, but more constrained.

Here, we force the unpacking and packing of buffers in order to use them. It makes things a little bit easier to read because we don't need to support both cases everywhere


src/finch/codegen/c.py
```diff
@@ -516,17 +528,41 @@ def block(self) -> "CContext":
...
+                raise KeyError(f"Slot {var_n} not found in context")
+            case asm.Stack(_, _):
+                return node
+            case _:
+                raise ValueError(f"Expected Slot or Stack, got: {type(node)}")
+
     def emit(self):
         return "\n".join([*self.preamble, *self.epilogue])
 
+    def cache(self, name, val):
```
mtsokol, line 557: Is `cache` still needed? It's used in only one place.

willow-ahrens: I'd like to keep it around, it comes in handy sometimes.

mtsokol (approved):

Issue #99, opened 2025-06-11
Implement NotationCompiler


willow-ahrens:
This is the pass which compiles FinchLogic to FinchNotation. In Finch.jl, it was called LogicCompiler. A challenge with the LogicCompiler is that the julia version was written to target the "fancy syntax" version of FinchNotation, not the parsed ast. In https://github.com/finch-tensor/Finch.jl/blob/main/src/scheduler/LogicCompiler.jl, you can see that it's writing `for i = ...`, not `loop(...)`

PR #117, opened 2025-06-26
LogicCompiler for Finch Logic

mtsokol:
Addresses #99 

This PR introduces `LogicCompiler` for lowering Finch Logic to Finch Notation. It moves physical tensors out of the plan and replaces them with variables. The plan becomes a module with a single function which accepts tensors as arguments.


src/finch/autoschedule/compiler.py
```diff
@@ -1,133 +1,81 @@
...
     Table,
     Value,
 )
-from ._utils import intersect, with_subsequence
-
-T = TypeVar("T", bound="LogicNode")
-
-
-@overload
-def compute_structure(
```
mtsokol, line 27: I removed `compute_structure` stubs&function from here. I added them some time ago in the initial draft but it's not used.

src/finch/autoschedule/compiler.py
```diff
@@ -136,52 +84,234 @@ class LogicLowerer:
...
+                            ),
+                            rhs,
+                        ),
+                    )
+                )
+                for idx in idxs_2:
+                    if idx in rhs_idxs:
+                        body = ntn.Loop(
+                            ntn.Variable(idx.name, int),
+                            ntn.Variable(f"{idx.name}_size", int),
```
mtsokol, line 164: I adopted a convention `<name>_size` for extend variables, so it ends as: `loop(i, i_size, body)`.

src/finch/finch_notation/nodes.py
```diff
@@ -405,9 +425,10 @@ def children(self):
         return [self.name, *self.args, self.body]
 
     @classmethod
-    def from_children(cls, name, *args, body):
+    def from_children(cls, name, *args_body):
```
mtsokol, line 428: We can't have standard parameter after vararg here.

tests/test_scheduler.py
```diff
@@ -673,6 +673,56 @@ def test_propagate_map_queries_backward():
     assert result == expected
 
 
+def test_materialize_squeeze_expand_productions():
```
mtsokol, line 676: I only moved the test up to make the file more coherent.

tests/test_scheduler.py
```diff
@@ -684,16 +734,16 @@ def test_scheduler_e2e_matmul(a, b):
...
     plan = Plan(
-        [
+        (
             Query(Alias("A"), Table(Literal(a), (i, k))),
             Query(Alias("B"), Table(Literal(b), (k, j))),
             Query(Alias("AB"), MapJoin(Literal(mul), (Alias("A"), Alias("B")))),
             Query(
                 Alias("C"),
-                Reorder(Aggregate(Literal(add), Literal(0), Alias("AB"), (k,)), (i, j)),
+                Aggregate(Literal(add), Literal(0), Alias("AB"), (k,)),
```
mtsokol, line 743: Reorder prevents "propagate queries backwards" optimization.

mtsokol: Solved in https://github.com/finch-tensor/finch-tensor-lite/pull/124 - I can revert that change.

src/finch/finch_notation/nodes.py
hameerabbasi: How about changing the order? `body, *args`

src/finch/autoschedule/compiler.py
willow-ahrens: we need the function eventually for caching compiled kernels.

willow-ahrens: We need compute_structure to eventually normalize the program so we can hash it and cache the compiled kernel.

src/finch/autoschedule/compiler.py
willow-ahrens: Could we be more precise about promoting loop indices to the correct loop index type here? We need to promote the integer indices. I think shape_type of each tensor gives the index type of each arg, but we need to promote here, same as osman's PR where we implement broadcast but here we can use index names to determine which indices go where.

src/finch/autoschedule/compiler.py
```diff
@@ -136,52 +88,247 @@ class LogicLowerer:
...
+
+        def __call__(self, obj):
+            return np.ndarray
+
+    match ex:
+        case Plan(bodies):
+            return Plan(tuple(_format_queries(body, bindings) for body in bodies))
+        case Query(lhs, rhs) if not isinstance(rhs, Reformat | Table):
+            assert isinstance(rhs, LogicExpression)
+            rep = SuitableRep(bindings)(rhs)
```
willow-ahrens, line 243: Is SuitableRep just a stub for now?

src/finch/autoschedule/compiler.py
```diff
@@ -136,52 +88,247 @@ class LogicLowerer:
...
+
+def initwrite(*args):  # TODO: figure out the implementation
+    raise NotImplementedError
+
+
+# TODO: replace with appropriate Tensor class
+_TensorType = np.ndarray
+
+
+def record_tables(
```
willow-ahrens, line 262: This is essentially what "defer_tables" was in Finch.jl. I'm glad for the rename and redesign here.

src/finch/autoschedule/compiler.py
```diff
@@ -136,52 +88,247 @@ class LogicLowerer:
...
+
+    def rule_0(node):
+        match node:
+            case Query(Alias(name) as alias, Table(Literal(val), fields)):
+                table_var = ntn.Variable(name, type(val))
+                table_vars[alias] = table_var
+                tables[alias] = val
+                for idx, field in enumerate(fields):
+                    assert isinstance(field, Field)
+                    dim_size_var = ntn.Variable(f"{field.name}_size", int)
```
willow-ahrens, line 290: Again, dimensions aren't always int

willow-ahrens (changes requested):
This looks good to me! I think the main request is to type the loop variables based on the appropriate type of the tensors. Same request when we're computing loop sizes.

src/finch/autoschedule/compiler.py
mtsokol: Sure! Reverted.

PR #123, opened 2025-07-01
Implemented  logarithm and logical functions

yusheng036:
For the logical implementation, I mirrored the bitwise functions counterparts.  I added some test function in test_algebra.py for init_values. I have also added the logarithm functions. 

```
from finch import logical_and
import numpy as np

a = np.array([[True, False,  True],
              [False, True, False]])
b = np.array([[True,  True, False],
              [False, False, True]])

print("logical_and:", logical_and(a, b))
```
Output: 
logical_and: [[ True False False]
 [False False False]]


hameerabbasi (commented):
Minor suggestion

PR #124, opened 2025-07-01
Use concordant merge for fields in `MapJoin` processing

mtsokol:
Hi @willow-ahrens,

I'm going back with the issue I described earlier with SDDMM schedule optimization. I managed to solve it and now the final schedule matches Finch.jl counterpart and is exactly as need, regardless of `Relabel`/`Reorder` nesting.

The culprit arose from the fact that `propagate_map_queries_backward` used non-concordant merge of Fields in `Reorder(Aggregate())` rule and that `MapJoin.fields` returned non-concordant merge of Fields of underling `args` (Turs out we had a TODO for it).

To give an example: if we have a `Reorder(Agg(MapJoin((F0, F1), (F1, F2)), F1), (F0, F2))` the original transformation would do `Agg(Reorder(MapJoin((F0, F1), (F1, F2)), (F0, F2, F1)), F1)`. `(F0, F2, F1)` and `(F1, F2)` aren't concordant which prevented Reorder from being removed and caused those transpose intermediaries I reported. 

With `merge_concordant(...)` the order of the merged list is concordant, here `(F0, F1, F2)`, assuming args are concord (otherwise falls back to previous merging). WDYT?

I think this could also apply to Finch.jl.

### DIFF
--- a/src/finch/finch_logic/_utils.py
+++ b/src/finch/finch_logic/_utils.py
@@ -1,0 +1,27 @@
+from collections.abc import Iterable
+
+from .nodes import Field
+
+
+class NonConcordantLists(Exception):
+    pass
+
+
+def merge_concordant(args: Iterable[Iterable[Field]]) -> list[Field]:
+    merge_list: list[Field] = []
+    visited: set[Field] = set()
+
+    for arg in args:
+        idx = 0
+        for f in arg:
+            if f not in visited:
+                visited.add(f)
+                merge_list.insert(idx, f)
+                idx += 1
+            else:
+                next_idx = merge_list.index(f)
+                if next_idx < idx:
+                    raise NonConcordantLists
+                idx = next_idx + 1
+
+    return merge_list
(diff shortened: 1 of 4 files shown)


src/finch/finch_logic/_utils.py
```diff
@@ -0,0 +1,27 @@
+from collections.abc import Iterable
+
+from .nodes import Field
+
+
+class NonConcordantLists(Exception):
+    pass
+
+
+def merge_concordant(args: Iterable[Iterable[Field]]) -> list[Field]:
```
mtsokol, line 10: Here we merge all `args` lists so that the merged lists retains the order of all constituent lists. Raises an error if inputs aren't concordant.

willow-ahrens:
In finch.jp, we process the fields at some point to make them concordant, I don't think we have that pass in this codebase. We should add that pass, but this is a nicer default behavior.

PR #125, opened 2025-07-02
Fix for bad query_property error - issue #102

### LINKED ISSUE #102: query_property gives a bad error when there is an TypeError in the query_fn
Consider this code:

```python
from algebra import register_property, query_property
def first(*args):
    return args[0] if args else None

def some_func(args):
    raise TypeError("Bad function")

register_property(first, "__call__", "return_type", some_func)
query_property(first, "__call__", "return_type", 1, 2, 3)
```

This will give the error:
```
function '<function first at 0x000001B7B154BEC0>' has no property 'return_type'. Hint: ...
```
I came across the error when I was passing incorrect number of parameters to a function inside my lambda expression while using register_property. Python uses `TypeError` for a number of different errors so it may be misleading to blanket except `TypeError` in this case:

https://github.com/finch-tensor/finch-tensor-lite/blob/517ce3df1ffa1fa192a97706bd5887ef58707fd1/src/finch/algebra/algebra.py#L87-L91

Thank you for looking into this

Osmansiddiquer:
Attempts to fix #102 
Overview: 
This pull request refines error handling in the `query_property` function within `src/finch/algebra/algebra.py`. The change ensures that `TypeError` is caught only during the hashability check, improving clarity and reducing unnecessary exception handling.

Error handling improvement:

* [`src/finch/algebra/algebra.py`](diffhunk://#diff-c16e3d6420f474c578e1330d5427fcfd763e84d817fccc1a308f5743c9bc1c39R84-L90): Updated the `query_property` function to isolate the `TypeError` exception to the hashability check, ensuring that the exception handling is more precise and does not interfere with subsequent logic.

#102 had been lingering around for a while. This fixes the bad case given in there.

### DIFF
--- a/src/finch/algebra/algebra.py
+++ b/src/finch/algebra/algebra.py
@@ -81,14 +81,16 @@ def query_property(obj: type | Hashable, attr: str, prop: str, *args) -> Any:
         AttributeError: If the property is not implemented for the given type.
     """
     if not isinstance(obj, type):
+        # Only catch TypeError for hashability check
         try:
             hash(obj)
+        except TypeError:
+            t = type(obj)
+        else:
             query_fn = _properties.get((obj, attr, prop))
             if query_fn is not None:
                 return query_fn(obj, *args)
-        except TypeError:
-            pass
-        t = type(obj)
+            t = type(obj)
     else:
         t = obj
 


hameerabbasi (commented):
Suggestion on better resolution.

willow-ahrens (commented):
Lgtm!

PR #132, opened 2025-07-21
Add argmin and  argmax to lazy reduction interface

siyi-hu:
Add argmin and  argmax to lazy reduction interface. Towards #3


src/finch/algebra/operator.py
```diff
@@ -27,6 +29,28 @@ def promote_max(a, b):
...
+class Pair:
+    value: Any
+    index: int
+
+    def __init__(self, value, index):
+        self.value = value
+        self.index = index
+
+
+algebra.register_property(Pair, "asarray", "__attr__", lambda x: x)
```
willow-ahrens, line 41: Should return Scalar(x)

src/finch/algebra/operator.py
```diff
@@ -27,6 +29,28 @@ def promote_max(a, b):
     return max(cast(a), cast(b))
 
 
+class Pair:
```
willow-ahrens, line 32: Should define a format which includes both value format and index format

src/finch/algebra/operator.py
```diff
@@ -147,3 +171,10 @@ def identity(x):
     "return_type",
     lambda op, x: x,
 )
+
+
+algebra.register_property(minby, "__call__", "return_type", lambda op, a, b: int)
+algebra.register_property(maxby, "__call__", "return_type", lambda op, a, b: int)
+
+algebra.register_property(minby, "__call__", "init_value", lambda op, arg: 0)
```
willow-ahrens, line 179: minby can't return 0 as init value, it should return `(-Inf, 0)`.

src/finch/algebra/operator.py
```diff
@@ -147,3 +171,10 @@ def identity(x):
...
+algebra.register_property(maxby, "__call__", "init_value", lambda op, arg: 0)
```
willow-ahrens, line 180: same here

src/finch/interface/lazy.py
```diff
@@ -1660,3 +1684,29 @@ def std(
...
+
+def argmin(x, /, *, axis: int | None = None, keepdims: bool = False, init=None):
+    """
+    Returns the indices of the minimum values in input array ``x`` along given axis.
+    """
+    if isinstance(axis, tuple):
+        raise ValueError("Type of axis should is only allowed to be int or None.")
+
+    indices = np.indices(x.shape)
+    paired_tensor = pairwise_indices(x, indices)
```
willow-ahrens, line 1697: just call `elementwise(Pair, values, indices)`. no need for a complex custom function.

willow-ahrens (changes requested):
left my initial thoughts. Good start!

src/finch/interface/lazy.py
```diff
@@ -1660,3 +1685,63 @@ def std(
...
+def argmin(x, /, *, axis: int | None = None, keepdims: bool = False, init=None):
+    """
+    Returns the indices of the minimum values in input array ``x`` along given axis.
+    """
+    if isinstance(axis, tuple):
+        raise ValueError("Type of axis should is only allowed to be int or None.")
+
+    x = defer(x)
+    computed_x = _compute(x)
+    if hasattr(computed_x, "to_numpy") and callable(computed_x.to_numpy):
```
willow-ahrens, line 1699: You shouldn't need to compute the arrays. This should be a fusable operation. the whole operation should be one line, as so:

`elementwise(first, reduce(minby, elementwise(Pair, A, IndicesTensor(A.size))))`.

Construct a lazy IndicesTensor using Osman's approach, like ConcatTensor.

willow-ahrens (changes requested):

src/finch/algebra/operator.py
hameerabbasi: Shouldn't this one be dependent on the datatype? Integer data types don't support infinity. The following code works for me:

```python
from typing import Literal
import numpy as np

def get_dt_bounds(dt: np.dtype, kind: Literal["min", "max"]) -> np.generic:
    dt = np.dtype(dt)
    if kind not in {"min", "max"}:
        raise ValueError(
            "Only `kind='min'` or `kind='max'` are accepted, found "
            f"`{kind=}`."
        )
    if np.isdtype(dt, kind="integral"):
        return dt.type(getattr(np.iinfo(dt), kind))

    if np.isdtype(dt, kind="real floating"):
        return dt.type(np.inf if kind == "max" else -np.inf)

    raise TypeError("Only complex and real floating value types")
```

siyi-hu:
Revised to use IndicesTensorFormat for pairwised tensor data for argmin/argmax. Several bugs remained to be fixed.

PR #134, opened 2025-07-23
Add `eval`uable reprs for IRs

mtsokol:
Hi @willow-ahrens,

`eval`uable reprs are added in this PR for all three IRs, with round-trip tests you suggested. `LiteralRepr` is used to aid Literal, Variable, and Value nodes - I wonder if `LeafRepr` would fit here more, WDYT?

### DIFF
--- a/src/finch/symbolic/term.py
+++ b/src/finch/symbolic/term.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Iterator
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass
 from inspect import isbuiltin, isclass, isfunction
 from typing import Any, Self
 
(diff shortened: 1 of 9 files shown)


src/finch/symbolic/term.py
```diff
@@ -71,6 +72,30 @@ def children(self) -> list[Term]:
         ...
 
 
+@dataclass(frozen=True, eq=True)
+class LiteralRepr:
```
mtsokol, line 76: Is `LeafRepr` better? Or `EvaluableRepr`?

willow-ahrens: I think it would be better as a function than a base class, right? Since we may want to use this logic on formats, or fields other than val sometimes, for example.

willow-ahrens (commented):
This is great, thanks! Should we refactor LiteralRepr into a simple function that we can call on any value?

src/finch/symbolic/term.py
mtsokol: Can you give an example of Field or Format that wouldn't fit here?

willow-ahrens: I think you inferred that I was asking for a bigger change than I was. What I mean is more that I'd like the logic of LiteralRepr to be put into a function `literal_repr`, and classes which inherit from LiteralRepr to instead define 
```
    def __repr__(self):
        return literal_repr(("val", self.val))
```

The reason I thought to push for this was that automatically traversing `__dir__` fields feels hacky.

mtsokol: Ah, right. Done! I changed the `LiteralRepr` class to a function.

PR #135, opened 2025-07-23
Introduces regression testing to compare with reference outputs.

### LINKED ISSUE #111: Compare to reference output in test suite, add option to regenerate.
See finch.jl test functionality. Is this an option in pytest? Is there a simple python package that does this?

Osmansiddiquer:
Aims to resolve #111. 
## Summary:
This pull request introduces regression testing infrastructure to support testing compiler outputs and adds fixtures, hooks, and example tests to streamline regression testing for non-deterministic outputs. Additionally, it updates configuration files to exclude regression references from linting and formatting.

### Regression Testing Infrastructure:

- **`tests/conftest.py`:** Added regression fixtures, helpers, and hooks, including `lazy_datadir`, `original_datadir`, `preserve_obtained`, `pytest_runtest_makereport`, and `program_regression`. These enhance regression testing by creating unique directories for test references, enabling cleanup of obtained files, and enabling support for non-deterministic output comparison. [[1]](diffhunk://#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R1-R6) [[2]](diffhunk://#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R29-R131)
- **`tests/references/README.md`:** Documented regression testing setup, including usage of `pytest-regressions`, custom fixtures like `program_regression` for handling non-deterministic outputs, and cleanup mechanisms for obtained files.

### Configuration Updates:

- **`.pre-commit-config.yaml`:** Excluded `tests/references` from linting and formatting hooks to prevent unintended modifications to regression reference files.
- **`pyproject.toml`:** Updated `tool.ruff` configuration to exclude `tests/references` from linting and added `pytest-regressions` as a test dependency.

### Example Tests and Reference Files:

- **`tests/test_regressions.py`:** Added example tests (`test_c_program`, `test_file_regression`, `test_tree_regression`) to demonstrate regression testing for C programs, file content, and tree structures using the new fixtures.
- **Regression Reference Files:** Added sample reference files for testing, including `test_c_program.c`, `test_file_regression.txt`, and `test_tree_regression.txt`. These files serve as expected outputs for the example tests. [[1]](diffhunk://#diff-c703e2d213d1eda57b39d4aed74fdd36f6089b7da8faf866d95d7a0b3d5ef2a2R1) [[2]](diffhunk://#diff-c6d95d864dadefb149d94497876d488442fb11dd7ebb4455ec1cb367274eb7c1R1) [[3]](diffhunk://#diff-20eb12e39c724d789297d689b4b53f95b763e30b860705ab27b5d5c3c08cbf3eR1-R77)

## Author comments: 
- This PR does not modify any compiler code to enable emission of code directly. Currently tested with output of Logic Compiler.
- Since this has not been thoroughly tested with all the compiler outputs, it is expected that some modifications to the program_regression fixture will be needed. I've documented the new fixtures and automatic cleanup mechanism in `conftest.py` as well as `README.md` and I hope extending this won't be too much of a hassle.
- Feel free to leave comments about how you feel about some of the design choices made as well as naming of fixtures.
-
Thank you for looking into this.


tests/conftest.py
```diff
@@ -20,3 +26,106 @@ def _random_wrapper_applier(arrays, wrapper):
...
+    memory addresses in program trees.
+    """
+
+    def _program_regression(program, formatter=pprint.pformat, extension=".txt"):
+        """
+        Compares the program with the regression fixture.
+        Args:
+            program: The program to compare.
+            formatter: Optional formatter function to convert the AST to a string.
+        """
```
Osmansiddiquer, line 123: ### The reasoning behind introducing `program_regression` fixture: 
We want to support as many programs as possible. This includes text based C programs as well as tree based programs like Finch Notation programs that result from lowering by Logic Compiler. Problem with just comparing pretty string rep of the tree programs is that it contains memory addresses which change every time. This is the non-determinism I refer to in the PR description.

Solution 1: Define a custom __repr__ for `Term` which seems to be the base class for all nodes. Downside is that most nodes are dataclasses, so they define their own __repr__ manually setting `repr = False` in the dataclass decorator for each node is tedious and error-prone.

Solution 2: Replace all memory addresses manually in the string representation of the tree. The downside is added computation but the approach is extensible. We can extend this to accommodate any other sources of non-determinism in the tree representation. Also, this avoids adding any complexity when implementing tree nodes.

Note: Pickling the tree object is not an option because it will give us an unreadable binary representation.

Osmansiddiquer: >Before passing my judgement I would like so see one output that you have in mind. Can you provide an example of a test case with memory addresses?

Thanks for the detailed response. Here's an example of the output I'm referring to, from `the test_logic_compiler` test. This is a sample from the pretty printed version of the finch notation output for the plan in that test. You can see one instance of a memory address popping up here:
```text
body=Block(bodies=(Assign(lhs=Variable(name=':i0_size', type_=<class 'int'>),
                                               rhs=Call(op=Literal(val=<function dimension at 0x000002B38D0E2CA0>),
                                                             args=(Variable(name=':A0',
                                                             type_=<class 'numpy.ndarray'>),
                                                             Literal(val=0)))),
                                  Assign(lhs=Variable(name=':i1_size',
                                          type_=<class 'int'>),
```
Currently, the `program_regression` fixture includes a regex rule that detects hex memory addresses and replaces them with `...`. So the above gets normalized to:
```text
body=Block(bodies=(Assign(lhs=Variable(name=':i0_size', type_=<class 'int'>),
                                               rhs=Call(op=Literal(val=<function dimension at ...>),
                                                             args=(Variable(name=':A0',
                                                             type_=<class 'numpy.ndarray'>),
                                                             Literal(val=0)))),
                                  Assign(lhs=Variable(name=':i1_size',
                                          type_=<class 'int'>),
```
You can look at the `test_tree_regression` for how `program_regression` fixture is used in this usecase.

---------------------------
> All in all, in our case those memory addresses in input files should be matched with aliases, and reference output should already have aliases there, via some notation [[MY_ALIAS:0x[0-9a-fA-F]+]] - that an ideal case. And ideally there should be a pip installable tool for that - I don't know how much effort would be in implementing this kind of parsing ourselves.


Using something like `Filecheck` is definitely a more powerful and flexible approach to compare textual output with expected patterns. However, I'm not sure how well it would generalize to AST-like output while also maintaining readability of expected output. Another potential downside is that it would require us to manually write these checks into the reference files (?). 
From what I understand, our current goal with the `program_regression` fixture is to simplify test writing by automatically updating reference outputs and reducing the burden of explicitly writing expected results, similar in spirit to the julia version. Compared to `Filecheck`, I think the current approach with fixtures focuses more on ease of use and automation by centralizing non-determinism handling (like memory addresses) inside the fixture, so test authors don't have to write expected results or manage regex patterns themselves. 
That said, I agree that handling regex logic within the fixture is not be universally robust across all program outputs since the same rules aren't applicable everywhere. Whether we continue extending the regex logic within the fixture or shift toward a more explicit matcher like `FileCheck` is really an open design decision. Since I don't have deep experience with compiler test infrastructures, I'm not confident in picking one approach over the other yet. Possibly @willow-ahrens can weigh in on what direction would be best.

mtsokol: Ah, so it's a hex of a function in the AST.

I'm happy with just replacing them with a placeholder as you do it here, so I think the current approach is sufficient.

mtsokol (approved):
I'm new to pytest-regressions but the workflow here with generating test cases and interacting with them is sound to me.

tests/conftest.py
hameerabbasi: > (FileCheck is LLVM oriented, I don't know if it's also a general tool for output matching).

It is, one can run arbitary commands and match the stdout.

> Ah, so it's a hex of a function in the AST.

One approach is to replace all function references with `<function actual_name at 0xaddress_placeholder>` -- that seems fine to me. Another option is to manually override the repr of anything that holds a function to be more deterministic.

tests/conftest.py
```diff
@@ -20,3 +26,106 @@ def _random_wrapper_applier(arrays, wrapper):
...
+def original_datadir(lazy_datadir) -> Path:
+    return lazy_datadir
+
+
+@pytest.fixture
+def preserve_obtained(request):
+    """
+    Fixture to preserve obtained files after the test run.
+    """
+    request.node.preserve_obtained = True
```
hameerabbasi, line 72: We should be careful we don't leave other files hanging around in tests.

tests/conftest.py
```diff
@@ -20,3 +26,106 @@ def _random_wrapper_applier(arrays, wrapper):
...
+        Compares the program with the regression fixture.
+        Args:
+            program: The program to compare.
+            formatter: Optional formatter function to convert the AST to a string.
+        """
+        if not isinstance(program, str):
+            program = formatter(program)
+        # Replace memory addresses with ...
+        file_regression.check(
+            re.sub(r"0x[0-9a-fA-F]+", "...", program), extension=extension
```
hameerabbasi, line 128: Two comments:

1. I would turn this into an `re.compile` in global scope and then `compiled_re.sub(...)`
2. Can we make this more targeted to specific patters we see? We may not want to replace ALL hex numbers.

hameerabbasi (changes requested):
Just one change.

Osmansiddiquer:
I've added option for test-specific substitutions, and we now store compiled patterns in global scope.

tests/conftest.py
```diff
@@ -20,3 +26,131 @@ def _random_wrapper_applier(arrays, wrapper):
...
+def lazy_datadir(request) -> Path:
+    """
+    Creates a unique per-test directory for reference files.
+    E.g., tests/references/<module_name>/<test_name>/
+    """
+    # test_module will be something like "tests.test_logic_compiler"
+    # We don't want the "tests." part
+    module_parts = request.module.__name__.split(".")
+    if module_parts[0] == "tests" and len(module_parts) > 1:
+        test_module = "/".join(module_parts[1:])
```
willow-ahrens, line 48: This should be done with pathlib.path. Could you also add pathlib.path to the style guide (saying we use that library instead of os.path)?

tests/conftest.py
willow-ahrens: I am kindof okay with replacing the memory addresses, so long as we normalize them so that the same addresses are the same and different addresses are different, rather than replace them all with the same string.

I am even more in favor of just defining a function or method to print objects reproducibly. So that as we print the AST, we keep track of objects which don't have an overloaded print function and use our own generic fallback that does normalization just in time. I'd prefer we just work this into mateusz's printer that we already have going on.

willow-ahrens: In short, I prefer Hameer's last comment, especially if we can configure the printer to make same addresses same and different addresses different.

willow-ahrens (changes requested):
Overall, looks good. At a high level, I would like to use the pretty printer that mateusz developed, and add functions to print generics that preserve sameness and difference of addresses while keeping things reproducible. (e.g. <generic_function_or_object at 0x1>, <generic_function_or_object at 0x2>, )

tests/test_regressions/test_regex_substitution_patterns.txt
```diff
@@ -0,0 +1,7 @@
+ProcessResult(
+  objects=[
+    Object(name='...', id=..., func=<function process at 0x7f8c2c0d1e50>),
+    Object(name='...', id=..., func=<function validate at 0x7f8c2c0d2a80>)
```
willow-ahrens, line 4: It looks like this is still printing addresses

tests/test_regressions.py
```diff
@@ -0,0 +1,132 @@
...
+        "    Object(name='obj2', id=67890, func=<function validate at 0x7f8c2c0d2a80>)"
+        "\n"
+        "  ],\n"
+        "  status='complete'\n"
+        ")"
+    )
+
+    substitutions = {
+        r"Object\(name='[^']+', id=\d+,": r"Object(name='...', id=...,",
+        r"status='[^']+'": r"status='...'",
```
willow-ahrens, line 129: instead of regexing the object addresses, it would be better to handle the identities of objects in a separate step before the printer. Here's what I mean:
Define a function `normalize(prgm)` which uses `PostWalk` to replace literals that hold objects which would normally call the default `object` repr with a new type `NormalizedObject(id, type).` To compute the ID, we use a dictionary mapping objects to normalized objects, and when an object doesn't exist we set the id to the length of the dictionary plus one and add the object.

willow-ahrens (changes requested):
a small request to test object identity better. Does this proposal make sense?

tests/test_regressions.py
Osmansiddiquer: Got it. Wanted to ask if it we should keep regex substitutions around or completely remove the option to avoid technical debt

willow-ahrens: I wouldn't be opposed if you wanted to remove them completely. I don't think they will be needed in this PR.

Osmansiddiquer:
Is this complete, or do we need to add support for preserving the diff. of addresses while keeping it reproducible? (e.g. <generic_function_or_object at 0x1>, <generic_function_or_object at 0x2>, ) at the moment?

willow-ahrens (approved):

PR #136, opened 2025-07-26
First implementation of galley tensor statistics

yusheng036:
This is the first Galley implementation. Added TensorDef and DenseStats, with all tests in test_galley.py now passing.

These are the following files:
1) tensor_stats.py defines the TensorStats abstract base class.
2) tensor_def.py encapsulates a tensor’s metadata (axes, dimensions, fill value).
3) dense_stat.py provides a concrete DenseStats implementation for estimating and combining non‑fill values.


kylebd99 (approved):
Overall it looks great!  I made a few small changes:

1. To conform to the CONTRIBUTING.md, I adjusted the getters/setters to use the @property interface. 
2. If a concrete class instance defines __init__, then the abstract class __init__ isn't used. So, I've removed the DenseStats implementation, and it now just calls the from_tensor method to add any sub-class specific members.
3. I changed from_tensor to return a None because it should only be used in the TensorStats __init__ where it does not need a return value.

PR #140, opened 2025-08-01
Add printers for IRs

mtsokol:
Hi @willow-ahrens,

This PR introduces three printers (one per IR). It's rather dull (and as simple as I could make it) repetition of existing interpreters - only evaluating to string representation. It might seem as a bit of duplication but I don't see a point in inventing an abstraction here or moving specific `case` statements to a function. Each pattern matching uses the simplest string representation there is. 

Take a look at the last test file in the changes - now matmul for Finch Notation can be printed as:

```
def matmul(C: ndarray, A: ndarray, B: ndarray) -> ndarray:
  m: int64 = dimension(A, 0)
  n: int64 = dimension(B, 1)
  p: int64 = dimension(A, 1)
  C: ndarray = declare(C, 0.0, add, ['m', 'n'])
  loop(i, m):
      loop(j, n):
          loop(k, p):
              a_ik: float64 = unwrap(read(A, ['i', 'k']))
              b_kj: float64 = unwrap(read(B, ['k', 'j']))
              c_ij: float64 = mul(a_ik, b_kj)
              increment(update(C, ['i', 'j'], add), c_ij)
  C: ndarray = freeze(C, add)
  return C
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
     - id: no-commit-to-branch
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.12.4
+  rev: v0.12.7
   hooks:
     - id: ruff-check
       args: [--fix, --config=pyproject.toml]


willow-ahrens (approved):
Thank you Mateusz. With this I can at least read the code from the other PR. This is super helpful, and written clearly.

PR #148, opened 2025-08-13
Add debug statement for finch_assembly

### LINKED ISSUE #61: Add debug statement in FinchAssembly
For debugging, a "Print" or "debug" statement would be nice. I think it would just take a message argument as well as a list of expressions, and print the message as well as the expressions, together with the expression values.

siyi-hu:
Add a "debug" function along with logger.py in finch_assembly module to extract and print complete data structure of any input variable for debugging purpose. fixes #61.


willow-ahrens (changes requested):
This looks good. Can you make two separate nodes:
`Print(args)` should print a concatenated string of each argument
`Debug(message, args)` should print `message` followed by the expression for each argument, then its value, e.g. `arg1=4, arg2=5.0, (x+y)=7`.

src/finch/finch_assembly/interpreter.py
```diff
@@ -329,6 +329,40 @@ def my_func(*args_e):
                                 f"Unrecognized function definition: {func}"
                             )
                 return AssemblyInterpreterModule(self, kernels)
+            case asm.Print(args):
```
willow-ahrens, line 332: I'm not sure why we need a separate case for variable. I think we can just evaluate each argument by calling `self` on it. You can expect that print will always have multiple arguments.

src/finch/finch_assembly/interpreter.py
```diff
@@ -329,6 +329,40 @@ def my_func(*args_e):
...
+                                for arg in args
+                                if isinstance(arg, asm.Variable)
+                            ]
+                            print(arg_decls)
+                        else:
+                            raise NotImplementedError(
+                                f"Unrecognized argument type: {args}"
+                            )
+                return None
+            case asm.Debug(message, args):
```
willow-ahrens, line 349: Let's skip debug for now.

willow-ahrens (changes requested):
Since Debug is a little complicated, let's reduce scope to just focus on Print for now. So we can remove Debug from the PR and only make Print work.

src/finchlite/finch_assembly/interpreter.py
```diff
@@ -335,34 +335,17 @@ def my_func(*args_e):
...
-                        else:
-                            raise NotImplementedError(
-                                f"Unrecognized argument type: {args}"
-                            )
-                return None
-            case asm.Debug(message, args):
                 match args:
                     case asm.Variable(name, _):
-                        print(f"{message} " + f"{name}=" + str(self(args)))
+                        print(f"{name}=" + str(self(args)))
```
willow-ahrens, line 340: Don't add the name, you can just print the value. Make it behave like Print(args) = print(cat(args))

tests/test_assembly_interpreter.py
```diff
@@ -225,3 +226,78 @@ def test_simple_struct():
...
+    assert result == 9.0
+
+    capture = capsys.readouterr()
+    expected_print = (
+        "Point(x=np.float64(1.0), y=np.float64(2.0)) \n"
+        "(1, 4) \n"
+        "Point(x=np.float64(1.0), y=np.float64(2.0)) (1, 4) \n"
+        "9.0 \n"
+        "Point(x=np.float64(1.0), y=np.float64(2.0)) (1, 4) 9.0 \n"
+    )
```
willow-ahrens, line 302: Turn this into a regression test with `file_regression`. Then we can merge this and do a separate PR for printing codegen

src/finchlite/finch_assembly/interpreter.py
```diff
@@ -334,6 +335,18 @@ def my_func(*args_e):
                                 f"Unrecognized function definition: {func}"
                             )
                 return AssemblyInterpreterModule(self, kernels)
+            case asm.Print(args):
+                if isinstance(args, tuple):
```
willow-ahrens, line 339: you know that `args` will be tuple here because the `args` field in `Print` is defined to be type tuple in the AST declaration. This if branch is redundant.

src/finchlite/finch_assembly/nodes.py
```diff
@@ -753,6 +770,17 @@ def __call__(self, prgm: AssemblyNode):
                         )
                     self(func)
                 return None
+            case Print(args):
+                if isinstance(args, tuple):
```
willow-ahrens, line 774: Same here, we know that args is tuple.

willow-ahrens (changes requested):
small nitpick, remove the casing for tuple

willow-ahrens (approved):

PR #149, opened 2025-08-18
DCStats implementation for all structures and estimate_nnz

yusheng036:
This is the implementation of Finch Notation that counts non-fill entries in 1D, 2D, 3D and 4D tensor, and return a set of Degree Constraints that consists of from_indices, to_ indices and no-fill value entries. The implementation is tested across test_notation_intepreter.py and test_galley.py to ensure that the structure of notation interperter works as intended.

```
import numpy as np
from finch import DCStats

a = np.zeros((2, 2, 2, 2), dtype=int)
a[0, 0, 0, 0] = 1
a[1, 0, 1, 0] = 1
a[0, 1, 0, 1] = 1

fields = ["i", "j", "k", "l"]
stats = DCStats(a, fields)

for dc in stats.dcs:
    print(dc)
```
Output:
DC(from_indices=frozenset(), to_indices=frozenset({'k'}), value=2.0)
DC(from_indices=frozenset(), to_indices=frozenset({'j'}), value=2.0)
DC(from_indices=frozenset(), to_indices=frozenset({'k', 'i', 'l', 'j'}), value=3.0)
DC(from_indices=frozenset({'l'}), to_indices=frozenset({'k', 'i', 'j'}), value=2.0)
DC(from_indices=frozenset({'i'}), to_indices=frozenset({'k', 'l', 'j'}), value=2.0)
DC(from_indices=frozenset({'k'}), to_indices=frozenset({'i', 'l', 'j'}), value=2.0)
DC(from_indices=frozenset(), to_indices=frozenset({'l'}), value=2.0)
DC(from_indices=frozenset({'j'}), to_indices=frozenset({'k', 'i', 'l'}), value=2.0)
DC(from_indices=frozenset(), to_indices=frozenset({'i'}), value=2.0)

This functionality was originally slated for a follow-up PR, but it’s tightly coupled to the prior changes, so I’ve combined them. The new estimate_non_fill_values (or estimate_nnz) method implements a degree-constraint–based estimator for the number of non-fill entries in sparse tensors.

```
import numpy as np
from finch import DCStats

a = np.zeros((2, 2, 2, 2), dtype=int)
a[0, 0, 0, 0] = 1
a[1, 0, 1, 0] = 1
a[0, 1, 0, 1] = 1

fields = ["i", "j", "k", "l"]
stats = DCStats(a, fields)
print("Estimated non-fill values:", stats.estimate_non_fill_values())
```
Output:
Estimated non-fill values: 3.0

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@ repos:
     - id: no-commit-to-branch
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.12.9
+  rev: v0.12.10
   hooks:
     - id: ruff-check
       args: [--fix, --config=pyproject.toml]


yusheng036:
> This is impressive! Thanks for all your hard work! I wonder if there's a way to autogenerate for arbitrary dimension, but I also know that finch doesn't work well for super high-dimensional tensors anyway.

There’s a Galley function that auto-generates for arbitrary dimensions, called calc_dcs_from_structure. It returns the number of I-index combinations per J-index combination in the tensor. The implementation is fairly complicated, so I’m deferring it to a future PR.

PR #151, opened 2025-08-19
Notation lowering

mtsokol:


src/finch/compile/bufferized_ndarray.py
```diff
@@ -7,46 +9,99 @@
...
 from .lower import FinchTensorFType
 
 
+class StrideState(Enum):
+    FREE = 0
+    ACTIVE = 1
+    VISITED = 2
+
+
+class BufferizedNDArrayStrideState:
```
mtsokol, line 24: The rationale behind StrideState is that we need to support transpositions/dims permutations.

We need to track which indices we visited to know which strides to use in each loop. A good example is:
```
for i = _
    for j = _
        A(i, j)
        B(j, i) 
```
when visiting loops we first have `[i, j]` and then descending to leaves `[j]` (or the other way around). Then when we access strides it's a different stride for A and for B. But if `self.nind: int` is all we have then we're going to choose the same stride idxs for both which is incorrect.

`BufferizedNDArrayStrideState` allow us to track which strides we visited and which are active in a given descent level.

src/finch/compile/bufferized_ndarray.py
```diff
@@ -7,46 +9,99 @@
...
+        )
+
+
 class BufferizedNDArray(Tensor):
     def __init__(self, arr: np.ndarray):
         itemsize = arr.dtype.itemsize
         for stride in arr.strides:
             if stride % itemsize != 0:
                 raise ValueError("Array must be aligned to multiple of itemsize")
-        self.strides = tuple(stride // itemsize for stride in arr.strides)
```
mtsokol, line 22: Current strides logic and `as_strided` usage breaks NumPy causing incorrect outputs. I simplified it a bit.

src/finch/compile/bufferized_ndarray.py
```diff
@@ -168,16 +234,19 @@ def lower_freeze(self, ctx, tns, op):
     def lower_thaw(self, ctx, tns, op):
         return tns
 
-    def unfurl(self, ctx, tns, ext, mode, proto):
+    def unfurl(self, ctx, tns, ext, mode, idxs):
```
mtsokol, line 237: I didn't see any usages of `proto` so I changed it to `idxs` which tracks indices.

src/finch/compile/bufferized_ndarray.py
```diff
@@ -216,31 +285,42 @@ class BufferizedNDArrayAccessor(Tensor):
...
-            ftype(self.tns), self.nind, ftype(self.pos), self.op
+            ftype(self.tns), self.stride_state, ftype(self.pos), self.op
         )
 
     @property
     def shape(self):
-        return self.tns.shape[self.nind :]
+        return self.stride_state.get_shape(self.tns.shape)
+
+    def access_indices(self, idxs, op):
```
mtsokol, line 308: Dedicated method for notation interpreter.

src/finch/compile/bufferized_ndarray.py
```diff
@@ -216,31 +285,42 @@ class BufferizedNDArrayAccessor(Tensor):
...
+                tuple(None for _ in range(self.tns.ndim)), fv=StrideState.VISITED
+            ),
+            pos,
+            op,
+        )
 
-    def access(self, indices, op):
-        pos = self.pos + np.dot(
-            indices, self.tns.strides[self.nind : self.nind + len(indices)]
+    def access(self, index, op):
```
mtsokol, line 319: For actual lowering, let's support one index at the time.

src/finch/compile/bufferized_ndarray.py
```diff
@@ -319,18 +399,18 @@ def lower_thaw(self, ctx, tns, op):
...
-        """
-        tns = self.tns.asm_unpack(ctx, f"{var_n}_tns", asm.GetAttr(val, "tns"))
-        nind = asm.Variable(f"{var_n}_nind", self.nind)
-        pos = asm.Variable(f"{var_n}_pos", self.pos)
-        op = asm.Variable(f"{var_n}_op", self.op)
-        ctx.exec(asm.Assign(pos, asm.GetAttr(val, "pos")))
-        ctx.exec(asm.Assign(nind, asm.GetAttr(val, "nind")))
-        ctx.exec(asm.Assign(op, asm.GetAttr(val, "op")))
-        return BufferizedNDArrayFields(tns, pos, nind, op)
+    # def asm_unpack(self, ctx, var_n, val):
```
mtsokol, line 402: It isn't used and needs to be adjusted, so I commented it out for now.

src/finch/compile/looplets.py
```diff
@@ -153,13 +153,13 @@ def __call__(self, ctx, idx, ext, body):
 
         def lookup_node(node):
             match node:
-                case ntn.Access(tns, mode, (j, *idxs)):
-                    if j == idx and isinstance(tns, Lookup):
+                case ntn.Access(tns, mode, idxs):
+                    if idx in idxs and isinstance(tns, Lookup):
```
mtsokol, line 157: For transpositions we might have `j` and `[i, j]` so `in` check is needed.

willow-ahrens (changes requested):
Thanks for sorting this out! I've got a few big requests, and I'm happy to help if needed. The usage of StrideState sidesteps one of Finch's core compilation assumptions: all input FinchNotation programs must be concordant. This means that we expect the loop order to always match the access index order. If we want to implement permutedims, then we insert a dummy loop to make things work:
```
for i = 1:n
   for j = 1:m
      for k = i:i
         A[i, j] = B[j, k]
```

Maintaining StrideState in addition to these concepts is redundant.

We need proto for later. the idx parameter should be deleted, as we know we will always access the first index first because the program is concordant.

willow-ahrens (approved):
looks great! Thanks for your help with this!

Issue #74, opened 2025-06-03
implement type checker for FinchAssembly


willow-ahrens:
I'm not sure if we should type check as we compile, insert explicit casts, etc...

PR #155, opened 2025-08-27
Fixed annihilator bitwise properties

### LINKED ISSUE #73: Bitwise annihilator property is wrong
1 isn't an annihilator for e.g. 2 under `or`. We need three logical functions:
1. bitwise_or: bitwise op
2. logical_or: returns true if either arg is truthy
3. shortcircuit_or: returns first truthy argument

yusheng036:
Hi all,
I’ve fixed the bitwise annihilator issue and clarified the distinction between is_identity and is_annihilator:
- If the value is a number (int), we use bitwise semantics.
- If the value is a boolean, we use boolean semantics (truthiness).

For init_value changes, I used -1 instead of True for bitwise_and . True casts to 1, which isn’t the identity for &, while -1 ensures x & -1 == x; and bool(-1) is True, so it behaves correctly for booleans too.

I currently require np.bool_ for boolean paths, but for integers I limited support to Python int and did not include np.integer  as NumPy has many integer dtypes. I’d appreciate guidance on whether we should keep the policy as Python int only, or
support np.integer as well.

Other cleanups while editing algebra.py:
1. Removed operator.add from is_annihilator: +∞ is not an annihilator “for all a” (e.g., (+∞) + (−∞) = NaN).
2. Removed floor division (//) from is_identity: there’s no universal identity (e.g., 3.2 // 1 -> 3.0 ≠ 3.2).

Fixes #73

### DIFF
--- a/src/finch/algebra/algebra.py
+++ b/src/finch/algebra/algebra.py
@@ -357,9 +357,21 @@ def is_identity(op: Any, val: Any) -> bool:
 register_property(operator.add, "__call__", "is_identity", lambda op, val: val == 0)
 register_property(operator.mul, "__call__", "is_identity", lambda op, val: val == 1)
 register_property(
-    operator.or_, "__call__", "is_identity", lambda op, val: not bool(val)
+    operator.or_, "__call__", "is_identity",
+    lambda op, val: (isinstance(val, (bool, np.bool_)) and not bool(val))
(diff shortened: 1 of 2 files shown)


yusheng036:
I saw that the issue include the implementation of short_circuit_and and short_circuit_or. I would be interested to do the implementation for them!

yusheng036:
Fixes #74

PR #157, opened 2025-08-29
algebra bug fixes

yusheng036:
Hi! I’ve fixed some algebra bugs and cleaned up the code to make it neater.


willow-ahrens:
I'll close this PR as it's not a priority to fix this for now, but I'll link it to the github issue so we can come back to this as these kind of correctness bugs become more important.

src/finchlite/algebra/algebra.py
```diff
@@ -455,23 +452,15 @@ def is_annihilator(op, val):
     return query_property(op, "__call__", "is_annihilator", val)
 
 
-for op, func in [
-    (operator.add, lambda op, val: np.isinf(val)),
```
willow-ahrens, line 459: Why is infinity not an annihilator of add?

src/finchlite/algebra/algebra.py
```diff
@@ -368,9 +368,6 @@ def is_identity(op: Any, val: Any) -> bool:
 )
 register_property(operator.and_, "__call__", "is_identity", lambda op, val: bool(val))
 register_property(operator.truediv, "__call__", "is_identity", lambda op, val: val == 1)
-register_property(
```
willow-ahrens, line 371: Is the issue here that x/1 = x but 1/x != x?

willow-ahrens (changes requested):
I think these changes look good overall, but I'm pretty sure that Inf annihilates +?

willow-ahrens (approved):
I added back in infinity annihilators for +. I know they're contradictory, but they are also useful for now.

src/finchlite/algebra/algebra.py
yusheng036: I have two reasons for this: 
1) two values cant be an annihilator for operator.add but + inf and -inf is an annihilator for add. This is the explanation

2) When +inf +(-inf) is inputted into numpy it gives NaN and doesnt return its annihilator value.

PR #159, opened 2025-09-02
Added comparison elementwise operators. 

TheRealMichaelWang:
Per issue #4, certain elementwise operators are missing. This PR adds comparison elementwise operators:
- less
- less equal
- greater
- greater equal
- equal
- not equal

I've also ran pre-commit hooks via `poetry run pre-commit run -a`.


willow-ahrens (changes requested):
I think we need dunder methods for lazy_tensor as well. Also, please add tests, next to the other elementwise functions. Thanks!

TheRealMichaelWang:
> I think we need dunder methods for lazy_tensor as well. Also, please add tests, next to the other elementwise functions. Thanks!

I added operator overloads for the new elementwise compare operators and I added tests for them in `test_interface.py`. Running the pre-check hooks works, but running the tests via `poetry run pytest` shows a lot of failed tests, most of which are something along the lines of `TypeError: unhashable type: 'TestEagerTensor'`. Is this normal?

TheRealMichaelWang:
@willow-ahrens When I run `poetry run pytest` I get thousands of errors related to `TypeError: unhashable type: 'Scalar'` and other unhashable type errors with other types.

It's really strange, I think I'm stuck.

willow-ahrens:
Apologies Michael, you've managed to find a particularly tricky contradiction in the finch-tensor-lite design. We ultimately don't want to be hashing arrays, but when we pass numbers into `Literal` nodes, we wrap the numbers in scalar arrays and then later, when we try to hash `Literal`, we have to hash the value of the literal. The fix for this is to be more consistent about which parts of Finch Logic are meant to be static, and which parts are considered "inputs" to the program. 

In resolving this, I'll state a few things I know to be true:
1. We shouldn't be hashing mutable things like arrays.
2. The data we put into a Literal should be considered static at the moment it's wrapped in the Literal node
3. At the interface level, it's impossible to tell the difference between `A + 1` and `A + (x:Int)`. We can't unilaterally decide whether scalars are constant or variable.

This feels like the ghost of issues https://github.com/finch-tensor/Finch.jl/issues/467 and https://github.com/finch-tensor/finch-tensor/issues/25 come back to haunt us.

I believe the resolution proposed in 25 is sound. We pick a set of scalars (-Inf, -1, 0, 1, Inf) to automatically promote to constants, and another set of scalars to treat as variables, and provide users a way to override the decision either way. When we convert arguments to Finch Logic in interface.jl, we should ensure there is a difference between something assumed to be constant and compiled as part of the program AST itself, and something which is just an expression where we expect the type to be constant.

When we hash `Literal`, I'm convinced we should just hash the `ftype`. Eventually, we should introduce `StaticScalar` which puts the value of the scalar into the ftype. `Asarray` of a number should produce scalar or staticscalar depending on whether the value is in the whitelisted set (-inf, -1, 0, 1, inf). For this PR, I'll make everything a regular scalar, and only hash the ftype of literals.

PR #160, opened 2025-09-03
rename

willow-ahrens:
renaming


mtsokol (approved):
LGTM - regarding the new name: we have two sensible options:
- `import finchlite`, same as `import llvmlite`, concise suffix notifying that this is a 'lite' version of some larger project.
- `import finch_lite`, similar to e.g. `import llvm_installer` - which actually matches PyPI/GitHub project name `finch-tensor` and `llvm-installer`.

I think I prefer `finchlite` now. We can always come up with our import conversions, such as:
```py
import finchlite as fl
# or
import finchlite as finch
```

PR #144, opened 2025-08-08
Safe Buffer implementation

Osmansiddiquer:
TLDR: Introduces safe buffer with tests for better bound checking in buffers
Aims to resolve #62 

This pull request introduces a new safety wrapper for buffer operations in the Finch assembly layer, ensuring that all buffer accesses are bounds-checked. The main additions are the `SafeBuffer` and `SafeBufferFormat` classes, as well as a convenient `make_safe` function to wrap any buffer with safety checks. Comprehensive tests are also included to verify correct behavior and error handling.

**New buffer safety features:**
* Added `SafeBuffer` class to `src/finch/finch_assembly/safe_buffer.py`, which wraps any existing buffer and enforces bounds checking on all load and store operations, raising `IndexError` for out-of-bounds access.
* Introduced `SafeBufferFormat` class, allowing safe buffer formats to be composed and used interchangeably with existing buffer formats.

**Testing:**
* Added `tests/buffers/test_safe_buffer.py` with parameterized tests to verify correct bounds checking, format compatibility, and resizing behavior for safe buffers.

PR #162, opened 2025-09-06
Junikimm717/safebuffers

junikimm717:
- Adds the notion of a "safe buffer" to finch, which emits some code to check for index out of bounds cases when working with finch load and store operations on arrays.
- The current behavior is to report an index out of bounds and automatically panic. Future work may be done to address this with say a revised calling convention, but at the moment this choice was made for simplicity :)

Things that should definitely be reviewed:
1. The testing code at the bottom of safebuffer.py. This is my first time touching the finch python codebase, and so there may be some ways that I worked with the array types (ie spamming ctypes everywhere) that aren't the "morally correct way" to do things in finch python. And following that,
2. Some of the "register_property" calls that I added in c.py, which were added in an attempt to fix the errors I was seeing.

Somehow, this PR ended up becoming more than about just safebuffers and also includes testing for trivial serialization with numpy types.


src/finch/codegen/numpy_buffer.py
```diff
@@ -238,3 +238,8 @@ def construct_from_numba(self, numba_buffer):
         Construct a NumpyBuffer from a Numba-compatible object.
         """
         return NumpyBuffer(numba_buffer[0])
+
```
willow-ahrens, line 241: I'm not sure why this needs to be here?

src/finch/codegen/safe_buffer.py
```diff
@@ -0,0 +1,160 @@
...
+    access = mod.finch_access
+    change = mod.finch_change
+    print(ab)
+    print(access(ab, ctypes.c_size_t(0)))
+    change(ab, ctypes.c_size_t(0), ctypes.c_double(2))
+    print(ab)
+    # should fail here.
+    print(access(ab, ctypes.c_size_t(3)))
+    change(ab, ctypes.c_size_t(3), ctypes.c_double(2))
+    print(ab)
```
willow-ahrens, line 160: Again, I think this should likely move to the tests.

src/finch/codegen/safe_buffer.py
```diff
@@ -0,0 +1,160 @@
...
+)
+
+import finch.finch_assembly as asm
+from finch.finch_assembly import Buffer
+from finch.util import qual_str
+from finch.codegen.c import CBufferFType, CCompiler, CContext, CStackFType, c_type
+from finch.codegen.numba_backend import NumbaBufferFType
+
+
+def SafeBuffer(
```
willow-ahrens, line 20: why a factory instead of a wrapper class like in https://github.com/finch-tensor/finch-tensor-lite/pull/144? I think it would be more straightforward to define
```
class SafeBuffer(AbstractBuffer...):
    def __init__(self, child):
        self.child_buffer = child
    def load(self, idx):
        assert idx < self.child_buffer
        return self.child_buffer.load
```

willow-ahrens (changes requested):
This is exactly on the right track!

Could you make the safe buffer a more straightforward wrapper, rather than using a factory, like in https://github.com/finch-tensor/finch-tensor-lite/pull/144?

Also, I think we need numba codegen, and tests for the assembly interpreter, c, and numba.

src/finchlite/codegen/numpy_buffer.py
```diff
@@ -238,3 +240,10 @@ def construct_from_numba(self, numba_buffer):
         Construct a NumpyBuffer from a Numba-compatible object.
         """
         return NumpyBuffer(numba_buffer[0])
+
+
+if __name__ == "__main__":
+    import finch.finch_assembly as asm
+
+    prgm = asm.AssemblyNode
+    pass
```
mtsokol, line 249: Not needed

mtsokol (commented):
I think for windows platform we need another if before `subprocess` which adds an executable path or `shell=True` to the arguments: https://stackoverflow.com/questions/25651990/oserror-winerror-193-1-is-not-a-valid-win32-application

willow-ahrens:
Why do we need getattr on safebuffer?

src/finchlite/codegen/safe_buffer.py
```diff
@@ -0,0 +1,203 @@
...
+        return self._underlying_format.c_data(*args, **kwargs)
+
+    def _c_check(self, ctx, buf, idx):
+        ctx.add_header("#include <stdio.h>")
+        ctx.add_header("#include <stdlib.h>")
+        idx_n = ctx.freshen("computed")
+        ctx.exec(
+            f"{ctx.feed}size_t {idx_n} = ({ctx(idx)});\n"
+            f"{ctx.feed}if ({idx_n} < 0 || {idx_n} >= ({self.c_length(ctx, buf)}))"
+            "{"
```
willow-ahrens, line 72: I think this line needs ctx.feed, right?

PR #165, opened 2025-09-09
BasicBlocks, CFG, and CFGBuilder for dataflow-analysis

idkwhttsay:


idkwhttsay:
TODO: Do a system design for how to build Basic Blocks and CFG based on the AST. Understand how to handle loops, breaks, continues, function calls, etc. and how to represent them in the CFG. 
Also, figure out a better way of representing the connections between blocks (probably a map will work best).

src/finchlite/finch_assembly/cfg.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def add_successor(self, successor: "BasicBlock") -> None:
+        if successor not in self.successors:
+            self.successors.append(successor)
+
+        if self not in successor.predecessors:
+            successor.predecessors.append(self)
+
+    def to_dict(self):
+        """Convert BasicBlock to a dictionary for JSON serialization."""
```
willow-ahrens, line 50: why do we need to serialize basic blocks?

src/finchlite/finch_assembly/cfg.py
```diff
@@ -0,0 +1,282 @@
...
+    Slot,
+    Stack,
+    Store,
+    TaggedVariable,
+    Unpack,
+    WhileLoop,
+)
+
+
+class BasicBlock:
```
willow-ahrens, line 30: basic blocks, etc, should live in a separate folder from finch_assembly, because dataflow is supposed to apply to all 3 IRs. The CFGBuilder for FinchAssembly can stay in this file though.

src/finchlite/finch_assembly/cfg.py
```diff
@@ -0,0 +1,282 @@
...
+            "predecessors": [str(block.id) for block in self.predecessors],
+        }
+
+    def __str__(self):
+        import json as _json
+
+        return _json.dumps(self.to_dict(), indent=4)
+
+
+class ControlFlowGraph:
```
willow-ahrens, line 64: same here, this should not be specific to FinchAssembly.

src/finchlite/finch_assembly/cfg.py
```diff
@@ -0,0 +1,282 @@
...
+            case ForLoop(var, _start, _var, body):
+                before_block = self.current_cfg.current_block
+
+                body_block = self.current_cfg.new_block()
+                after_block = self.current_cfg.new_block()
+
+                before_block.add_successor(body_block)
+                before_block.add_successor(after_block)
+
+                # TODO: figure out a RIGHT way to represent 'ForLoop' initialization
```
willow-ahrens, line 208: I think we should introduce a new variable, so that
```
for i = start:stop
    body
```
becomes
```
j = start
while j <= end
    i = j
    body
    j = j + 1
```
Then we can handle it like the while case.

src/finchlite/finch_assembly/cfg.py
```diff
@@ -0,0 +1,282 @@
...
+            case BufferLoop(_buf, var, body):
+                before_block = self.current_cfg.current_block
+
+                body_block = self.current_cfg.new_block()
+                after_block = self.current_cfg.new_block()
+
+                before_block.add_successor(body_block)
+                before_block.add_successor(after_block)
+
+                # TODO: figure out a RIGHT way to represent 'BufferLoop' initialization
```
willow-ahrens, line 225: Same as above, make it a while loop and call the CFGBuilder on the whileLoop

src/finchlite/finch_assembly/cfg.py
willow-ahrens: Treat the ControlFlowGraph like another IR. Put it in the dataflow directory, give it a cfg.py file that holds a PrinterContext.

src/finchlite/finch_assembly/cfg.py
```diff
@@ -0,0 +1,282 @@
...
+    def __init__(self, func_name: str):
+        self.block_counter = 0
+        self.name = func_name
+        self.blocks: dict[str, BasicBlock] = {}
+
+        # initialize ENTRY and EXIT blocks
+        self.entry_block = self.new_block()
+        self.exit_block = self.new_block()
+
+        self.current_block = self.new_block()
```
willow-ahrens, line 76: Instead of using the CFG to hold the "current block", can we use the CFGBuilder to hold it? It doesn't feel like a meaningful attribute outside of the CFGBuilder

willow-ahrens (changes requested):
This is a great start! I have a few style changes and suggestions to make this more predictable, sound, and concise.

src/finchlite/finch_assembly/cfg.py
idkwhttsay: (recap)

TODO: make a printer context class for basic blocks + cfgs in the dataflow/cfg.py

src/finchlite/finch_assembly/nodes.py
```diff
@@ -219,7 +219,7 @@ class Assign(AssemblyTree):
         rhs: The right-hand side to evaluate.
     """
 
-    lhs: Variable | Stack
+    lhs: TaggedVariable | Variable | Stack
```
idkwhttsay, line 222: @willow-ahrens  i'm extremely skeptical about this

idkwhttsay:
@willow-ahrens i'm getting this from pytest:

2693 passed, 216 warnings in 15.90s
src/finchlite/autoschedule/executor.py:7: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]

is it supposed to be like that?

all other warnings are for the same reason (default body types), i 99.9999% sure that it's not my fault but I think it needs to be fixed, it makes me anxious, ahahah

src/finchlite/dataflow/cfg.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def new_block(self):
+        bid = f"{self.name}_{self.block_counter}"
+        self.block_counter += 1
+        block = BasicBlock(bid)
+        self.blocks[bid] = block
+        return block
+
+
+class CFGPrinterContext:
```
idkwhttsay, line 44: is it what you meant by PrinterContext? or do you want a AssemblyPrinterContext type printer, instead of converting everything to json?

src/finchlite/dataflow/cfg.py
idkwhttsay, line 1: also i'm not sure about directory structure and file naming

idkwhttsay:
copilot reviews are actually useful, lol

tests/reference/test_asm_cfg_printer_dot.txt
```diff
@@ -0,0 +1,26 @@
+dot_product: #entry=dot_product_0, #exit=dot_product_1
+    dot_product_0: #preds=[]
+
+    dot_product_1: #preds=[dot_product_4, dot_product_5]
+
+    dot_product_2: #preds=[dot_product_0, dot_product_3]
+        a_0
+        b_0
```
idkwhttsay, line 8: @willow-ahrens add the beginning of the function, should I just list the variables as statements or do this:

var_0 = var_0
var_1 = var_1

?

src/finchlite/finch_assembly/nodes.py
idkwhttsay: and my concern is that it might generate some errors in future (?)

willow-ahrens: It's okay to assign to a taggedvariable. This is okay.

tests/test_printers.py
```diff
@@ -280,6 +282,75 @@ def test_tagged_asm_printer_if(file_regression):
...
+                        asm.Return(var),
+                    )
+                ),
+            ),
+        )
+    )
+
+    root = asm.nodes.number_assembly_ast(root)
+    cfgs = CFGBuilder()(root)
+    cfg_printer = CFGPrinterContext()
```
willow-ahrens, line 349: Shouldn't we be able to just call `str`?

willow-ahrens (changes requested):
small tweaks, this is almost done!

idkwhttsay:
@willow-ahrens ready for review!

willow-ahrens (approved):
I just renamed/unified some of the finch assembly passes that it exports

PR #171, opened 2025-09-16
Testing for Safe Buffers

junikimm717:
Adds some requested changes to safebuffers testing:

1. Reduced test cases for index accessing
2. Added regression testing for codegen involving buffer access

### DIFF
--- a/tests/test_codegen.py
+++ b/tests/test_codegen.py
@@ -472,34 +472,10 @@ def test_c_load_safebuffer(size, idx):
     [
         (*params, compiler)
         for params in [
-            (
-                -1,
-                2,
-            ),
-            (
-                -1,
-                3,
-            ),
-            (
-                0,
-                2,
-            ),
-            (
-                1,
-                2,
-            ),
-            (
-                2,
-                3,
-            ),
-            (
-                2,
-                2,
-            ),
-            (
-                3,
-                2,
-            ),
+            (-1, 2),
+            (1, 2),
+            (2, 3),
+            (2, 2),
         ]
         for compiler in [asm.AssemblyInterpreter(), NumbaCompiler()]
     ],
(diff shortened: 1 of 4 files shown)


junikimm717:
It is not clear to me why windows specifically would fail like this 💀
I personally do not have a windows machine on hand so could someone take a look as well?

willow-ahrens:
Key info: It's not your fault, the backend is not selecting reproducible int types for different machines.

tests/test_codegen.py
```diff
@@ -441,6 +441,67 @@ def test_simple_struct(compiler):
...
 
 
+@pytest.mark.parametrize(
+    ["compiler", "extension"],
+    [
+        (CGenerator(), ".c"),
+        (NumbaGenerator(), ".py"),
+    ],
+)
+def test_safe_loadstore_regression(compiler, extension, file_regression):
```
mtsokol, line 451: @junikimm717 Answering your question: On [win32 (which is our GitHub runner)](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models) `long int` is 32 and `long long int` is 64.

So for linux/macos int64 and usize_t -> c_long, c_ulong,
for windows int64 and usize_t -> c_longlong, c_ulonglong.

We can either:
- somehow parametrize test output file, which I'm not sure is supported by the pytest plugin we're using,
- use `ctypes.c_long` etc. already in the test,
- parametrize inputs, or use fixtures, sth like:
```py
def test_safe_loadstore_regression(compiler, extension, file_regression, c_int64, c_size_t):
    ...

@pytest.fixture
def c_int64():
    return ctypes.c_longlong if platform() == "win32" else ctypes.c_long
```

willow-ahrens: I ran into this on my APL to C compiler that I built out of a fork of finch-tensor-lite for my DSL class. In Finch.jl, we were running regressions on 32-bit and 64-bit systems separately. It was a pain because julia couldn't cross-compile, so every time the regression output changed we had to spin up a 32-bit runner to regen the reference output. Ideally, we would be able to cross-compile with Finch, so that we can regen both kinds of output on the same system. For now, I think we should skip regression tests on windows, and file a github issue to implement cross-compilation.

mtsokol (approved):

willow-ahrens (approved):
This is good enough for now. By the way, how did you generate reference files for windows?

mtsokol:
> By the way, how did you generate reference files for windows?

Right now we only have one `tests/reference/test_safe_loadstore_regression_compiler1__py_win_.py` windows-specific reference file (the rest of tests are compatible with all platforms). I manually updated it, checking the error log in CI.

Issue #172, opened 2025-09-19
number the variables in finch assembly


idkwhttsay:
implement a node that is an extension to the "Variable" finch assembly node ands rewrite the original Variable nodes in the AST to these so called "numbered variables"

PR #173, opened 2025-09-19
number variables in finch assembly

idkwhttsay:
PR for issue #172

created NumberedASTGenerator class, it deep copies the entire AST and numbers variables by changing Variable(name, type) to SSAVariable(name, type, id). SSAVariable is a new type of Assembly Node

### DIFF
--- a/src/finchlite/finch_assembly/nodes.py
+++ b/src/finchlite/finch_assembly/nodes.py
@@ -104,8 +104,10 @@ class Variable(AssemblyExpression):
 @dataclass(eq=True, frozen=True)
 class TaggedVariable(AssemblyExpression):
     """
-    Represents an SSA form of Variable assembly node that refers to variable
-    named `name`, with type `type`, and has ID `id`.
+    Represents a numbered Variable assembly node that refers to variable
+    named `name`, with type `type`, and has ID `id`. Used in the the
+    representation of programs which need to distinguish between multiple
+    uses of the same variable, such as SSA or numbered occurrence forms.
 
     Attributes:
         variable: The variable that it references.


src/finchlite/finch_assembly/nodes.py
```diff
@@ -101,6 +101,31 @@ def __repr__(self) -> str:
...
+    named `name`, with type `type`, and has ID `id`.
+
+    Attributes:
+        name: The name of the variable that it references.
+        type: The type of the variable that it references.
+        id: The ID of the variable
+    """
+
+    name: str
+    type: Any
```
willow-ahrens, line 117: Instead of separating name and type here, can we just have a var node that holds the variable? This will make it easier to process the IR, as we don't need to duplicate cases for SSAVariable and Variable.

src/finchlite/finch_assembly/nodes.py
```diff
@@ -101,6 +101,31 @@ def __repr__(self) -> str:
         return literal_repr(type(self).__name__, asdict(self))
 
 
+@dataclass(eq=True, frozen=True)
+class SSAVariable(AssemblyExpression):
```
willow-ahrens, line 105: Let's call this TaggedVariable

src/finchlite/finch_assembly/nodes.py
```diff
@@ -709,3 +734,73 @@ def __call__(self, prgm: AssemblyNode):
...
+            case Block(bodies):
+                return Block(tuple(self(b) for b in bodies))
+            case Module(funcs):
+                return Module(tuple(self(f) for f in funcs))
+            case Unpack(lhs, rhs):
+                return Unpack(self(lhs), self(rhs))
+            case Repack(val):
+                return Repack(self(val))
+            case node:
+                raise NotImplementedError(node)
```
willow-ahrens, line 803: This is solid, but you can write this more succinctly with the `PostWalk` rewrite strategy. Check other places in the codebase for how `PostWalk` is used.

willow-ahrens (changes requested):
This is exactly the right functionality! I've just got a few comments on code style.

idkwhttsay:
@willow-ahrens also if you want to test it, you would have to change AssemblerPrinterContext and create a test.py file (at least that's how I've done it, not the best way tho). I'll attach the files and results in this comment:

Screenshot 2025-09-20 at 7 27 32 PM

[test.py](https://github.com/user-attachments/files/22448724/test.py)
[ast_printer.py](https://github.com/user-attachments/files/22448725/ast_printer.py)

src/finchlite/finch_assembly/nodes.py
```diff
@@ -485,9 +508,10 @@ def children(self):
         return [self.name, *self.args, self.body]
 
     @classmethod
-    def from_children(cls, name, *args, body):
+    def from_children(cls, *children):
         """Creates a term with the given head and arguments."""
-        return cls(name, args, body)
+        name, *arg_nodes, body = children
+        return cls(name, tuple(arg_nodes), body)
```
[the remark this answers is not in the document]
idkwhttsay: as a result, it started working properly, no idea why is that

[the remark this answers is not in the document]
willow-ahrens: Thanks! It looks like we hadn't tested the rewriting on FinchAssembly (We use it a lot for FinchLogic currently)

willow-ahrens (changes requested):
Thanks! The PR itself looks great! To add a test, look at how `file_regression` is used in the rest of the testsuite. You can set up a "regression" test to compare the output of a compiler pass to a reference output, and it automatically handles the regeneration of reference outputs with the `--regen-all` flag. Although this code looks bulletproof to me, I think we should have *some* small amount of test coverage, perhaps you can add your test example as a single, simple regression test.

src/finchlite/finch_assembly/nodes.py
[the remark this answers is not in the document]
mtsokol: Right, `def from_children(cls, name, *args, body)` doesn't make sense if we won't pass `body` as a keyword argument.

mtsokol (changes requested):
Please also add your new node to `AssemblyPrinterContext.__call__`.

idkwhttsay:
> > [test.py](https://github.com/user-attachments/files/22448724/test.py) [ast_printer.py](https://github.com/user-attachments/files/22448725/ast_printer.py)
> 
> @idkwhttsay I think a regression test would be useful here. About the screenshot you enclosed: Should a new `a_*` variable be created only only on a new assignment? So `eq(a_0, 5): a_1: int64 = ...`.

Well, anyways we know the original variable is `a`, so it won't make any problems. I would even say it would make everything much easier, because when I'll get the output from the dataflow analysis (var_id -> whatever is the output), I can just map the outputs to each of the tagged definitions

src/finchlite/finch_assembly/nodes.py
```diff
@@ -714,6 +719,26 @@ def __call__(self, prgm: AssemblyNode):
...
+                            )
+                ctx_2(body)
+                body_code = ctx_2.emit()
+                feed = self.feed
+                self.exec(
+                    f"{feed}def {func_name}_{id}({', '.join(arg_decls)}) -> "
+                    f"{qual_str(return_t)}:\n"
+                    f"{body_code}\n"
+                )
+                return None
```
mtsokol: Can we remove it and modify `case Function` to not to duplicate it all? So `match` statement there will know whether we have variable or tagged(variable).

mtsokol (changes requested):

mtsokol:
> @mtsokol should I also add number_assembly_ast and TaggedVariable to finch_assembly/**init**.py?

Sure, let's add it!

mtsokol (approved):
LGTM! Let's wait for @willow-ahrens vote as well.

PR #177, opened 2025-09-22
Added is_idempotent, mapjoin and aggregate for tensordef and improved description

yusheng036:
I have added this following changes:
-  Implemented is_idempotent in finchlite.algebra to be used in tensordef.aggregate.
-  Implemented mapjoin and aggregate functionality for tensordef to builds the shape (axes, sizes, fill) after undergoing mapjoin and aggregate. 
- Added max and min for is_identity
- improved description for tensor_stats.py
- implemented merge_dc_join for join-like operators


willow-ahrens (commented):
@kylebd99, is it possible to re-use FinchLogic instead of TensorDef?

PR #178, opened 2025-09-22
Add remaining element-wise functions

### LINKED ISSUE #4: Implement elementwise ops

joaosferreira:
This PR fixes #4 by adding the remaining element-wise functions.


joaosferreira:
Following up on the discussion regarding `elementwise` and how to make it work with `np.clip`.

I tried implementing `lazy.clip` as follows:

```python
def clip(x, /, *, min=None, max=None) -> LazyTensor:
    return elementwise(np.clip, defer(x), defer(min), defer(max))
```

However, `np.clip` complains of receiving too many positional arguments:

```python
>>> import finchlite
>>> import numpy as np
>>> a = np.array([[1, 2], [3, 4]])
>>> finchlite.clip(a, 1, 3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: clip() takes 1 positional argument but 3 were given
```

My suggestion is to add a `kwargs` parameter to `elementwise` in order to do something like:

```python
def clip(x, /, *, min=None, max=None) -> LazyTensor:
    return elementwise(np.clip, defer(x), a_min=defer(min), a_max=defer(max))
```

willow-ahrens:
let's just make a new function, `splat_clip(a, min, max) = np.clip(a, min=min, max=max)` and then pass `splat_clip` to elementwise

willow-ahrens (approved):
Thanks for your help with this!

PR #188, opened 2025-09-28
Update typing to resolve mypy warnings

mtsokol:


src/finchlite/finch_logic/nodes.py
```diff
@@ -327,7 +332,7 @@ def fields(self) -> list[Field]:
 
 
 @dataclass(eq=True, frozen=True)
-class Query(LogicTree):
+class Query(LogicTree, Generic[LNVar1, LNVar2]):
```
mtsokol, line 335: How about making all `LogicNode` fields in nodes' classes a generic type with a `LogicNode` upper bound? This way we can have more precise typing for function, arguments, dicts in our compiler's logic, making it possible to jump to a definition of some field of a specific node (as in `propagate_into_reformats` in `matched` here).

PR #192, opened 2025-10-02
worklist algorithm

idkwhttsay:


idkwhttsay:
@willow-ahrens is there an LLVM printing style for the outputs of the dataflow? For now I just output the output_states of each block after the CopyPropagation. I haven't checked if the outputs are correct, but they look correct

idkwhttsay:
i hate git

PR #193, opened 2025-10-03
Create MallocBuffers

junikimm717:
This pull request implements MallocBuffers for finch. This is an interface alongside the more standard NumpyBuffer to interact with buffers. The functions that make MallocBuffers possible are implemented in a c file in the same directory as the malloc_buffer.py file.

Some work is probably still warranted regarding ctypes? I am getting a strange error on one of the codegen tests (something about c_double.__rmul__ requiring a "return type"), and it is not immediately clear to me exactly where the proper patch should be.

### DIFF
--- a/src/finchlite/codegen/malloc_buffer.py
+++ b/src/finchlite/codegen/malloc_buffer.py
@@ -1,0 +1,212 @@
+import ctypes
+from pathlib import Path
+
+import numpy as np
+
+from finchlite.codegen.c import CBufferFType, CStackFType, c_type, load_shared_lib
+from finchlite.codegen.numpy_buffer import CBufferFields
+from finchlite.finch_assembly import Buffer
+from finchlite.util import qual_str
+
+backend_lib = load_shared_lib(
+    (Path(__file__).parent / "malloc_buffer_backend.c").read_text(encoding="utf-8")
+)
+
+
+class CMallocBuffer(ctypes.Structure):
+    pass
+
+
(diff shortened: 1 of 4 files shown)


willow-ahrens (commented):
This looks like it's on the right track. I have a question I'd like to hear your feedback on. I was thinking maybe it may make sense to jit different implementations for different element types in the buffer. What do you think of this? Is there any advantage?

src/finchlite/codegen/malloc_buffer.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def deserialize_from_c(self, obj, c_buffer):
+        pass
+        # this is handled by the resize callback
+
+    def construct_from_c(self, c_buffer):
+        """
+        Construct a MallocBuffer from a C-compatible structure.
+
+        TODO: incomplete
```
willow-ahrens, line 210: is this incomplete?

tests/test_codegen.py
```diff
@@ -147,6 +148,124 @@ def test_codegen(compiler, buffer):
     assert_equal(result, expected)
 
 
+@pytest.mark.parametrize(
+    ["compiler", "buffer"],
+    [
+        (CCompiler(), MallocBuffer),
+        (asm.AssemblyInterpreter(), MallocBuffer),
```
willow-ahrens, line 155: solid, I was looking for a check like this and I found this.

willow-ahrens (approved):
This looks good to me.

PR #195, opened 2025-10-06
Implemented aggregate and its test

yusheng036:
I have implemented this following test:

1) Implemented aggregate
2) implemented its test


kylebd99 (changes requested):
This looks good to me! My only request is that you add a few more test cases. Currently, they all use the same set of DCs and result in the same expected value (which happens to be the same one that you would get w/out the aggregation). This was a bad test suite for me to write. A better one should include some DCs where the aggregate changes the estimate # of non-fill values.

PR #196, opened 2025-10-06
Wma/add einsum ir

willow-ahrens:


src/finchlite/finch_einsum/nodes.py
TheRealMichaelWang, line 1: Looks good on the whole. Later, an `EinsumExpr` for 
- Querying the size of a dimension/axis of a tensor
- Querying the coordinate array of a COO matrix
- Querying the value array of a COO matrix

These can all be added in as single `EinsumExpr` node that queries some property from a tensor. 

Can be done later, but is 100% needed for implementing Einsum -> Insum

PR #199, opened 2025-10-06
fix binary function issue

### LINKED ISSUE #198: Operator/NumPy functions cannot accept more than two arguments
My mapjoin function does not accept more than two arguments because both operator.mul and numpy.multiply are strictly binary functions. For example, when I try to write:

new_fill_value = op(*(s.fill_value for s in args))

it fails if args has more than two elements.
This is the error message when you try to run op.mul with more than 3 arguments:
```
import operator as op
op.mul(1, 2, 3)
```
Result:
TypeError: mul expected 2 arguments, got 3

yusheng036:
This is the changes I have implemented: 
1) changed mapjoin (reduce to op.mul)
2) changed its respective test

fixes #198


willow-ahrens (approved):

kylebd99 (approved):
LGTM

PR #204, opened 2025-10-07
Wma/degensym

willow-ahrens:


idkwhttsay:
> the use of gensym in cfg builder meant that the variable names in the CFG output depended on how many times the program was run.

yeah, i also noticed but didn't give enough attention to it. also it was the reason why "pytest tests/test_printers.py" failed while "pytest" succeeded, so everytime I had to run the entire pytest test suit

idkwhttsay (approved):
lgtm

PR #205, opened 2025-10-08
WorkList algorithm + CopyPropagation on finch assembly (resolved conflicts)

idkwhttsay:
#192 resumed with conflicts resolved
 
Implement a Copy Propagation dataflow analysis on the Finch Assembly
Print Control Flow Graphs with the lattice values from the copy propagation
Add a comprehensive test for cfg building, numbering Assembly AST, copy propagation

### DIFF
--- a/src/finchlite/symbolic/dataflow.py
+++ b/src/finchlite/symbolic/dataflow.py
@@ -3,7 +3,11 @@ from abc import ABC, abstractmethod
 
 
 class BasicBlock:
-    """A basic block of Finch's Control Flow Graph."""
+    """Linear sequence of statements with a single entry and exit.
+
+    Successor/predecessor lists encode the CFG edges. Statements execute in
+    order; no internal branching occurs within a BasicBlock.
+    """
 
     def __init__(self, id: str) -> None:
         self.id = id
@@ -41,7 +45,11 @@ class BasicBlock:
 
 
 class ControlFlowGraph:
-    """Control-Flow Graph (CFG) for FinchAssembly."""
+    """Collection of BasicBlocks plus explicit ENTRY and EXIT nodes.
+
+    Provides helpers to allocate uniquely named blocks and holds the global
+    block mapping used by dataflow analyses.
+    """
 
     def __init__(self):
         self.block_counter = 0


tests/test_printers.py
```diff
@@ -470,80 +402,204 @@ def test_tagged_asm_printer_dot(file_regression):
     file_regression.check(str(prgm), extension=".txt")
 
 
-def test_asm_cfg_printer_dot(file_regression):
-    c = asm.Variable("c", np.float64)
+def test_tagged_asm_printer_comprehensive(file_regression):
```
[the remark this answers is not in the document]
willow-ahrens: I'm fine with that, a little test library of asts for inputs? That's great.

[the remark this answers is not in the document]
willow-ahrens: actually, on second thought, it might be more work than it's worth to unify all the different requirements between different tests and compiler passes. I'd say that it's not a pressing issue, so maybe we leave this as-is.

src/finchlite/finch_assembly/__init__.py
```diff
@@ -1,5 +1,11 @@
+from .assembly_dataflow import (
```
willow-ahrens: Could we keep cfg_builder in Finch assembly, but move the dataflow passes to a separate folder outside of finch assembly? I think of the specific dataflow analyses as something separate from the definition of the IR.

src/finchlite/finch_assembly/assembly_dataflow.py
```diff
@@ -259,3 +269,140 @@ def __call__(
...
+            return self._variables_equal(val1, val2)
+
+        if isinstance(val1, Literal) and isinstance(val2, Literal):
+            return val1.val == val2.val
+
+        return False
+
+    def __str__(self) -> str:
+        """
+        Print the CFG with dataflow annotations embedded in statements.
```
willow-ahrens: Okay. Here's the refactor I'd like to see: Make a class called "AbstractAssemblyDataflow". Move this logic to "DataFlowAnalysis" in Dataflow.py, and just have a method that the child class is supposed to implement called "stmt_str" or similar, to print a statement together with the variables and lattice values in the statement. AbstractAssemblyDataflow should handle the printing of each assembly statement, and find the variables in it to print their values. Then the AssemblyCopyPropagation defines a method "print_lattice_value" to print the singular lattice value associated with a copy-propagation analysis.

willow-ahrens (changes requested):
Solid start! See my comments for the refactoring about printing. Eventually, we'll want to handle copy propagation with global value numbering, right?

tests/test_printers.py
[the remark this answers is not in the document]
idkwhttsay: okay!

src/finchlite/finch_assembly/dataflow.py
```diff
@@ -0,0 +1,150 @@
...
+        Copy propagation is a forward analysis.
+        """
+        return "forward"
+
+    def transfer(self, stmts, state: dict) -> dict:
+        new_state = state.copy()
+
+        for stmt in stmts:
+            match stmt:
+                case Assign(lhs, rhs):
```
willow-ahrens: Don't we need to check if the rhs is a variable? If it's a variable, we should check whether that variable has a copy, and use that instead. I'm asking that you do the "searching for root variable" calculation that happened in the print function in the transfer function instead. That should simplify the printing.

src/finchlite/finch_assembly/dataflow.py
```diff
@@ -0,0 +1,150 @@
...
+    def _get_variable_name(self, var) -> str | None:
+        match var:
+            case Variable(name, _):
+                return name
+            case TaggedVariable(Variable(name, _), _):
+                return name
+            case _:
+                return None
+
+    def _values_equal(self, val1, val2) -> bool:
```
willow-ahrens: I'd rather the algorithm be written carefully to know whether the values in dictionaries are Literals, Variables, or taggedVariables. If all that matters is the var, I'd like to see tagged_var.var == tagged_var2.var

src/finchlite/finch_assembly/dataflow.py
```diff
@@ -0,0 +1,150 @@
...
+
+        return result
+
+    def print_lattice_value(self, name, state: dict) -> str:
+        visited: set[str] = set()
+        value = state.get(name)
+
+        # iterate through states until we reach a Literal or a variable
+        # without a mapping (root value)
+        while True:
```
willow-ahrens: This step indicates to me that the transfer function wasn't written correctly. You should always pass around the root variable for each copy, right?

tests/reference/test_asm_copy_propagation_comprehensive.txt
```diff
@@ -0,0 +1,112 @@
...
+    a_0 = 10
+    b_0 = a_1
+    c_0 = b_1
+    d_0 = 5
+
+main_6: #succs=[EXIT_1]
+
+main_7: #succs=[main_9]
+    assert(gt(c_1: 10, 5))
+    a_2 = add(a_3: 10, d_1: 5)
```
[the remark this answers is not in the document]
idkwhttsay: 1) do you mean print lattice values for the printed ast, not the printed cfg?

2) did it, looks much better now

idkwhttsay:
> Okay, reviewed! For the comments about printing don't feel you need to take them too seriously. For the comments about the root finding: Is the pointer chasing a necessary part of the dataflow? This lecture notes has more info: https://www.cs.cornell.edu/courses/cs4120/2019sp/lectures/20dataflow/lec20-sp16.pdf

there's is a slight chance that I might have misunderstood the copy propagration. I'll read the book and take a look at this lecture notes

idkwhttsay:
> > Okay, reviewed! For the comments about printing don't feel you need to take them too seriously. For the comments about the root finding: Is the pointer chasing a necessary part of the dataflow? This lecture notes has more info: https://www.cs.cornell.edu/courses/cs4120/2019sp/lectures/20dataflow/lec20-sp16.pdf
> 
> there's is a slight chance that I might have misunderstood the copy propagration. I'll read the book and take a look at this lecture notes

Screenshot 2025-11-01 at 6 13 24 PM

based on this, seems like my code is working correctly

src/finchlite/finch_assembly/dataflow.py
```diff
@@ -0,0 +1,147 @@
...
+
+                    # resolve RHS to its root lattice value
+                    resolved_rhs = self._resolve_root(rhs, new_state)
+                    new_state[var_name] = resolved_rhs
+
+                    # invalidate any copies that directly point to this variable name
+                    to_remove: list[str] = []
+                    for name, val in new_state.items():
+                        # check if variable `var_name` is used in `val` expression
+                        if self._check_var_use(var_name, val):
```
idkwhttsay: here I also had a dumb mistake. now it iterates over the entire expression and checks the uses of the variable `var`

idkwhttsay:
> > > Okay, reviewed! For the comments about printing don't feel you need to take them too seriously. For the comments about the root finding: Is the pointer chasing a necessary part of the dataflow? This lecture notes has more info: https://www.cs.cornell.edu/courses/cs4120/2019sp/lectures/20dataflow/lec20-sp16.pdf
> > 
> > 
> > there's is a slight chance that I might have misunderstood the copy propagration. I'll read the book and take a look at this lecture notes
> 
> Screenshot 2025-11-01 at 6 13 24 PM
> based on this, seems like my code is working correctly

just realized, you were right. there is absolutely no reason to track the root lattice value, because if the root lattice value is still a variable (or an expression), on the instruction level it's still a memory load which will not change anything in terms of speed. it's only useful if a root lattice value is a literal, which is extremely unlikely

src/finchlite/symbolic/dataflow.py
```diff
@@ -100,10 +101,18 @@ def __str__(self) -> str:
             # get the input state for this block
             input_state = self.input_states.get(block.id, {})
 
-            # delegate formatting of each statement to subclass
-            lines.extend(
-                f"    {self.stmt_str(stmt, input_state)}" for stmt in block.statements
-            )
+            if self.direction() == "forward":
+                state = copy.deepcopy(input_state)
```
idkwhttsay: i'm doing a deepcopy to not change the original input_state, otherwise it would be a bug that I won't be able to find in futute

src/finchlite/finch_assembly/dataflow.py
```diff
@@ -0,0 +1,180 @@
...
+    direct copies; expressions and literals are not propagated.
+    """
+
+    def direction(self) -> str:
+        """
+        Copy propagation is a forward analysis.
+        """
+        return "forward"
+
+    def print_lattice_value(
```
willow-ahrens: use a PostOrderDFS instead of writing your own recursive traversal.

PR #210, opened 2025-10-10
Add Einsum Lowerer

TheRealMichaelWang:
- Added `EinsumLowerer` which converts finch logic IR into finch einsum IR
- Added pytests for `EinsumLowerer`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,13 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Install Poetry
         uses: snok/install-poetry@v1
+      - name: Configure Poetry
+        run: |
+          if [[ "${{ runner.os }}" == "Windows" ]]; then
+            poetry env use python
+          else
+            poetry env use python${{ matrix.python }}
+          fi
       - name: Install package
         run: |
           poetry install --extras test
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,8 +4,8 @@ description = ""
 version = "0.1.0"
 readme = "README.md"
 authors = [{name = "Willow Ahrens", email = "willow.marie.ahrens@gmail.com"}]
-requires-python = '>=3.11,<4.0'
-dependencies = ['numpy>=1.19', 'numba>=0.61', 'donfig>=0.8.1.post1,<0.9.0', "lark (>=1.3.0,<2.0.0)"]
+requires-python = '>=3.11,<3.14'
+dependencies = ['numpy>=1.19,<2.4', 'numba>=0.61,<0.63.1', 'donfig>=0.8.1.post1,<0.9.0', "lark (>=1.3.0,<2.0.0)"]
 
 [tool.poetry]
 packages = [{include = "finchlite", from = "src"}]
@@ -13,12 +13,12 @@ requires-poetry = '>=2.0,<3.0'
 
 [project.optional-dependencies]
 test = [
-    "pytest>=7.4.4,<8",
-    "pre-commit>=3.6.0,<4",
-    "pytest-cov>=4.1.0,<5",
-    "sparse>=0.17.0",
-    "scipy>=1.7,<2",
-    "numpy>=0.61.0",
+    "pytest>=7.4.4,<7.6",
+    "pre-commit>=3.6.0,<3.9",
+    "pytest-cov>=4.1.0,<4.2",
+    "sparse>=0.17.0,<0.17.1",
+    "scipy>=1.7,<1.17",
+    "numpy>=0.61.0,<2.4",
     "mypy==1.15.*",
     "pytest-mypy==1.*",
     "pytest-regressions~=2.8.1",


TheRealMichaelWang:
I restored `ein.Produce` as an einsum-level statement. But instead of adding an abstract einsum plan statement node, I changed the type of `bodies` in `ein.Plan` to `EinsumNode`, to make it as general as possible.

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,219 @@
...
+                ret_args: list[ein.EinsumExpr] = []
+                for arg in m_args:
+                    match arg:
+                        case ein.Call(ein.Literal(op2), _) if op2 == operation:
+                            ret_args.extend(flatten_args(arg.args))
+                        case _:
+                            ret_args.append(arg)
+                return tuple(ret_args)
+
+            return ein.Call(ein.Literal(operation), flatten_args(args))
```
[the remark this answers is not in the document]
willow-ahrens: I think I'm just asking you to inline this function

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,219 @@
...
+                        self.rename_einsum(
+                            self.lower_to_einsum(body, bodies, bindings, definitions),
+                            self.get_next_alias(),
+                            definitions,
+                        )
+                    )
+
+        return ein.Plan(tuple(bodies))
+
+    def lower_to_einsum(
```
willow-ahrens: Inline these by wrapping the pattern in each case inside of the rhs of a query node

TheRealMichaelWang:
The `lower_to_einsum` function in EinsumLowerer was inlined into `compile_plan`, and other functions were renamed for clarity. However, to avoid code repetition, I implemented 2 separate functions `compile_mapjoin` and `compile_aggregate`.

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+        bindings: dict[str, Any] = {}
+        definitions: dict[str, ein.Einsum] = {}
+        return self.compile_plan(prgm, bindings, definitions), bindings
+
+    def reorder_einsum(
+        self, einsum: ein.Einsum, idxs: tuple[ein.Index, ...]
+    ) -> ein.Einsum:
+        return ein.Einsum(einsum.op, einsum.tns, idxs, einsum.arg)
+
+    def compile_mapjoin(
```
willow-ahrens: delete this function

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+                            name,
+                            body.rhs.fields,
+                            operation,
+                            init,
+                            arg,
+                        )
+                    )
+                case Query(
+                    Alias(name),
+                    Reorder(Aggregate(Literal(operation), Literal(init), arg, _), idxs),
```
willow-ahrens: Delete this case, you can assume the optimizer doesn't emit Reorder(Aggregate()).

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+            case Relabel(
+                Alias(name), idxs
+            ):  # relable is really just a glorified pointwise access
+                return ein.Access(
+                    tns=ein.Alias(name),
+                    idxs=tuple(ein.Index(idx.name) for idx in idxs),
+                )
+            case Literal(value):
+                return ein.Literal(val=value)
+            case Aggregate(Literal(operation), Literal(init), arg, _):
```
willow-ahrens: You can assume the optimizer removes aggregate from inside pointwise expressions.

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+from finchlite.symbolic import gensym
+
+
+class EinsumLowerer:
+    def __call__(self, prgm: Plan) -> tuple[ein.Plan, dict[str, Any]]:
+        bindings: dict[str, Any] = {}
+        definitions: dict[str, ein.Einsum] = {}
+        return self.compile_plan(prgm, bindings, definitions), bindings
+
+    def reorder_einsum(
```
willow-ahrens: do not use this function.

willow-ahrens (changes requested):
Overall, this is an improvement. Still, this code handles many more cases than it needs to. When we write compilers, we try to separate concerns by splitting into multiple passes. Let the optimizer produce a really straightforward, normalized form of the AST, and only write the simplest lowerer you can for normal form programs. Don't try to normalize the AST at the same time as you lower it. See autoschedule/compiler.py for a similar set of match cases to handle, but note that you don't need to disambiguate between transpose and mapjoin, and you don't need to match reformats (I think?).

src/finchlite/autoschedule/einsum.py
TheRealMichaelWang: `reorder_einsum` has been deleted. Hadn't noticed that it was still there.

src/finchlite/autoschedule/einsum.py
TheRealMichaelWang: Case deleted.

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+    ) -> ein.Plan:
+        bodies: list[ein.EinsumNode] = []
+
+        for body in plan.bodies:
+            match body:
+                case Plan(_):
+                    bodies.append(self.compile_plan(body, bindings, definitions))
+                case Query(Alias(name), Table(Literal(val), _)):
+                    bindings[name] = val
+                case Query(Alias(name), MapJoin(Literal(operation), args)):
```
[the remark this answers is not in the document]
TheRealMichaelWang: Done. In this case, the query statement will be turned into:
```python
case Query(Alias(name), rhs):
    einarg = self.compile_operand(rhs, bodies, bindings, definitions)
    bodies.append(ein.Einsum(
        op=ein.Literal(overwrite),
        tns=ein.Alias(name),
        idxs=tuple(ein.Index(field.name) for field in body.rhs.fields),
        arg=einarg,
    ))
```

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+                            bodies,
+                            bindings,
+                            definitions,
+                            name,
+                            body.rhs.fields,
+                            operation,
+                            args,
+                        )
+                    )
+                case Query(Alias(name), Reformat(_, MapJoin(Literal(operation), args))):
```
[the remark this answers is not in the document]
TheRealMichaelWang: Done

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+                            bindings,
+                            definitions,
+                            name,
+                            body.rhs.fields,
+                            operation,
+                            args,
+                        )
+                    )
+                case Query(
+                    Alias(name), Reorder(MapJoin(Literal(operation), args), idxs)
```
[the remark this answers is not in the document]
TheRealMichaelWang: Done

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,262 @@
...
+                ):
+                    bodies.append(
+                        self.compile_mapjoin(
+                            bodies, bindings, definitions, name, idxs, operation, args
+                        )
+                    )
+                case Query(
+                    Alias(name), Aggregate(Literal(operation), Literal(init), arg, _)
```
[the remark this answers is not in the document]
TheRealMichaelWang: Can you elaborate? Do you mean match `Query(Alias(name), Aggregate(Literal(operation), Literal(init), Reorder(arg), _)`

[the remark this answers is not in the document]
TheRealMichaelWang: Through testing I can confirm `Query(Alias(name), Aggregate(Literal(operation), Literal(init), Reorder(arg), _)`

src/finchlite/autoschedule/einsum.py
TheRealMichaelWang: I think this is the old code.

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,151 @@
...
+                        ein.Einsum(
+                            op=ein.Literal(overwrite),
+                            tns=ein.Alias(name),
+                            idxs=tuple(
+                                ein.Index(field.name) for field in body.rhs.fields
+                            ),
+                            arg=einarg,
+                        )
+                    )
+                case lgc.Query(lgc.Alias(name), lgc.Reorder(rhs, idxs)):
```
willow-ahrens: this is also redundant

src/finchlite/autoschedule/einsum.py
```diff
@@ -0,0 +1,151 @@
...
+                    bindings[name] = val
+                case lgc.Query(
+                    lgc.Alias(name), lgc.Aggregate(lgc.Literal(operation), lgc.Literal(init), arg, idxs)
+                ) | lgc.Query(
+                    lgc.Alias(name),
+                    lgc.Aggregate(lgc.Literal(operation), lgc.Literal(init), lgc.Reorder(arg, _), idxs),
+                ):
+                    einidxs = tuple(ein.Index(field.name) for field in idxs)
+                    if init != init_value(operation, type(init)):
+                        bodies.append(
```
willow-ahrens: instead of appending to a list of bodies, return either an einsum or a plan of einsums.

willow-ahrens (changes requested):
Some more simplification requested.

willow-ahrens (changes requested):
Here's my small ask

Issue #194, opened 2025-10-05
implement Print statement for all IRs


willow-ahrens:
Finch assembly now has print statements as of #148. Next up: implement print statements for C and numba backends, FinchNotation, FinchLogic, FinchEinsum, FinchFuse. All print statements should respect python logging levels.

PR #217, opened 2025-10-22
Print statement for codegen backends

siyi-hu:
Add a print statement to c++ and numba backends to print out the content of input arguments. Solving issue #194 .


willow-ahrens (approved):

willow-ahrens (changes requested):
I'm not sure about this. It seems like extra complexity, and nobody has asked for it in a while.

willow-ahrens:
@siyi-hu we didn't end up needing print statements at this point, but your hard work on the project is still appreciated! Thank you!

PR #227, opened 2025-10-29
Implementation of AnnotatedQuery and nodes_to_logic

### LINKED ISSUE #225: Implementation of nodes_to_logic
The Finch Logic AST and DCStats live at two different abstraction layers in the Finch ecosystem:
* Finch Logic AST describes what computation should happen .it’s a semantic representation of tensor programs. It knows how expressions are structured, but it has no quantitative understanding of sparsity, density, or non-zero patterns.
* DCStats quantify how large or expensive those logical expressions might be. They provide statistical metadata such as estimated non-zeros (nnz), index dependencies, and density correlations via degree constraints (DCs).
Without a bridge, these two systems can’t talk to each other:
* The Logic AST knows structure but can’t evaluate statistical effects of transformations.
* The DCStats system knows how to merge and reduce cardinalities but has no way to interpret abstract nodes like MapJoin or Aggregate.

The Role of nodes_to_logic:
The nodes_to_logic module acts as a semantic–statistical bridge. It recursively walks the Finch Logic AST and ties each node type (Table, MapJoin, Aggregate, Alias, etc.) to the appropriate pre-existing DCStats constructor or merge rule.

### LINKED ISSUE #226: Implementation of AnnotatedQuery
AnnotatedQuery is the core data structure that the Logical Optimizer uses to represent and reason about a Finch logical plan that has been annotated with metadata, specifically, index relationships, operators, and statistical estimates. It’s essentially the optimizer’s working model of a query: part semantic, part statistical.

Annotation Step: The optimizer builds an AnnotatedQuery object that attaches rich metadata to this logical form:
* idx_lowest_root → the lowest node where each index first appears
* idx_op and idx_init → what operation and initializer apply to each index (for reductions)
* parent_idxs → which indices depend on which others (index hierarchy)
* connected_components and connected_idxs → graph partitions of indices that interact
* reduce_idxs → indices being reduced out
* ST → the statistics backend class 
* point_expr → the full Finch Logic expression with attached stats (TensorStats)

yusheng036:
I have finished my implementation for the following above functions and would love some feedbacks!!

Fix #226  & Fix #225

### DIFF
--- a/src/finchlite/galley/LogicalOptimizer/annotated_query.py
+++ b/src/finchlite/galley/LogicalOptimizer/annotated_query.py
@@ -15,10 +15,9 @@ class AnnotatedQuery:
     output_name: Alias | None
     reduce_idxs: list[str]
     point_expr: "LogicNode"
-    idx_lowest_root: OrderedDict[str, int]
+    idx_lowest_root: OrderedDict[str, LogicNode]
     idx_op: OrderedDict[str, Any]
     idx_init: OrderedDict[str, Any]
-    hash_to_node: OrderedDict[int, "LogicNode"]
     parent_idxs: OrderedDict[str, list[str]]
     original_idx: OrderedDict[str, str]
     connected_components: list[list[str]]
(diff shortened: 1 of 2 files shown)


yusheng036:
Hi @kylebd99,

I’ve finished all of the requests from last week’s meeting.
Here are the updates:
1. Added reorder and reformat to annotated_query.
2. Rewrote all tests to use insert_statistics.

However, I ran into a bug that I haven’t been able to solve. I was hoping we could take a look at it together during the meeting. The issue involves DenseStats.mapjoin — I keep getting a fill value of 0 when it should be 1.

src/finchlite/galley/LogicalOptimizer/annotated_query.py
```diff
@@ -0,0 +1,158 @@
...
+)
+
+
+@dataclass
+class AnnotatedQuery:
+    ST: type
+    output_name: Alias | None
+    reduce_idxs: list[str]
+    point_expr: "LogicNode"
+    idx_lowest_root: OrderedDict[str, int]
```
kylebd99: Per our discussion yesterday, let's just make the value type in this dictionary a FinchLogic node and remove the notion of node_id at all.

kylebd99 (approved):
Only a few small comments, otherwise LGTM. Fix those and go ahead and merge it.

PR #229, opened 2025-11-06
Improve hash and equality for FinchLogic Literal

kylebd99:
Equality and hash are undefined or poorly defined for array types, so we use the id for hashing them and we apply all() to get a boolean for equality.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
     - id: no-commit-to-branch
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.3
+  rev: v0.14.4
   hooks:
     - id: ruff-check
       args: [--fix, --config=pyproject.toml]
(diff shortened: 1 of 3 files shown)


src/finchlite/finch_logic/nodes.py
```diff
@@ -73,13 +73,22 @@ class Literal(LogicNode):
...
+            return hash(id(self.val))
 
-    def __eq__(self, other):
-        if not isinstance(other, Literal):
+    def __eq__(self, value):
+        if not isinstance(value, Literal):
             return False
-        return ftype(self.val) == ftype(other.val)
-
+        equality = self.val == value.val
```
willow-ahrens: If hash is pointer hash, then equality must be pointer equality, not `all`.

willow-ahrens (changes requested):
I think equality should be pointer equality if hash isn't implemented.

kylebd99:
Yeah, that’s a good point, I’ll fix that.

kylebd99 (commented):
W.r.t. hash and equality having the same semantics, in a perfect world I would check whether the value type is hashable and if so always use pointer equality. However, this seems like it would basically require a try except block which seems messy. So, for the moment, I just check for a valid equality implementation and use pointer id if it returns an array.

PR #230, opened 2025-11-07
Add Support for GetAttribute and Indirect Access to Einsum Interpreter 

TheRealMichaelWang:
- Added support for GetAttribute Einsum AST Node
  - Currently supports `shape`, `elems`, and `coords` which query attributes from sparse tnesors
  - Has a dim field which is required for some queries
- Added support for indirect access
  - Includes support for indirect access with one index-operand (i.e. `B[ACoords[k]]`)
  - Includes support for mixing indirect accessors and regular indices (i.e. `B[AcoordsI[k], j]`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
     - id: no-commit-to-branch
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.10
+  rev: v0.14.14
   hooks:
     - id: ruff-check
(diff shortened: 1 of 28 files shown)


willow-ahrens (changes requested):
Great work, thanks! My first request: Please rename to GetAttr (like it is in FinchAssembly, FinchNotation). Additionally, please just use e.g. `coords3` rather than `coords`, 3, to keep a uniform interface to attributes in finch-lite. Once you've fixed this I'll take another look.

willow-ahrens (changes requested):

PR #233, opened 2025-11-11
[DRAFT] SuperTensor einsum interpreter

jessicaliu06:
### `supertensor.py`
- Represents a tensor using the following attributes:
    - a base tensor
    - a map that encodes which modes of the logical tensor are combined into each mode of the base tensor, and in what order these logical modes are combined. 
- Supports item access using logical coordinates.

### `interpreter.py`
- **Input:** an einsum AST where the tensor name of each `Access` node is bound to a `SuperTensor`, and the indices of each `Access` node are _logical_ indices on the `SuperTensor`
    - Note: The base tensor of a `SuperTensor` in the input AST will have some arbitrary shape.
- **Algorithm:**
    - Collect all `Access` nodes in the einsum expression.
    - Reshape the base tensors to combine index groups in a useful way.
    - Bind each tensor name to the correctly-shaped base tensor.
    - Rewrite each `Access` node with the proper indices to access the correctly-shaped base tensor.
    - Call the standard einsum interpreter on the rewritten AST to compute the base tensor of the output tensor.
    - Wrap the output base tensor into an output `SuperTensor`.

### `test.py`
- Contains a simple test case, for which the output of the SuperTensor interpreter matches the output of the regular interpreter.

### DIFF
--- a/src/finchlite/supertensor/interpreter.py
+++ b/src/finchlite/supertensor/interpreter.py
@@ -74,7 +74,7 @@ class SuperTensorEinsumInterpreter:
                     ]
 
                     # Restore the logical shape of the SuperTensor.
-                    supertensor = self.bindings[access.tns.name]
+                    supertensor = self.bindings[access.tns]
                     logical_tns = np.empty(
                         supertensor.shape, dtype=supertensor.base.dtype
                     )
@@ -87,8 +87,8 @@ class SuperTensorEinsumInterpreter:
                     )
 
                     # Map the tensor name to its corrected base tensor and index list.
-                    corrected_bindings[access.tns.name] = corrected_supertensor.base
-                    corrected_idx_lists[access.tns.name] = new_idx_list
+                    corrected_bindings[access.tns] = corrected_supertensor.base
+                    corrected_idx_lists[access.tns] = new_idx_list
 
                 # Construct the corrected index list and mode map for the output.
                 new_output_idx_list = sorted(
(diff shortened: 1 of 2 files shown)


src/finchlite/supertensor/interpreter.py
```diff
@@ -0,0 +1,216 @@
...
+                case ein.Access():
+                    for child in curr.children:
+                        postorder(child)
+                    accesses.append(curr)
+                case _:
+                    if hasattr(curr, "children"):
+                        for child in curr.children:
+                            postorder(child)
+            
+        postorder(node)
```
willow-ahrens: use PostOrderDFS from symbolic.py, inline this function

src/finchlite/supertensor/interpreter.py
```diff
@@ -0,0 +1,216 @@
...
+                    tns_name = access.tns.name
+                    idxs = [idx.name for idx in access.idxs]
+                    input_idx_list.append((tns_name, idxs))
+                
+                idx_groups = SuperTensorEinsumInterpreter._group_indices(output_name, output_idxs, input_idx_list)
+
+                # Assign a new index name to each group of original indices.
+                new_idxs = {}
+                for k, (tensor_set, _) in enumerate(idx_groups):
+                    new_idxs[tensor_set] = f"i{k}"
```
willow-ahrens: use Namespace from Symbolic.py to create fresh index variable names.

src/finchlite/supertensor/interpreter.py
```diff
@@ -0,0 +1,216 @@
...
+                    tns_name = access.tns.name
+                    supertensor = self.bindings[access.tns.name]
+                    idxs = [idx.name for idx in access.idxs]
+                    inputs.append((tns_name, supertensor, idxs))
+
+                corrected_bindings = {}
+                corrected_idx_lists = {}
+                for tns_name, supertensor, input_idx_list in inputs:
+                    new_idx_list = []
+                    mode_map = []
```
willow-ahrens: we can construct a dictionary globally, which maps idx -> newidx, which I think would be helpful here.

PR #234, opened 2025-11-11
Traversal change and istree and isdescendant change

yusheng036:
Hi! I have addressed todays meeting change by changing the location of the Traversal (PostOrderDFS and PreOrderDFS) to traversal.py in symbolic and added istree and isdescendant into utility.py in Finch_Logic. I have also added each of their respective test in the test files

### DIFF
--- a/src/finchlite/symbolic/traversal.py
+++ b/src/finchlite/symbolic/traversal.py
@@ -21,15 +21,7 @@ def intree(n1, n2):
     """
     Return True iff `n1` occurs in the subtree rooted at `n2`.
     """
-    target = n1
-    stack = [n2]
-    while stack:
-        n = stack.pop()
-        if n == target:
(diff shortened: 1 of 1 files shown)


kylebd99 (commented):
Overall, these look great! A couple of small notes:
1. I think you should put intree and isdescendant in traversal.py as well because that's not specific to the finch_logic node.
2. Instead of putting the tests in test_logic_interpreter, I'd make a test_traversal.py file to explicitly test the traversal functions. We'll probably add to that file, so it seems worthwhile to have its own test file.

kylebd99 (changes requested):

yusheng036:
Hi @willow-ahrens,

That was an oversight on my part. I’ve updated the code to reflect the requested changes.

Yu Sheng

PR #236, opened 2025-11-12
Add abstract types for statements + typing fixes

willow-ahrens:


src/finchlite/finch_assembly/nodes.py
```diff
@@ -573,15 +583,15 @@ def children(self):
 
 
 @dataclass(eq=True, frozen=True)
-class Block(AssemblyTree):
+class Block(AssemblyTree, AssemblyStatement):
```
idkwhttsay: are you sure that block is also an AssemblyStatement? I think it's just an AssemblyTree in the same way as Function is

idkwhttsay (approved):
lgtm

src/finchlite/autoschedule/compiler.py
```diff
@@ -191,6 +192,36 @@ class LogicLowerer:
     def __init__(self, mode: str = "fast"):
         self.mode = mode
 
+    @overload
```
mtsokol: Once we decide to use parametric types we can get rid of these `@overload`s.

mtsokol (commented):
LGTM, but I have two questions:
1. `Freeze`, `Thaw`, and `Declare` have `def result_format` but they implement `NotationStatement` instead of `NotationExpression`, should we switch it?
2. In all `nodes.py` files `*Statement` trait is introduced where `*Tree` is already implemented (except for `Module` class in notation). What is the rationale here? Could we use, for example `NotationTree` where `NotationStatement` is used as a type hint?

src/finchlite/autoschedule/compiler.py
willow-ahrens: For now, I'll leave the overloads, but later we should probably use parametric types. I just didn't want to use them in the IR nodes.

mtsokol: Makes sense!

PR #245, opened 2025-11-23
Implementation of cansplitpush, repeat_operator, replace_and_remove_nodes and find_lowest_nodes

### LINKED ISSUE #241: Change from str to Field in AnnotatedQuery
Making types clearer by changing the backbone of AnnotatedQuery to use Field instead of str. These change was made to ensure the semantics of AnnotatedQuery becomes clearer. In addition, idx_lowest_root would be a LogicExpression rather than a LogicNode, since the lowest root should always be an expression rather than a statement.

### LINKED ISSUE #242: Implement cansplitpush and repeat_operator
Hi! I have implemented two new functions in algebra/operator.py: cansplitpush and repeat_operator. 

The function cansplitpush return True if a reduction with operator `x` can be 'split-pushed' through a pointwise operator `y`.  CanSplitPush is allowed when the operator is a repeat_operator, both operator 'x' and operator 'y' are the same and operator 'x' is both commutative and associative.

The function repeat_operator returns the operator if there exists an operator g such that f(x, x, ..., x)  (n times)  is equal to g(x, n). If the operator is idempotent, then no operator is returned. If the operator is operator.add, then it returns operator.mul as x + x + ....(n times) = x . n.
If the operator is operator.mul, then it returns exponent as m x m x ....(n times) = m ^ n

### LINKED ISSUE #243: Implement find_lowest_root
find_lowest_roots determines how far a reduction over an index can be safely pushed down inside a Finch Logic expression without changing the algebraic rules of the program. These are the three rules that find_lowest_root obey to:

Image

### LINKED ISSUE #244: implement replace_and_remove_nodes
replace_and_remove_nodes perform replace and remove arguments of a pointwise MapJoin expression.

Argument:
```
expr : LogicNode
        The expression to transform. Typically a `MapJoin` in a pointwise subexpression.
node_to_replace : LogicNode
        The node to replace when it appears as an argument to `expr`, or as `expr` itself.
new_node : LogicNode
        The node that replaces `node_to_replace` wherever it is found.
nodes_to_remove : Collection[LogicNode]
       A collection of nodes that, if present as arguments to a `MapJoin`, should be removed from its argument list.
```

Returns:
```
    LogicNode
        A new `MapJoin` node with updated arguments if `expr` is a `MapJoin`,
        `new_node` if `expr == node_to_replace`, or the original `expr`
        otherwise.
```

yusheng036:
Hi! I have implemented cansplitpush, repeat_operator, replace_and_remove_nodes and find_lowest_nodes and have written test for every one of the functions. I have changed the backbone of annotated_query to make its semantic clearer.  I have also file an issue for each function. 

Fix #241, Fix #242, Fix #243 & Fix #244

Thank you for your time!!

### DIFF
--- a/tests/test_galley.py
+++ b/tests/test_galley.py
@@ -1408,39 +1408,95 @@ def test_get_idx_connected_components(parent_idxs, connected_idxs, expected):
 
 
 @pytest.mark.parametrize(
-    "arg_names,node_to_replace,nodes_to_remove,expected_names",
+    "expr,node_to_replace,new_node,nodes_to_remove,expected_names",
     [
-        (["a", "b", "c"], "b", [], ["a", "a", "c"]),
-        (["a", "b", "c"], "b", ["c"], ["a", "a"]),
-        (["a", "b", "c"], "c", ["c"], ["a", "b"]),
-        (["a", "b", "c"], "b", ["b"], ["a", "c"]),
-        (["a", "b", "c"], "c", [], ["a", "b", "a"]),
+        (
+            MapJoin(
+                Literal("op"),
+                (
+                    Table(Literal("a"), (Field("a"),)),
+                    Table(Literal("b"), (Field("b"),)),
+                    Table(Literal("c"), (Field("c"),)),
+                ),
+            ),
+            Table(Literal("b"), (Field("b"),)),
+            Table(Literal("a"), (Field("a"),)),
+            set(),
+            ["a", "a", "c"],
+        ),
+        (
+            MapJoin(
+                Literal("op"),
+                (
+                    Table(Literal("a"), (Field("a"),)),
+                    Table(Literal("b"), (Field("b"),)),
+                    Table(Literal("c"), (Field("c"),)),
+                ),
+            ),
+            Table(Literal("b"), (Field("b"),)),
+            Table(Literal("a"), (Field("a"),)),
+            {Table(Literal("c"), (Field("c"),))},
+            ["a", "a"],
+        ),
+        (
+            MapJoin(
+                Literal("op"),
+                (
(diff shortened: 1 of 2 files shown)


tests/test_galley.py
```diff
@@ -1372,5 +1386,154 @@ def test_get_reducible_idxs(reduce_idxs, parent_idxs, expected):
...
+        name[k]: [name[n] for n in v] for k, v in connected_idxs.items()
+    }
+
+    components = get_idx_connected_components(parent_field_idxs, connected_field_idxs)
+    result = [[field.name for field in comp] for comp in components]
+
+    assert result == expected
+
+
+@pytest.mark.parametrize(
```
kylebd99: For this test, rather than giving a set of arg names. I would make all of the inputs FinchLogicNodes, and I would make sure that you have a test which involves a pointwise expression that has some level of nesting.

tests/test_galley.py
```diff
@@ -1372,5 +1386,154 @@ def test_get_reducible_idxs(reduce_idxs, parent_idxs, expected):
...
+        ),
+        # Leaf case:
+        # root = Table(A(i)), reduce over i → [A]
+        (
+            Table(Literal("A"), (Field("i"),)),
+            "i",
+            ["A"],
+        ),
+        # Nested case:
+        # root = MapJoin(mul, [A(i,j), B(i)]), reduce over i → [A]
```
kylebd99: Nit: I think you mean B(j) in this comment.

kylebd99 (approved):
Hey, sorry for the slow review, I think everything looks good. I had a couple of very minor comments, but I'm going to go ahead and approve it. Please make the changes, then go ahead and merge.

PR #247, opened 2025-11-25
Hash Tables!

junikimm717:
Creating Hash Tables that can map immutable data types to other immutable data types


junikimm717:
@willow-ahrens 
Could I get some help with some issues relating to mypy and numba?
It appears that mypy suddenly refuses to believe that numba.types is a module that exists 💀 
All other tests pass though

mtsokol:
> one possible concern is that these files are technically build artifacts from the STC build system... I'll investigate if I can just get away with slipping in the repository though

When you fetch submodules the files are there so it doesn't matter from the build process POV.

mtsokol:
SG, let's move https://github.com/finch-tensor/finch-tensor-lite/pull/248 in first.

junikimm717:
@willow-ahrens the numba serialization has been implemented, and it passes the tests
(some other lowering code was apparently generating faulty code though 💀)
Once I can get the green light and have this PR merged, I'll probably be stepping away from the project for some time.

mtsokol:
> was it decided to not use a git submodule in the end?

Yes, we decided it's easier for students to understand it.

PR #248, opened 2025-11-27
Immutable Tuples as C Structs, not pointers

junikimm717:
Turn immutable tuples into C structs, and keep mutable tuples as C pointers.

### DIFF
--- a/src/finchlite/codegen/c.py
+++ b/src/finchlite/codegen/c.py
@@ -13,14 +13,13 @@ from typing import Any
 
 import numpy as np
 
-from finchlite.finch_assembly.struct import MutableStructFType
-
 from .. import finch_assembly as asm
 from ..algebra import query_property, register_property
 from ..finch_assembly import (
     AssemblyStructFType,
     BufferFType,
     ImmutableStructFType,
+    MutableStructFType,
     TupleFType,
 )
 from ..symbolic import Context, Namespace, ScopedDict, fisinstance, ftype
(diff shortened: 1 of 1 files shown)


junikimm717:
While doing this PR, I have been stomping out other bugs, including stuff from my own codegen tests
Did you know that ctypes conversions of numpy scalars all share the same buffer?

```python
import numpy as np
import numpy.ctypeslib as ctl

a = np.int64(2)
b = np.int64(1)

s1 = ctl.as_ctypes(a)
s2 = ctl.as_ctypes(b)

print(s1.value, s2.value)
# this prints 1 1 :skull:
```

willow-ahrens (changes requested):
Looks good, I would only ask for an ImmutableFType class-based interface.

src/finchlite/codegen/c.py
```diff
@@ -204,12 +211,24 @@ def construct_from_c(fmt, c_obj):
...
+    scalar_ctype = np.ctypeslib.as_ctypes_type(obj.dtype)
+    ptr_ctype = ctypes.POINTER(scalar_ctype)
+    return arr.ctypes.data_as(ptr_ctype).contents
+
+
 register_property(
     np.generic,
     "serialize_to_c",
     "__attr__",
-    lambda fmt, obj: np.ctypeslib.as_ctypes(obj),
```
mtsokol: Can we have it as `np.ctypeslib.as_ctypes(np.array(obj)),` instead? It looks that 0-d arrays print your example correctly. Then we shouldn't need `scalar_to_ctypes_copy`.

PR #254, opened 2025-12-06
Fix up the malloc buffers

junikimm717:
Adding fixes to malloc buffers so they are compliant with the rest of the backend.

### DIFF
--- a/src/finchlite/codegen/malloc_buffer.py
+++ b/src/finchlite/codegen/malloc_buffer.py
@@ -7,7 +7,10 @@ from textwrap import dedent
 
 import numpy as np
 
-from finchlite.codegen.c_codegen import (
+from ..finch_assembly import Buffer, Stack
+from ..finch_assembly.nodes import AssemblyExpression
+from ..util import qual_str
+from .c_codegen import (
     CBufferFType,
     CContext,
     CStackFType,
@@ -16,10 +19,7 @@ from finchlite.codegen.c_codegen import (
     load_shared_lib,
     serialize_to_c,
 )
-from finchlite.codegen.numpy_buffer import CBufferFields
-from finchlite.finch_assembly import Buffer
-from finchlite.finch_assembly.nodes import AssemblyExpression, Stack
-from finchlite.util import qual_str
+from .numpy_buffer import CBufferFields
 
 
 class CMallocBufferStruct(ctypes.Structure):
@@ -120,7 +120,10 @@ class MallocBufferBackend:
 
     @classmethod
     def library(cls, ftype: MallocBufferFType) -> CMallocBufferLibrary:
-        # lazy compile the library.
+        """
+        Returns compiled library to operate on a buffer outside of
+        the kernel.
+        """
         if ftype in cls._library:
             return cls._library[ftype]
         ctx = CContext()


src/finchlite/codegen/c_codegen.py
```diff
@@ -604,12 +605,24 @@ def __init__(
...
         self.types = types
         self.slots = slots
+        self.datastructures: dict[Hashable, Any] = {}
 
     def add_header(self, header):
         if header not in self._headerset:
             self.headers.append(header)
             self._headerset.add(header)
 
+    def add_datastructure(self, key: Hashable, handler: "Callable[[CContext], Any]"):
```
mtsokol: Instead of `"Callable[[CContext], Any]"` you can `from __future__ import annotations` [ref](https://docs.python.org/3/library/__future__.html#module-contents) and have `Callable[[CContext], Any]`.

src/finchlite/codegen/malloc_buffer.py
```diff
@@ -138,62 +252,83 @@ def element_type(self):
...
+        assert isinstance(buf.obj, CBufferFields)
         ctx.exec(f"{ctx.feed}({buf.obj.data})[{ctx(idx)}] = {ctx(value)};")
 
-    def c_resize(self, ctx, buf, new_len):
+    def c_resize(self, ctx: "CContext", buf: "Stack", new_len: "AssemblyExpression"):
+        assert isinstance(buf.obj, CBufferFields)
+        methods: CMallocBufferMethods | None = ctx.datastructures.get(self)
+        assert methods is not None, (
+            "A Mallocbuffer must be unpacked before being operated on!"
+        )
```
mtsokol: I prefer `if self not in ctx.datastructures: raise Exception`

src/finchlite/codegen/malloc_buffer.py
```diff
@@ -1,53 +1,168 @@
...
+class MallocBufferBackend:
+    _library: dict[Hashable, CMallocBufferLibrary] = {}
+
+    @classmethod
+    def gen_code(
+        cls,
+        ctx: "CContext",
+        ftype: "MallocBufferFType",
+        inline: bool = False,
+    ):
```
mtsokol: Return types?

src/finchlite/codegen/malloc_buffer.py
```diff
@@ -1,53 +1,168 @@
...
+        ctx.add_header(libcode)
+        return methods
+
+    @classmethod
+    def library(cls, ftype: "MallocBufferFType") -> CMallocBufferLibrary:
+        # lazy compile the library.
+        if ftype in cls._library:
+            return cls._library[ftype]
+        ctx = CContext()
+        methods = cls.gen_code(ctx, ftype)
```
mtsokol: I would prefer to arrange it a bit differently.

You call `gen_code` with `inline=True` in `c_unpack` method. In `library` you call without it. I assume that `def library` uses it to fetch the methods for `MallocBuffer.__init__`, `MallocBuffer.__del__` and `MallocBuffer.resize`. I assume that we want to call `gen_code` for code generation only once in known compilation phase, and other calls are only for fetching methods, right?

Can we decide where/when `gen_code` happens exactly? and call `gen_code` only there? In `c_unpack`?
Right now in some time in the future calling init, resize, and free will cause `gen_code` to be called three times.

If you have enough time, could you add one test which calls `MallocBuffer.__init__`, `__del__`, and `resize`?

Is there a test which calls `MallocBufferFType.c_unpack`?

junikimm717: Wait what

```python
        if ftype in cls._library:
            return cls._library[ftype]
```
            
Like this implies we won't ever generate code more than once right
(Well once to extract the methods and one as an inline method for the kernel)

mtsokol (approved):

PR #277, opened 2025-12-22
Kbd convert alias to tables

willow-ahrens:
@kylebd99 This PR is nice.

### DIFF
--- a/src/finchlite/interface/fuse.py
+++ b/src/finchlite/interface/fuse.py
@@ -52,6 +52,7 @@ Performance:
 
 from finchlite.autoschedule import LogicExecutor, LogicNormalizer
 from finchlite.autoschedule.formatter import LogicFormatter
+from finchlite.autoschedule.optimize import DefaultLogicOptimizer
 from finchlite.finch_logic.stages import LogicEvaluator
 from finchlite.finch_notation.interpreter import NotationInterpreter
 
@@ -78,7 +79,9 @@ _DEFAULT_SCHEDULER = None
 
 INTERPRET_LOGIC = LogicInterpreter()
 OPTIMIZE_LOGIC = LogicNormalizer(
-    LogicExecutor(LogicStandardizer(LogicFormatter(MockLogicLoader())))
+    LogicExecutor(
+        DefaultLogicOptimizer(LogicStandardizer(LogicFormatter(MockLogicLoader())))
+    )
 )
 INTERPRET_NOTATION = LogicNormalizer(
     LogicExecutor(


willow-ahrens:
I'll try to fix the test_scheduler failures

willow-ahrens:
I need to make a few final tweaks to the optimizer, then this will be ready.

tests/test_logic_interpreter.py
```diff
@@ -80,25 +106,40 @@ def test_materialize():
...
-    expected = TableValue(
-        np.array([[((1 + 1) * 1), ((2 + 1) * 2)], [((3 + 1) * 3), ((4 + 1) * 4)]]),
-        (i, j),
+    expected = fl.asarray(
+        np.array([[((1 + 1) * 1), ((2 + 1) * 2)], [((3 + 1) * 3), ((4 + 1) * 4)]])
     )
 
-    assert result == expected
-    finch_assert_equal(C, np.array([[1, 1], [1, 1]]))
+    assert (result.to_numpy() == expected.to_numpy()).all()
```
mtsokol: I'd prefer `numpy.testing.assert_array_equal` instead.

tests/reference/test_scheduler_e2e_sddmm_plan.txt
```diff
@@ -0,0 +1,4 @@
+A_4 = Aggregate(add, 0, Reorder(MapJoin(mul, MapJoin(mul, Table(A_2, i, i_2), Table(A_3, i_3, i_2)), Table(A, i_3, i)), i_3, i, i_2), i_2)
```
mtsokol: Shouldn't `MapJoin` of `i, i_3, i_2` and `i_3, i` be concordant? Do we allow it?

tests/test_scheduler.py
```diff
@@ -543,302 +437,85 @@ def test_flatten_plans():
     assert result == expected
 
 
-def test_propagate_map_queries_backward():
```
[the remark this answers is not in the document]
willow-ahrens: I was just being lazy, I knew that the sddmm test and the matmul tests would stress these and I was running out of time to finish the PR. If you'd like to add back those tests that would be super helpful.

tests/test_logic_interpreter.py
willow-ahrens: I think we bounced away from that because it was kinda tricky to overload it for BufferizedNDArray. Can you get assert_array_equal to work with our overloaded tensors?

tests/reference/test_scheduler_e2e_sddmm_plan.txt
willow-ahrens: I think I had set up this example to require a transpose for someone.

PR #279, opened 2025-12-24
TEST: Reset name counts

mtsokol:
A test helper function for resetting counter value in aliases to avoid regenerating reference files each time a new test is introduced.


willow-ahrens (approved):
This is good, and we should also remove it once we solve https://github.com/finch-tensor/finch-tensor-lite/issues/280

PR #283, opened 2026-01-06
tweak the level/fiber interface (starting with dense levels)

willow-ahrens:
@mtsokol, we'll talk about this at the meeting, I was wondering if you could finish this PR starting from this set of changes?

### DIFF
--- a/src/finchlite/tensor/fiber_tensor.py
+++ b/src/finchlite/tensor/fiber_tensor.py
@@ -77,6 +77,9 @@ class LevelFType(FType, ABC):
         ...
 
     @abstractmethod
+    def get_fields_class(self, tns, buf_s, pos, op, dirty_bit): ...
+
+    @abstractmethod
     def level_unfurl(self, ctx, tns, ext, mode, proto, pos):
         """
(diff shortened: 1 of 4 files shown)


src/finchlite/tensor/fiber_tensor.py
```diff
@@ -273,73 +340,43 @@ def buffer_factory(self):
...
 
-    def lower_unwrap(self, ctx, obj):
-        raise NotImplementedError
+    def lower_unwrap(self, ctx, tns):
+        return self.lvl_t.level_lower_unwrap(ctx, tns.obj, tns.obj.pos)
 
-    def lower_increment(self, ctx, obj, val):
-        raise NotImplementedError
+    def lower_increment(self, ctx, tns, val):
+        return self.lvl_t.level_lower_increment(ctx, tns.obj, val, tns.obj.pos)
```
mtsokol: If we store `pos` in `tns.obj` which is `FiberTensorField`s then we don't need a separate `pos` parameter, right?

mtsokol:
I'm going to merge it tonight to unblock other PRs.

PR #285, opened 2026-01-08
Update matmul example

mtsokol:
Makes sure `A @ B` with Numba backend works again.


src/finchlite/autoschedule/optimize.py
```diff
@@ -323,7 +335,15 @@ def _heuristic_loop_order(root: LogicExpression) -> tuple[Field, ...]:
             case Reorder(Table(_, idxs_1), idxs_2):
                 chains.append(list(intersect(intersect(idxs_1, idxs_2), root.fields())))
     chains.extend([f] for f in root.fields())
-    result = toposort(chains)
+
+    try:
+        result = toposort(chains)
```
mtsokol: Some einsum tests don't use concordant indices so we need an escape with a warning instead of failing optimization.

PR #294, opened 2026-01-14
Rs/get lazy tensor stats

Ruchika-2003:
I made changes to the tensor_def, dense_stats and also the test cases for dense_stats

### DIFF
--- a/src/finchlite/galley/LogicalOptimizer/logic_to_stats.py
+++ b/src/finchlite/galley/LogicalOptimizer/logic_to_stats.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections import OrderedDict
 
 from ...finch_logic import (
+    Plan,
     Aggregate,
     Alias,
     Field,
@@ -79,13 +80,22 @@ def insert_statistics(
                     raise ValueError(f"No TensorStats bound to alias {node.tns}")
                 
                 new_indices = tuple(f.name for f in node.idxs)
-                tensor = ST.relabel(node.tns,new_indices)
+                tensor = ST.relabel(base_stats,new_indices)
 
             if (node not in cache) or replace:
                 cache[node] = tensor
             return cache[node]
+        
+        case Plan():
+            last_result = None 
+            for body in node.bodies:
+                last_result = insert_statistics(ST,body,bindings,replace,cache)
+            return last_result
 
+        case _:
+            raise TypeError(f"Unhandled node type: {type(node)}")
 
+'''
 def get_lazy_tensor_stats(
     lazy_tensor: LazyTensor, StatsImpl: TensorStats
 ) -> TensorStats:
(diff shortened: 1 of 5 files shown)


src/finchlite/galley/TensorStats/dense_stat.py
```diff
@@ -29,12 +29,12 @@ def estimate_non_fill_values(self) -> float:
 
     @staticmethod
     def mapjoin(op: Callable, *args: TensorStats) -> TensorStats:
```
kylebd99: I think that the existing version of DenseStats was a bit wrong here. Rather than manually creating a new TensorDef in the DenseStats mapjoin and aggregate functions, we should be calling the mapjoin and aggregate functions in the TensorDef class to get "new_def".

src/finchlite/galley/LogicalOptimizer/logic_to_stats.py
```diff
@@ -27,66 +28,72 @@ def insert_statistics(
...
+        case Reorder():
+            child = insert_statistics(ST, node.arg, bindings, replace, cache)
+            cache[node] = child
+            return child
+        
+        case Table():
+            if isinstance(node.tns, Literal):
+                idxs = [f.name for f in node.idxs]
+                tensor = ST(node.tns.val, idxs)
+            elif isinstance(node.tns, Alias):
```
kylebd99: The statistics object associated with an alias won't necessarily have the correct set of fields. To fix this, you'll want to call relabel on the statistics returned from bindings.

kylebd99: i.e. the fields may be named something other than node.idxs

src/finchlite/galley/TensorStats/tensor_def.py
```diff
@@ -49,7 +54,7 @@ def from_tensor(cls, tensor: Any, indices: Iterable[str]) -> "TensorDef":
             fill_value=fv,
         )
 
-    def reindex_def(self, new_axis: Iterable[str]) -> "TensorDef":
+    def reindex_def(self, new_axis:tuple[str, ...]) -> "TensorDef":
```
kylebd99: I'm tempted to move this to the bottom of the file, rename it to reorder and also make a relabel function. This would allow us to just call the TensorDef reorder and relabel functions to get the new def within the DenseStats and DCStats' reorder and relabel functions.

src/finchlite/galley/TensorStats/tensor_def.py
```diff
@@ -6,17 +6,21 @@
...
+    Table,
+)
 from ...algebra import fill_value, is_idempotent, is_identity
 
 
 class TensorDef:
     def __init__(
         self,
-        index_set: Iterable[str],
+        index_set : tuple[str, ...],
```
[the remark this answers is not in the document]
Ruchika-2003: Okay I'll do that

tests/test_galley_tensor_stats.py
```diff
@@ -1357,3 +1367,374 @@ def test_varied_reduce_DC_card(dims, dcs, reduce_indices, expected_nnz):
     )
 
     assert reduce_stats.estimate_non_fill_values() == expected_nnz
+
```
kylebd99: I don't think that the Annotated Query Tests should be here. I think this is a merge artifact.

src/finchlite/galley/TensorStats/tensor_def.py
Ruchika-2003: Hmm okay, I'll try working on that

willow-ahrens:
@Ruchika-2003 is this stale?

willow-ahrens:
@Ruchika-2003 is this stale?

PR #302, opened 2026-01-27
Wma/move pos

willow-ahrens:
move pos out of level fields

### DIFF
--- a/src/finchlite/__init__.py
+++ b/src/finchlite/__init__.py
@@ -1,4 +1,4 @@
-from .algebra import Tensor, TensorFType, element_type, fill_value, shape_type
+from .algebra import Tensor, TensorFType
 from .codegen import (
     NumpyBuffer,
     NumpyBufferFType,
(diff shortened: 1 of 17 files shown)


mtsokol:
@willow-ahrens The only question I have is about `buf_s` in FiberTensorFields.
We create a slot variable for buffer in `asm_unpack()` which calls `level_asm_unpack()` of ElementLevelFType. At that moment we don't have ElementLevelFields to store it so:

- We can keep the slot in FiberTensorFields in `buf_s` as right now,
- We get rid of slot variable for `buf_s` and access it with `...lvl.val` (but we lose slot feature). Then we can actually skip unpack for levels and repack for fiber_tensor completely:
  - I realized that storing numpy buffer in a list to support resizing of `[arr]` (where we do `arr = slot[0]` and `slot[0] = arr`, so that `arr = np.resize(arr, size)` works in Numba) isn't really needed with `...lvl.val` - instead resizing can be done with `...lvl.val = np.resize(...lvl.val, size)`.

Should we keep `buf_s` in FiberTensorFields and slots with `[arr]`?

willow-ahrens (commented):
Looks great, thanks!

PR #308, opened 2026-02-01
Add Support for Indirect Access in Pointwise Interpreter

TheRealMichaelWang:
Per discussion from last week, I implemented `ein.Access` using numpy advanced indexing which added support for bare indices (size still has to be inferred from an `ein.Access`). More specifically, 
- A global (relative to the pointwise interpreter) dictionary `loop_sizes` has been added to store inferred sizes of indicies that are later re-used as bare-indicies
- Support for bare indicies (i.e `a[i + 4]` or `a[i, j] + (i * j)`) is now implemented.
- `ein.Access` case for indirect access uses numpy-style advanced indexing
  - Function that computes tuple of tensors used for indexing is implemented as a separate method in the pointwise interpreter, called `map_idxs`. This method can/will be later re-used to implement indirect access.
  - Implementation `ein.Access` case for indirect access is now 4 statements, excluding asserts.
- Refactored implementation for ein.GetAttr to use 1 case implementation and to use `hasAttr` and `getAttr` methods instead to get properties from python object.

### DIFF
--- a/src/finchlite/finch_einsum/interpreter.py
+++ b/src/finchlite/finch_einsum/interpreter.py
@@ -109,13 +109,15 @@ class TensorEinsumMachine:
                     raise ValueError(f"Unbound variable: {name}")
                 return self.bindings[node]
             case ein.Access(tns, (ein.Literal(dim),)):
-                tns = self(tns)
-                return tns[dim]
+                evaled: Any = self(tns)
+                return evaled[dim]
             case ein.GetAttr(obj, ein.Literal(attr)):
-                obj = self(obj)
-                if not hasattr(obj, attr):
+                evaled = self(obj)
+                if not hasattr(evaled, attr):
                     raise ValueError(f"Object {obj} has no attribute {attr}")
-                return getattr(obj, attr)
+                return getattr(evaled, attr)
+            case _:
+                raise ValueError(f"Unknown tensor einsum type: {type(node)}")
 
 
 class PointwiseEinsumMachine:


TheRealMichaelWang:
@willow-ahrens  I've completed the todos by adding back the reference implementation of `ein.Access` case in the pointwise interpreter and removing the `dim` field in the `ein.GetAttr` node.

I've also changed the orientation of how the coordinate array was stored in sparse tensor: it used to be transposed for readability-purposes where each row was an individual coordinate and each column represents a dim. Now that transposing is removed and allows for cleaner access with `ein.Access(ein.GetAttr(tns, "coords"), ein.Literal(dim))`

src/finchlite/finch_einsum/interpreter.py
```diff
@@ -82,48 +84,105 @@ def __init__(self, xp=np, verbose=False):
...
+                    vals = [xp.astype(v, common_dtype) for v in vals]
                 return func(*vals)
-            case ein.Access(tns, idxs):
-                assert len(idxs) == len(set(idxs))
+            case ein.Index(_) as idx:
+                tns = self.xp.arange(self.dims[idx])
+                for _ in range(len(self.loops) - self.loops.index(idx) - 1):
+                    tns = self.xp.expand_dims(tns, -1)
+                return tns
+            case ein.Access(tns, (ein.Literal(dim),)):
```
willow-ahrens: The case for literals should handle this? when we call self(literal), does the result not work as an index?

src/finchlite/finch_einsum/interpreter.py
```diff
@@ -82,48 +84,105 @@ def __init__(self, xp=np, verbose=False):
...
+        self.tns_ctx = TensorEinsumMachine(bindings)
 
     def __call__(self, node):
         xp = self.xp
         match node:
             case ein.Literal(val):
+                # If val is already an array, return it directly
+                # (e.g., from recursive access).
+                # Otherwise, broadcast the scalar to match the loop dimensions
+                if hasattr(val, "ndim") and val.ndim > 0:
```
willow-ahrens: if literal is handled differently when it's a tensor, put this special case in the tns_ctx. maybe this is dead code already?

src/finchlite/finch_einsum/interpreter.py
```diff
@@ -82,48 +84,105 @@ def __init__(self, xp=np, verbose=False):
...
+    def __init__(self, bindings):
+        self.bindings = bindings
+
+    def __call__(self, node):
+        match node:
+            case ein.Alias(name):
+                if node not in self.bindings:
+                    raise ValueError(f"Unbound variable: {name}")
+                return self.bindings[node]
+            case ein.Access(tns, (ein.Literal(dim),)):
```
willow-ahrens: can we generalize to any set of literals or Colon?

tests/test_einsum.py
```diff
@@ -1,17 +1,63 @@
...
+        prgm_obj = ein.Plan((stmt, ein.Produces((stmt.tns,))))
+        ctx = ein.EinsumInterpreter(xp=np)  # Use numpy instead of lazy module
+        bindings = {ein.Alias(k): v for k, v in kwargs.items()}
+        return ctx(prgm_obj, bindings)[0]
+
+    def patched_einsum(subscripts, *args, **kwargs):
+        stmt, bindings = ein.parse_einsum(subscripts, *args)
+        prgm_obj = ein.Plan((stmt, ein.Produces((stmt.tns,))))
+        ctx = ein.EinsumInterpreter(xp=np)  # Use numpy instead of lazy module
+        return ctx(prgm_obj, bindings)[0]
```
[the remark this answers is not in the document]
TheRealMichaelWang: ? This code works fine? Bindings is now computed within parse_einsum.

willow-ahrens:
moved to https://github.com/finch-tensor/insum-tensor

PR #314, opened 2026-02-05
DEV/CI: pixi-build setup & Pixi for CI and dev docs

### LINKED ISSUE #535: Switch to Pixi, add optional gcc backends

lucascolley:


lucascolley:
hopefully that fixes CI

pixi.toml
```diff
@@ -1,23 +1,71 @@
...
+
+[package.host-dependencies]
+python = ">=3.10,<3.13"
+poetry-core = "*"
+
+[package.run-dependencies]
+numpy = "*"
+numba = "*"
+donfig = "*"
+lark = "*"
```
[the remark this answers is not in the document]
lucascolley: > > So we need to make the C compiler tests opt-in or automatically skipped when there isn't one available?
> 
> yes.

hey @willow-ahrens @mtsokol ! Just wondering if there is any update on this?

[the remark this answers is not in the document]
lucascolley: (if there is some actionable work I can try to help out!)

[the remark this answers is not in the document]
lucascolley: gentle bump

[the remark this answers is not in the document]
willow-ahrens: Thanks for the reminder, I'm still working on this! We're reorganizing the github org and projects a little bit lately. We're resolved now to switch to pixi, make dep on GCC optional, and also add an optional julia dep.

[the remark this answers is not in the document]
willow-ahrens: I may need a little help getting the gcc dep to work, but I'll reach out if it's an issue.

.github/workflows/benchmark.yml
```diff
@@ -7,17 +7,12 @@ jobs:
...
-      - name: Install package
-        run: |
-          poetry install --extras benchmark --extras test
+          pixi-version: v0.71.2
+          environments: >-
+            default
+            ${{ matrix.environment }}
       - name: Install debug symbols
         run: |
           sudo apt-get update && sudo apt-get install -y libc6-dbg
```
lucascolley: not sure if this does anything now

.github/workflows/publish.yml
```diff
@@ -18,10 +18,10 @@ jobs:
...
             if [ $? -ne 0 ]; then exit 1; fi
             echo "VERSION=$VERSION" >> $GITHUB_ENV
-      - name: Build and publish to pypi
-        uses: JRubics/poetry-publish@v1.17
-        with:
-          pypi_token: ${{ secrets.PYPI_TOKEN }}
+      - name: Authenticate with Prefix.dev
+        run: pixi auth login --token ${{ secrets.PREFIX_API_KEY }} https://prefix.dev
+      - name: Build and Publish Package
+        run: pixi publish --target-channel https://prefix.dev{{ secrets.PREFIX_CHANNEL_NAME }}
```
lucascolley: maybe you still want to publish to PyPI... but this is cool too :)

willow-ahrens: I've added back the pypi step, but it's through pypa now, not poetry. pixi doesn't have a pypi step, does it?

hameerabbasi (approved):

pyproject.toml
```diff
@@ -29,9 +25,115 @@ benchmark = [
     "pytest-codspeed>=5",
 ]
 
+[tool.pixi.workspace]
```
lucascolley, line 36: Any reason you prefer putting this in `pyproject.toml`?

pyproject.toml
```diff
@@ -29,9 +25,115 @@ benchmark = [
...
+
+[tool.pixi.package.host-dependencies]
+python = ">=3.10,<3.13"
+setuptools = "*"
+
+[tool.pixi.package.run-dependencies]
+numpy = ">=1.19,<2.4"
+numba = ">=0.61,<0.63.1"
+donfig = ">=0.8.1.post1,<0.9.0"
+lark = ">=1.3.0,<2.0.0"
```
lucascolley: In a previous commit, I had added `package.build.config.ignore-pypi-mapping = false`. That would reduce duplication here.

pyproject.toml
```diff
@@ -29,9 +25,115 @@ benchmark = [
...
-requires = ["poetry-core>=1.0.8"]
-build-backend = "poetry.core.masonry.api"
+requires = ["setuptools>=77"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
+[tool.setuptools.package-data]
+finchlite = ["py.typed"]
```
lucascolley: Any reason to prefer setuptools over uv_build?

willow-ahrens: from the uv docs: "The uv build backend currently only supports pure Python code. An alternative backend is required to build a [library with extension modules](https://docs.astral.sh/uv/concepts/projects/init/#projects-with-extension-modules)."

I believe we plan to add mlir bindings as another backend, which would need a extension module right?

willow-ahrens: https://docs.astral.sh/uv/concepts/build-backend/#choosing-a-build-backend

hameerabbasi:
@lucascolley I've just focused on the C-compilation issues, feel free to push to the branch with your suggestions.

pyproject.toml
willow-ahrens: Yes, I didn't want to duplicate the toplevel toml values (package name, etc.). Is it more standard to use pixi.toml separately? How much do each of these things need to stay in sync?

pyproject.toml
lucascolley: thanks! Looks like they recommend hatchling instead, but not sure how that compares to setuptools

pyproject.toml
willow-ahrens: Oh, I see. That would indeed be nice. We should do that, sorry for the churn.

willow-ahrens:
@hameerabbasi Thank you so much for windows C fixes!

pyproject.toml
lucascolley: Probably for many fields you can just keep them in pyproject and omit them from pixi.toml.

pixi-build-python automatically discovers the package version from pyproject for example, see https://github.com/data-apis/array-api-extra/blob/main/pixi.toml

pyproject.toml
hameerabbasi: When we have MLIR, `scikit-build-core` or `pixi-build-cmake` is probably the best thing to use. But for now, anything does.

pyproject.toml
lucascolley: > Is it more standard to use pixi.toml separately?

It's still a matter of personal preference. In my experience, `tool.pixi` is great when the full pyproject fits in 1 or 2 editor pages, but a separate file is nicer for larger projects.

willow-ahrens: I think, in particular, I don't want to duplicate version.

lucascolley: Then good news, you don't need to :)

pyproject.toml
willow-ahrens: okay. we can start with pixi-build-cmake then.

pyproject.toml
willow-ahrens: Sorry, github was buffering and I couldn't see your earlier message about that! I'll probably keep it all centralized as much as I can for this PR.

pyproject.toml
lucascolley: I don't think it is a good idea to have both `setuptools` and CMakeLists producing different Python packages in the same repo at the same time. I would recommend just using `pixi-build-python` with `setuptools` or `scikit-build-core` for now, then if you want to ship a _non_-Python package (e.g. just C parts), you can use `pixi-build-cmake` in a different directory to define a different package.

CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+cmake_minimum_required(VERSION 3.20)
```
hameerabbasi: This is quite a convoluted way of doing things for pure Python. I'd suggest uv_build unless we need compiled extensions.

hameerabbasi (changes requested):

CMakeLists.txt
lucascolley: agreed per my latest comment. Fine to use any Python build backend for now, but definitely stick with `pixi-build-python`.

lucascolley: I'll push a commit

lucascolley:
since the job names have changed, you may need to adjust the required status checks in the repo settings

hameerabbasi (approved):
@willow-ahrens Someone with admin permissions needs to edit the required jobs in the repo settings; before this can be merged.

willow-ahrens:
I had already edited, so it looks like it merged!

willow-ahrens:
Huge thank you to @lucascolley! Is there anything we need to do over on https://github.com/conda-forge/staged-recipes/pull/32124?

lucascolley:
I will take a look at updating it!

lucascolley:
ah, it would be good to make a release to PyPI before we proceed with https://github.com/conda-forge/staged-recipes/pull/32124 ! Would it be a good time for a release soon @willow-ahrens ?

willow-ahrens:
@lucascolley The plan has been to rename this repo to finch-tensor once it was ready to switch over. I think it's soon at the point! (There's still going to be bugs, but it's better to get things lined up to where we want them to be, nothing is 1.0 yet). I've added you in the tracking issue for this change here: https://github.com/finch-tensor/finch-tensor-lite/issues/554

PR #325, opened 2026-02-15
Rs/implementing uniform stats

Ruchika-2003:


kylebd99 (approved):
I went through and did some minor changes to the __init__.py files to clean up exports, and I adjusted test_galley_tensor_stats to use Fields in the pytest params rather than converting strings to fields within functions. Overall though, this looks good! The functionality all looks correct to me, and it should put us on a good footing going forward.

Issue #259, opened 2025-12-09
Wrapper operators


willow-ahrens:
Instead of properties like return_type, isassociative, we implement a, abstract FinchOperator Class with a bunch of properties for each method.

willow-ahrens:
Remaining task: Abstract operator classes for Numba and C operators

PR #330, opened 2026-02-17
Jx/wrapper operators

jxu44:
addresses #259

### DIFF
--- a/src/finchlite/interface/eager.py
+++ b/src/finchlite/interface/eager.py
@@ -4,7 +4,6 @@ from abc import ABC
 from collections.abc import Callable, Sequence
 from typing import Any
 
-from ..algebra import register_property
 from . import lazy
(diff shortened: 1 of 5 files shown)


src/finchlite/algebra/algebra.py
```diff
@@ -198,274 +200,752 @@ def promote_type_stable(a, b) -> type:
...
 
-    return _return_type_closure
+class ReflexiveFinchOperator(FinchOperator):
 
+    reflected_method: str
 
-def _return_type_reflexive_operator(meth, rmeth):
-    def _return_type_closure(op, a, b):
-        if hasattr(a, meth):
+    def return_type(self, a: Any, b: Any):
```
willow-ahrens: here, just return "type(self(a(True), b(True)))"

PR #334, opened 2026-02-27
Add galley optimizer

HoangNguyen0309:
galleyOptimizer implmentation.

Works on its own, fails in pipeline (see test_galleyOptimzer 1 to 5)

Added galleyOptimizer into pipleline in fuze.py: circular import error, current fix is wack.

### DIFF
--- a/src/finchlite/autoschedule/galley/logical_optimizer/annotated_query.py
+++ b/src/finchlite/autoschedule/galley/logical_optimizer/annotated_query.py
@@ -86,11 +86,12 @@ class AnnotatedQuery:
         self.cache = cache
         output_name = q.lhs
         expr = q.rhs
+        output_order: list[Field] = []
         if isinstance(expr, Reorder):
-            output_order = list(q.rhs.arg.fields())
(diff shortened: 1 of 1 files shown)


HoangNguyen0309:
Bug: Galley optimizer fails on A @ B @ C with MapJoin expects at least one argument with stats

Example:
A = asarray(shape=(100, 2))   
# A @ B @ C
B = asarray(shape=(2, 100))
C = asarray(shape=(100, 2))
out = compute((lazy(A) @ lazy(B)) @ lazy(C), ctx=INTERPRET_NOTATION_GALLEY)

After normalization, query is: Aggregate(add, 0, MapJoin(mul, MapJoin(mul, A, B), C), j, k),
the greedy optimizer raises ValueError: MapJoin expects at least one argument with stats during reduce_idx.

In get_reduce_query, when reducing over an index that only one MapJoin argument has, the code builds MapJoin(mul, [single_arg]), which is valid.

Fix should be somewhere in annotedQuery as it is not processing the new normalized query but still follows old unnormalized one.

PR #343, opened 2026-03-22
galley components and benchmarking

HoangNguyen0309:


src/finchlite/interface/__init__.py
```diff
@@ -105,6 +105,7 @@
     INTERPRET_NOTATION,
     INTERPRET_NOTATION_GALLEY,
     OPTIMIZE_LOGIC,
+    INTERPRET_NOTATION_GALLEY_NO_COMPONENTS, # TESTING PIPELINE
```
kylebd99: see my note in fuse.py

src/finchlite/autoschedule/galley/logical_optimizer/greedy_optimizer.py
```diff
@@ -1,41 +1,39 @@
...
-    # USING COMPONENTS:
-    if use_components:
-        component_idx = 0
-        while component_idx < len(aq.connected_components):
-            component = aq.connected_components[component_idx]
-            reducible_idxs = aq.get_reducible_idxs_for_component(component)
-            if not reducible_idxs:
+    component_idx = 0
+    while True:
+        components = aq.connected_components
```
kylebd99: If the list of components is being recalculated every time, then we shouldn't need to keep a component_idx variable. We can just use the first component in the list. In general, keeping an index into a list whose size is changing is an anti-pattern. This should also mean that you can remove lines 23-25 & 27.

PR #344, opened 2026-03-23
Support multiple steppers

mtsokol:
This PR:
- Change the semantics of dim (for dense and sparse levels) from dimension to "last index to visit", so that `Extent(start, end)` includes both ends. It makes (for me) it simple to reason about steppers (we know we visit `last_i`).
- Stepper now supports multiple steppers, and generates all combinations of possible intersections. 

Next tasks after this PR:
- Make `scansearch` and `resize_if_smaller` lower separately to C and Numba
- Simplifications pass for Notation IR - to actually start eliminating unnecessary sub-loops generated in stepper pass.
- Writable sparse_list_level.

@willow-ahrens Would you have time to take a look? It's a thorough rewrite of StepperPass and sparse_list_level so not necessarily worth a deep review - can be just approve.

### DIFF
--- a/src/finchlite/compile/looplets.py
+++ b/src/finchlite/compile/looplets.py
@@ -1,15 +1,12 @@
 from abc import ABC, abstractmethod
 from collections.abc import Callable
 from dataclasses import dataclass
-from functools import reduce
+from functools import partial, reduce
 from typing import Any
-
-import numpy as np
 
 from .. import finch_assembly as asm
 from .. import finch_notation as ntn
 from ..algebra import ffunc
-from ..compile.extents import intersect_extents
 from ..compile.lower import (
     LoopletContext,
     LoopletPass,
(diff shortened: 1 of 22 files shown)


willow-ahrens:
I'm still very much in favor of picking the range convention that's used by python, simply because then when we explain why we chose that, it's very hard for others to argue with "we adopt the same range convention as python," as well as keeping an internal consistency with Tensor, which starts at 0 and end at n, but contain 0 to n - 1. In my opinion, the arguments e.g. for and against 0 or 1 indexing are not as strong as the argument to keep things consistent. Once we rewrite the julia implementation, we're not planning to continue support for it, so keeping consistent with the julia implementation isn't a priority going forward.
For reference, triton also appears to stick with python semantics:

https://triton-lang.org/main/python-api/generated/triton.language.arange.html#triton.language.arange

I'm having a hard time finding any other framework which uses inclusive intervals.

willow-ahrens:
It sounds like we're resolved to keep the right endpoints exclusive for consistency with other compilers. We can maybe add this to the docs for extent.

Thus, a singleton Extent would be Extent(a, a+1), and when we subtract stop-start we get the number of items in the range

mtsokol:
It's done now - Extents are right-end exclusive.

src/finchlite/compile/looplets.py
```diff
@@ -209,70 +153,59 @@ def stepper_body(ctx, node: ntn.NotationNode, total_ext, chunk_ext):
                     ctx.exec(st.seek(ctx, ext))
                     steppers.append(st)
 
-        ctx_2 = ctx.scope()
-
-        idx_start = ntn.Variable(ctx.freshen("i_start"), idx.result_format)
-        ctx.exec(asm.Assign(ctx.ctx(idx_start), ctx.ctx(ext.get_start())))
+        full_body = Rewrite(PostWalk(lambda node: stepper_body(ctx, node)))(body)
```
willow-ahrens, line 156: Do we need to truncate any of the looplets, or are we assuming looplets always have infinite range?

src/finchlite/tensor/level/sparse_list_level.py
```diff
@@ -235,13 +250,33 @@ def thunk_preamble(ctx, idx):
...
         def seek_fn(ctx, ext):
-            code = """finch
-            if (idx_s[q] < {start})
-              q = scansearch(idx_s, {start}, q, q_stop - 1)
-            end
-            """.format(start=ctx.ctx(ext.get_start()))
-
-            return parse_assembly(code, tmp_locals)
+            start = ctx.ctx(ext.get_start())
+            return asm.Block(
```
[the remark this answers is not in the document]
willow-ahrens: what do you mean by this?

src/finchlite/compile/looplets.py
```diff
@@ -331,44 +263,74 @@ class SequencePass(LoopletPass):
     def priority(self):
         return 4
 
+    @classmethod
+    def get_sequence_variations(
+        cls, ctx, seqs: list[Sequence], heads, tails, ext: SymbolicExtent
+    ) -> list[tuple]:
+        ext_start, ext_end = ext.get_start(), ext.get_end()
+
+        match seqs:
```
willow-ahrens, line 272: this is kindof a cool function I like it. I wish the julia version was this clean.

willow-ahrens (approved):
I have a few comments, I think I'm okay to merge if you are, I'm just wanting to stay up to date with how you're thinking about this.

src/finchlite/tensor/level/sparse_list_level.py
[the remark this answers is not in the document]
mtsokol: Oh, outdated for sure - I reverted that change - and used our assembly parser again.

src/finchlite/compile/looplets.py
mtsokol: As I understand it now we don't need truncating for steppers/sequences and in general conceptually we're closer to an infinite range. We have conditions, that make sure we don't jump out of extension we're oprerating on, like:
```
if lt(q, q_stop):
    i_stop: int64 = load(slot(#_A#15_idx_slot, np_buf_t(int64)), q)
else:
    i_stop: int64 = #_A#15.lvl.lvl.dimension
```

mtsokol:
> I have a few comments, I think I'm okay to merge if you are, I'm just wanting to stay up to date with how you're thinking about this.

I'll merge it then and move to #371.

PR #347, opened 2026-03-24
 Improved @fused decorator

### LINKED ISSUE #317: Expand Fused to Handle Control Flow
In other JIT decorator approaches like PyTorch, basic control flow like while, for, and if is allowed. It would be great to support this functionality in our @fused decorator as well. 

The basic approach to this would be to automatically analyze the straight-line sections of the code, extract them as sub-functions, and perform fusion on them. For example, 
```
@fused
def example(A,B, out):
            out = A @ B
            for i in range(1,100):
                      out = np.multiply(out, ( A @ B))
            return out
```
Should be automatically turned into:
```
@fused
def f1(A,B):
        return A @ B

@fused
def f2(out, A,B):
           return np.multiply(out, ( A @ B))

def example(A,B, out):
            out = f1(A,B)
            for i in range(1,100):
                      out = f2(out, A, B)
            return out
```

One benefit of this is that this would allow us to eventually capture recursion during our program analysis. For example, we could pull unnecessary/repeated operations out of for-loops.

kylebd99:


kylebd99:
@willow-ahrens module functions now work, I'm good to merge after a brief look-over

benchmarks/jit_benchmarks.py
```diff
@@ -0,0 +1,44 @@
+import time
```
willow-ahrens, line 1: I think we should probably use asv to benchmark finchlite. I'll make a separate issue for it

src/finchlite/algebra/ffunc.py
```diff
@@ -835,6 +835,17 @@ def __repr__(self) -> str:
...
+    def __call__(self, a):
+        return np.iscomplexobj(a)
+
+    def __repr__(self) -> str:
+        return "iscomplexobj"
+
+
+iscomplexobj = _IsComplexObj()
+
+
```
willow-ahrens, line 848: what is this for?

src/finchlite/finch_fused/parser.py
```diff
@@ -0,0 +1,450 @@
...
+    converter = _FusedToPythonAST()
+    func_def = ast.fix_missing_locations(converter.parse_function(function))
+    module_node = ast.Module(body=[func_def], type_ignores=[])
+    code = compile(module_node, "<fused>", "exec")
+    globals_dict: dict[str, Any] = {
+        "__builtins__": builtins,
+        "operator": operator,
+        **converter._extra_globals,
+    }
+    exec(code, globals_dict)  # noqa: S102
```
willow-ahrens, line 450: I'm not sure where to add this comment, but I think we need to make sure that we can handle qualified names, e.g. xp.reduce. Once that works I think we're good to merge

willow-ahrens (changes requested):
This looks great!

src/finchlite/algebra/ffunc.py
kylebd99: This got added to allow for a particular scalar conversion case. I'm trying to remember, but I think that this was necessary as part of allowing it to convert to `__index__` for the range(sum(B)) example.

src/finchlite/finch_fused/parser.py
kylebd99: This should work now, see test_jit_module_function and test_jit_local_module_function

willow-ahrens (approved):

PR #348, opened 2026-03-27
LogicLoader Stats and Stats Factories

Ruchika-2003:


willow-ahrens (changes requested):
I merged a fix on main to fix the galley bugs related to n-ary ops, then I fixed circular import errors, then I added a factory constructor pattern for the stats. Things look better now, but I'm short on time and I used a little AI to speed along the big refactoring. I've looked at every line that was generated, but I'd appreciate another close reading from someone else. 

Ruchika, I think the cacheloader doesn't have tests currently. Could you add tests for it?

src/finchlite/finch_logic/tensor_stats.py
kylebd99, line 1: A bit awkward to have TensorStats defined here rather than in the tensor_stats folder, but I imagine that it avoids some messy local imports for the type system. So, LGTM.

src/finchlite/finch_logic/tensor_stats.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    @abstractmethod
+    def fill_value(self) -> Any: ...
+
+
+T = TypeVar("T", bound=TensorStats)
+
+
+class StatsFactory(ABC, Generic[T]):
```
kylebd99, line 35: I like the move to a factory design. The static method approach where we just passed around types made more sense in julia and this feels more pythonic.

src/finchlite/finch_logic/stages.py
```diff
@@ -56,14 +61,16 @@ def __call__(
         self,
         term: lgc.LogicStatement,
         bindings: dict[lgc.Alias, TensorFType],
+        stats: dict[lgc.Alias, Any] | None = None,
```
kylebd99: Should 'Any' be TensorStats here?

src/finchlite/finch_logic/stages.py
kylebd99, line 43: Do we think that LogicTransform also wants to see the stats or is this intended for syntactic rewrites?

tests/test_galley_logical_optimizer.py
kylebd99, line 1: This is a small and different issue, but I should go back and make these tests more front-end oriented. All of the tests that manually construct AnnotatedQueries should probably use the actual constructor...

src/finchlite/autoschedule/executor.py
```diff
@@ -36,17 +53,25 @@ def rule_0(node):
...
+    def __init__(
+        self,
+        ctx: LogicLoader | None = None,
+        stats_factory: StatsFactory | None = None,
+    ):
         if ctx is None:
             ctx = DefaultLogicFormatter()
         self.ctx: LogicLoader = ctx
+        if stats_factory is None:
+            stats_factory = DenseStatsFactory()  # probably have dummy stats here
```
kylebd99: Honestly, I'm tempted to not make this and ctx both required parameters. It doesn't seem sensible for them to be undefined.

src/finchlite/autoschedule/executor.py
```diff
@@ -56,17 +81,32 @@ def __call__(
...
+            stats_factory=self.stats_factory,
+        )
 
         bindings = dict(zip(binding_ftypes.keys(), bindings.values(), strict=False))
 
         binding_shapes = dict[lgc.Field | None, int]()
         for var, tns in bindings.items():
             for idx, dim in zip(binding_idxs[var], tns.shape, strict=True):
                 if idx is not None:
+                    # Assigning shape for each dimension
```
kylebd99: nit: delete

src/finchlite/autoschedule/galley_optimize.py
kylebd99, line 95: Nit: Add TODO to make this a LogicLoader

kylebd99 (approved):
A couple of very small notes. But, overall this LGTM. I like the changes.

src/finchlite/finch_logic/stages.py
willow-ahrens: LogicTransform is ununsed, AFAIK. I think we should remove all the "transform" classes because they never really worked.

src/finchlite/autoschedule/galley_optimize.py
willow-ahrens: already an issue

PR #355, opened 2026-04-01
Wma/naryfix

willow-ahrens:
move all ops to a file ffunc.py, make everything use ffunc

### DIFF
--- a/src/finchlite/autoschedule/galley/logical_optimizer/annotated_query.py
+++ b/src/finchlite/autoschedule/galley/logical_optimizer/annotated_query.py
@@ -97,7 +97,7 @@ class AnnotatedQuery:
         idx_starting_root: OrderedDict[Field, LogicExpression] = OrderedDict()
         idx_top_order: OrderedDict[Field, int] = OrderedDict()
         top_counter = 1
-        idx_op: OrderedDict[Field, Any] = OrderedDict()
+        idx_op: OrderedDict[Field, FinchOperator] = OrderedDict()
         idx_init: OrderedDict[Field, Any] = OrderedDict()
 
         def aggregate_annotation_rule(node: LogicNode) -> LogicNode:
@@ -139,7 +139,7 @@ class AnnotatedQuery:
             stats_point = cache_point[point_expr]
             idx_dim_size = stats_point.dim_sizes[idx]
             lowest_roots = AnnotatedQuery.find_lowest_roots(
-                Literal(agg_op), idx, idx_starting_root[idx]
+                agg_op, idx, idx_starting_root[idx]
             )
             original_idx[idx] = idx
             if len(lowest_roots) == 1:
@@ -155,7 +155,6 @@ class AnnotatedQuery:
                         # If the lowest root doesn't contain the reduction index, we
                         # attempt to remove the reduction via a repeat_operator, i.e.
                         # ∑_i B_j = B_j*|Dom(i)|
-                        assert isinstance(agg_op, FinchOperator)
                         f = repeat_operator(agg_op)
                         if f is None:
                             continue
@@ -202,7 +201,6 @@ class AnnotatedQuery:
                     connected_idxs[idx1].add(idx2)
                 mergeable_agg_op = (
                     idx1_op == idx2_op
-                    and isinstance(idx1_op, FinchOperator)
                     and is_associative(idx1_op)
                     and is_commutative(idx1_op)
                 )
@@ -454,7 +452,7 @@ class AnnotatedQuery:
 
     @staticmethod
     def find_lowest_roots(
-        op: Literal, idx: Field, root: LogicExpression
+        op: FinchOperator, idx: Field, root: LogicExpression
     ) -> list[LogicExpression]:
         """
         Compute the lowest MapJoin / leaf nodes that a reduction over `idx` can be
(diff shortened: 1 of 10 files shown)


src/finchlite/compile/lower.py
```diff
@@ -98,16 +99,22 @@ def ftype(self):
         )
 
 
-class ExtentOp(FinchOperator, metaclass=SingletonMeta):
+class ExtentOp(FinchOperator):
     __qualname__ = "ExtentOp"  # TODO: unify with the rest of FinchOperators
 
+    def __init__(self) -> None:
+        self.__qualname__ = "ExtentOp"
```
mtsokol: Why do we need it when `__qualname__ = "ExtentOp"` is there for a class?

src/finchlite/finch_assembly/simplification.py
```diff
@@ -42,7 +47,7 @@ def simplify(cls, term: asm.AssemblyNode):
                     ),
                 )
             ) if s1 == s2 and idx1 == idx2:
-                if isinstance(op, InitWrite) and op.value == arg.val:
+                if op == ffunc.init_write(arg.val):
```
mtsokol, line 45: So only for `_InitWrite` we have `def __eq__` implemented? Isn't it better to have all FinchOperators as dataclasses?

src/finchlite/finch_einsum/interpreter.py
```diff
@@ -11,68 +9,68 @@
...
-    operator.or_: "bitwise_or",
-    operator.xor: "bitwise_xor",
-    operator.lshift: "bitwise_left_shift",
-    operator.rshift: "bitwise_right_shift",
-    np.logical_and: "logical_and",
-    np.logical_or: "logical_or",
-    np.logical_not: "logical_not",
-    promote_min: "minimum",
-    promote_max: "maximum",
+    ffunc.add: "add",
```
mtsokol, line 16: This all looks nice!

tests/reference/test_asm_cfg_printer_comprehensive.txt
```diff
@@ -6,7 +6,7 @@ compute_2: #succs=[compute_3]
     [28] x: int64 = x
     [29] y: int64 = y
     [0] temp: int64 = x
-    [1] temp: int64 = add(temp, y)
+    [1] temp: int64 = finchlite.ffunc.add(temp, y)
```
mtsokol: I would prefer to have it as `add(temp, y)` for readability, and for other ops. For each op it's 16 characters more.

src/finchlite/compile/lower.py
```diff
@@ -159,7 +166,7 @@ class SymbolicExtent(FTyped):
     @classmethod
     def from_notation(cls, node: ntn.NotationNode):
         match node:
-            case ntn.Call(ntn.Literal(op), (start, end)) if op is ExtentOp():
+            case ntn.Call(ntn.Literal(op), (start, end)) if isinstance(op, ExtentOp):
```
mtsokol: I think we should have only one `ExtentOp()` instance across the codebase. It's the same question about dataclasses.

src/finchlite/codegen/numba_codegen.py
```diff
@@ -608,11 +608,11 @@ def __call__(self, prgm: asm.AssemblyNode):
                     val_code,
                 )
                 return None
-            case asm.Call(asm.Literal(val), args):
-                finch_op = as_finch_operator(val)
-                if not isinstance(finch_op, NumbaOperator):
-                    raise TypeError(f"{finch_op} has no Numba representation.")
-                return finch_op.numba_literal(val, self, *args)
+            case asm.Call(asm.Literal(op), args):
+                assert isinstance(op, NumbaOperator)
```
mtsokol: I think we don't need that assertion.

src/finchlite/autoschedule/galley/logical_optimizer/annotated_query.py
```diff
@@ -541,7 +555,11 @@ def get_reduce_query(
 
         use_root = False
         match root_node:
-            case MapJoin(Literal(op), args) as mj if is_distributive(op, reduce_op):
+            case MapJoin(Literal(op), args) as mj if (
+                isinstance(op, FinchOperator)
+                and isinstance(reduce_op, FinchOperator)
```
mtsokol: With FinchOperator as dataclass we don't need that boilerplate.

mtsokol (changes requested):
@willow-ahrens I see that you removed `SingletonMeta`. I prefer to either have it back or make all operators dataclasses:

Right now `Add() == Add()` is `false` which I think is incorrect and will cause much confusion, e.g. when you want to check if `op == SomeOp()`.

With dataclasses we have the comparison and pattern matching solved.
With SingletonMeta we have comparison solved and we're certain there is only one instance occupying memory of that op. With dataclasses it's up to GC how many are live. 

--- 

And a second question about printing ops without `finchlite.ffunc` prefix.

tests/reference/test_asm_cfg_printer_comprehensive.txt
willow-ahrens: That's fine, but we need to adjust the eval function so that these ops are in scope then.

src/finchlite/finch_assembly/simplification.py
willow-ahrens: by default, I think FinchOperator should probably set `__eq__(x, y)` to `x is y`

src/finchlite/compile/lower.py
willow-ahrens: I've been changing a lot of functions, some things slipped past. I'll fix this

src/finchlite/compile/lower.py
willow-ahrens: Sorry, this one didn't get treated the same as the functions in ffunc.py, I need to fix that.

tests/reference/test_asm_cfg_printer_comprehensive.txt
mtsokol: We can have from finchlite.ffunc import * at the top

willow-ahrens: I passed it into the eval func scope dictionary explicitly

src/finchlite/finch_assembly/simplification.py
willow-ahrens: I think the default behavior does this already.

src/finchlite/codegen/numba_codegen.py
willow-ahrens: removed

src/finchlite/autoschedule/galley/logical_optimizer/annotated_query.py
willow-ahrens: Oh, I see. We could match FinchOperator as a type you're saying?

willow-ahrens: I'll have to try this.

mtsokol: We could have it as:
```py
case MapJoin(Literal(FinchOperator() as op), args) as mj if is_distributive(op, reduce_op)
```

src/finchlite/compile/lower.py
mtsokol: I think it's OK to leave it with `isinstance(op, _MakeExtent)` - this class comes from that file anyway.

PR #362, opened 2026-04-06
Galley Branch and Bound - DFS version

HoangNguyen0309:


src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -2,13 +2,111 @@
...
-from collections import OrderedDict
+import os
+from collections import OrderedDict, deque
+from functools import reduce
 
-from ....finch_logic import Query
+from ....finch_logic import Field, Query
 from .annotated_query import AnnotatedQuery
 
+# Set env ``GALLEY_PRINT_BNB_DFS_MEMO=1`` to trace DFS memo (verbose).
```
kylebd99: Rather than using env variables, we should use the python logging infrastructure. Look at util/logging.py for reference here.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -2,13 +2,111 @@
...
+_PRINT_BNB_DFS_MEMO = os.environ.get("GALLEY_PRINT_BNB_DFS_MEMO", "") == "1"
+# All-fork sibling prefix blocking in ``branch_and_bound_dfs`` (module doc).
+_PREFIX_BLOCK_DFS = os.environ.get("GALLEY_BNB_DFS_PREFIX_BLOCK", "1") == "1"
+
+
+def _aq_with_stats(aq: AnnotatedQuery) -> AnnotatedQuery:
```
kylebd99: Same comment from last time, we should use the built in AnnotatedQuery copy function instead of making a new function.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -2,13 +2,111 @@
...
+    as we might use it later.
+    """
+    c = aq.copy()
+    for attr in ("bindings", "cache", "cache_point"):
+        if hasattr(aq, attr):
+            setattr(c, attr, getattr(aq, attr))
+    return c
+
+
+def _reducible_idxs_for_component(
```
kylebd99: Same as last time, it's a bad code pattern to redefine functions. We should just use AnnotatedQuery's function for this and wrap the result in list() if needed.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -2,13 +2,111 @@
...
+    stage: str, n: int, memo: dict[frozenset, tuple[float, list]]
+) -> None:
+    """Print ``memo`` after each stack pop / memo assign (DFS ``k=inf``)."""
+    print(f"\n--- branch_and_bound_dfs memo {stage} #{n} ---")
+    for vk in sorted(memo.keys(), key=_vars_key_sort_key):
+        c, ord_ = memo[vk]
+        label = "{}" if not vk else "{" + ", ".join(sorted(repr(x) for x in vk)) + "}"
+        print(f"  {label} -> (cost={c}, order={[repr(x) for x in ord_]})")
+    print(f"  [n_entries={len(memo)}]")
+
```
kylebd99: DFS helper functions should be moved to right above branch_and_bound_dfs

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
...
+        if cost > _dfs_memo_cost(memo, vars_key):
+            continue
+
+        if _PREFIX_BLOCK_DFS and _dfs_order_prefix_blocked(order, blocked_prefixes):
+            continue
+
+        if _PREFIX_BLOCK_DFS:
+            w = sibling_witness_best.get(tuple(order), float("inf"))
+            if w < cost:
+                continue
```
kylebd99: I think there's an issue with how the pruning is being implemented here. Specifically, when you're looking for a "sibling witness" you seem to be looking for previously plans that differ in the last index or two. Instead, we should be looking for previously discovered plans that cover a superset of the currently considered index set. For example, suppose that we have discovered that the cost of i->j->k->l is 100. Then, later in the search, we consider the plan i->l and find it has cost 200. We should realize that i->j->k->l is a superset of the indices in i->l and prune the i->l branch. So, we don't really care about "siblings" so much as we care about supersets.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
     return (optimal_orders[component_set], optimal_subquery_costs)
 
 
+def branch_and_bound_dfs(
+    input_aq: AnnotatedQuery,
+    component: list,
+    k: float,
+    max_subquery_costs: OrderedDict,
+) -> tuple:
```
kylebd99, line 129: I think all of this code should probably live in a new file as well. E.g. you could split this file into branch_and_bound_bfs.py and branch_and_bound_dfs.py.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -205,7 +496,7 @@ def exact_query_to_plan(input_aq: AnnotatedQuery) -> tuple[list[Query], float]:
 
     # --- Rebuild queries by applying reductions in elimination order ---
     queries = []
-    cur_aq = input_aq.copy()
+    cur_aq = _aq_with_stats(input_aq)
     for idx in elimination_order:
         if idx in cur_aq.get_reducible_idxs():
             q = cur_aq.reduce_idx(idx)
```
kylebd99: This is dead code.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -2,13 +2,111 @@
...
+from functools import reduce
 
-from ....finch_logic import Query
+from ....finch_logic import Field, Query
 from .annotated_query import AnnotatedQuery
 
+# Set env ``GALLEY_PRINT_BNB_DFS_MEMO=1`` to trace DFS memo (verbose).
+_PRINT_BNB_DFS_MEMO = os.environ.get("GALLEY_PRINT_BNB_DFS_MEMO", "") == "1"
+# All-fork sibling prefix blocking in ``branch_and_bound_dfs`` (module doc).
+_PREFIX_BLOCK_DFS = os.environ.get("GALLEY_BNB_DFS_PREFIX_BLOCK", "1") == "1"
```
kylebd99: This should be passed as a parameter rather than an env value.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
...
+        )
+
+    sorted_keys = sorted(memo.keys(), key=_vars_key_sort_key)
+    # --- optimal_subquery_costs and optimal_perm_by_vars only used for testing ---
+    optimal_subquery_costs = OrderedDict((vk, memo[vk][0]) for vk in sorted_keys)
+    optimal_perm_by_vars = OrderedDict((vk, memo[vk][1]) for vk in sorted_keys)
+    return (best_complete, optimal_subquery_costs, optimal_perm_by_vars)
+
+
 def pruned_query_to_plan(
```
kylebd99, line 209: There shouldn't be two versions of this function. All of the interior is identical except for the calls to branch_and_bound_dfs. Having this much identical code is an anti-pattern.

src/finchlite/autoschedule/galley_optimize.py
```diff
@@ -39,12 +42,19 @@ def optimize_query(
     use_components: bool = True,
     *,
     use_exact_branch_and_bound: bool = False,
```
kylebd99: Rather than having two different parameters here, there should be one string parameter that say, e.g. "optimizer='bfs'/'dfs'/'greedy'"

src/finchlite/autoschedule/galley_optimize.py
```diff
@@ -39,12 +42,19 @@ def optimize_query(
...
+    use_branch_and_bound_dfs: bool = False,
 ):
     """Rewrite a single logical Query via greedy or exact branch-and-bound reduction."""
     annotated_query = AnnotatedQuery(stats_factory, query, stats_bindings)
     use_greedy = not use_exact_branch_and_bound
-    new_queries, _ = pruned_query_to_plan(
-        annotated_query, use_components=use_components, use_greedy=use_greedy
+    # NOTE: this is for test only
+    prune = (
```
kylebd99: This should be an if/elif/elif/else branch instead of a variable set equal to a function. Code clarity is paramount.

src/finchlite/autoschedule/galley_optimize.py
```diff
@@ -56,6 +66,7 @@ def optimize_plan(
     use_components: bool = True,
     *,
     use_exact_branch_and_bound: bool = False,
```
kylebd99: Same here about a string variable for setting the optimizer.

src/finchlite/autoschedule/galley_optimize.py
```diff
@@ -101,10 +113,12 @@ def __init__(
         use_components: bool = True,
         *,
         use_exact_branch_and_bound: bool = True,
```
kylebd99: Same here about a string variable for setting the optimizer.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -2,13 +2,111 @@
...
+disabled). Optional **ordered-prefix dominance** (default on): for each
+adjacent pair ``(d,e)`` along a witness path ``T`` (every split
+``T = P + (d,) + (e,) + …``), if ``cost(T)`` beats the best known cost for the
+sibling prefix ``P + (e,)`` (which must have been visited), extensions of
+``P + (e,)`` are skipped. If a witness path is seen **before** ``P + (e,)`` is
+visited, its cost is stored and the same skip applies when that prefix is
+reached later (e.g. ``ABCDE`` before ``ABCE``). This is a heuristic unless a
+problem-specific proof applies; set env ``GALLEY_BNB_DFS_PREFIX_BLOCK=0`` to
+disable.
+
```
kylebd99: A later comment discusses how this pruning rule needs to be expanded a bit. However, I would also stress that it is not a heuristic in terms of the quality of the plan that is produced. It will still find the optimal plan and this pruning will never eliminate superior plans.

benchmarks/galley_benchmarks.py
```diff
@@ -440,6 +537,9 @@ def main() -> None:
     print("### Galley BnB vs greedy benchmarks ###", flush=True)
     main_bnb_benchmarks()
     print("", flush=True)
+    print("### Galley layered BnB vs DFS BnB benchmarks ###", flush=True)
+    main_bnb_dfs_benchmarks()
```
kylebd99: I don't understand what the point of this benchmark is. The example that was picked does not show any substantive difference between bfs and dfs. It would be great to have an example where one is dramatically better than the other in terms of optimization speed. (They should both always produce the plans with equivalent cost.)

kylebd99 (changes requested):

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
...
+    ``max_subquery_costs`` superset bound, incumbent pruning, and optional
+    penultimate-fork ordered-prefix blocking (env ``GALLEY_BNB_DFS_PREFIX_BLOCK``).
+
+    Prints ``memo`` after each stack ``pop`` and each ``memo`` assignment.
+    Returns ``(best_complete, optimal_subquery_costs, optimal_perm_by_vars)``.
+    """
+    if k != float("inf"):
+        res = branch_and_bound(input_aq, component, k, max_subquery_costs)
+        return (*res, None)
+
```
kylebd99: I would remove k, max_subquery_costs, and this call to branch_and_bound. This algorithm should perform that first "greedy" pass in the course of performing the dfs rather than as a pre-processing step.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
...
+        if vars_key == component_set:
+            if cost < best_complete_cost:
+                best_complete_cost = cost
+                best_complete = (order, queries, aq, cost)
+            continue
+
+        if cost >= best_complete_cost:
+            continue
+
```
kylebd99, line 163: This check should be unecessary because you will have already compared with the best plan in the memo.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
...
+        pop_idx += 1
+        if __debug__ and _PRINT_BNB_DFS_MEMO:
+            _print_dfs_memo_snapshot("pop", pop_idx, dict(memo))
+        # Stale frame: a cheaper path to this vars_key already updated memo[vars_key].
+        if cost > _dfs_memo_cost(memo, vars_key):
+            continue
+
+        if _PREFIX_BLOCK_DFS and _dfs_order_prefix_blocked(order, blocked_prefixes):
+            continue
+
```
kylebd99, line 154: You shouldn't need to explicitly keep a set of blocked prefixes. This pruning should just happen by checking the memo table.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
```diff
@@ -130,10 +229,136 @@ def branch_and_bound(
...
+            total_cost = cost + step_cost
+            new_vars = vars_key | frozenset(reduced_vars)
+
+            bound = float("inf")
+            for vars2 in max_subquery_costs:
+                if vars2 >= new_vars:
+                    bound = min(bound, max_subquery_costs[vars2])
+
+            if total_cost > bound:
+                continue
```
kylebd99, line 185: This will go away when you remove k.

kylebd99 (changes requested):

benchmarks/galley_benchmarks.py
```diff
@@ -430,6 +522,16 @@ def main_bnb_benchmarks() -> None:
...
 
 
+def main_bnb_bfs_vs_dfs_benchmarks() -> None:
+    """BFS exact BnB vs DFS exact BnB (same cost; compare optimize_plan_s)."""
+    for case in BNB_BFS_VS_DFS_CASES:
+        _run_bnb_bfs_vs_dfs_case(case)
+
+
+# Backward-compatible name (older docs / one-liners).
+main_bnb_layered_vs_dfs_benchmarks = main_bnb_bfs_vs_dfs_benchmarks
```
kylebd99: Why the renaming here?

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
kylebd99: This function is identically defined in both files. This kind of unecessary redefinition is a code anti-pattern. Also, rather than defining a new version of the get_cost_of_reduce_idx from AnnotatedQuery, in order to get the reduced_idxs, you should just have that function return the reduced_idxs as well. Currently, get_reduce_query is effectively being called twice, once explicitly and another time within aq.get_cost_of_reduce_idx.

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound.py
kylebd99: This still feels potentially true, I wonder if there's a way that we can dedup this...

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound_dfs.py
```diff
@@ -0,0 +1,153 @@
+"""
+Depth-first branch-and-bound for Galley reduction order. For a fixed component it
+searches the same elimination-order space as layered ``branch_and_bound`` with
+``k = float('inf')`` and returns the same optimal cost.
+
+Pruning uses the minimum cost among all **known** supersets of the child state's
+variable set: greedy ``k=1`` bounds, entries in the DFS ``memo``, and the best
```
kylebd99: stale comment

src/finchlite/autoschedule/galley_optimize.py
```diff
@@ -1,12 +1,20 @@
...
-Galley logical optimizer: applies greedy query rewriting to logical plans, with
-an optional exact branch-and-bound path for query bodies.
+Galley logical optimizer: rewrites logical query bodies via a selectable backend.
+
+``optimizer`` selects greedy reduction, exact layered (BFS-style) branch-and-bound,
+or exact DFS branch-and-bound. The layered exact path may run a greedy ``k=1`` pass
+per component to derive pruning bounds; the DFS path does not reuse those bounds
+inside ``branch_and_bound_dfs`` (see ``pruned_query_to_plan_dfs``). Greedy mode is
+heuristic; exact ``bfs`` / ``dfs`` modes search for an optimal-cost elimination order
+within the same cost model. DFS pruning is described (with optimality assumptions)
```
kylebd99: what optimality assumptions?

src/finchlite/autoschedule/galley/logical_optimizer/branch_and_bound_dfs.py
```diff
@@ -0,0 +1,153 @@
...
+                continue
+
+            memo[new_vars] = new_cost
+            new_aq = aq.copy()
+            reduce_query = new_aq.reduce_idx(idx)
+            new_order = list(order) + [idx]
+            new_queries = list(queries) + [reduce_query]
+            children.append((new_vars, new_order, new_queries, new_aq, new_cost))
+
+        children.sort(key=lambda x: x[4])
```
kylebd99: Nice! I don't know if we talked about this, but making sure that we follow the lowest cost path first is an important optimization!

benchmarks/galley_benchmarks.py
```diff
@@ -218,6 +220,23 @@ def make_bnb_slow_example() -> LazyTensor:
...
+    (28, 13),
+    (13, 6),
+    (6, 11),
+    (11, 52),
+    (52, 61),
+    (61, 7),
+]
+
+
+def make_bnb_core_dfs_wins_chain() -> LazyTensor:
```
kylebd99, line 232: Do we believe that dfs is always faster or can we come up with an example where bfs is better? If not, we should go ahead and make dfs the default.

kylebd99 (approved):
Just a few nits, but other than that LGTM!

kylebd99 (approved):

PR #363, opened 2026-04-06
Jx/enforce ftype

jxu44:
datatype work to enable removal of property registry


willow-ahrens:
close in favor of https://github.com/finch-tensor/finch-tensor-lite/pull/369

PR #367, opened 2026-04-07
additional methods for BufferizedNDArray

sreevickrant:
The following methods were added to BufferizedNDArray:
- reshape
- permute_dims
- expand_dims


willow-ahrens (changes requested):
Looks good! Can we add a finchlite.reshape toplevel method which calls the tensor's internal reshape definition too? Make it conform to https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.reshape.html

sreevickrant:
`finchlite.reshape(...)` is exposed as a top level method. Some of the error handling was written by Claude based on the API spec. Works for BuffNDArray only.

src/finchlite/interface/eager.py
```diff
@@ -265,6 +267,9 @@ def full(
 def permute_dims(arg, /, axis: tuple[int, ...]):
     if isinstance(arg, lazy.LazyTensor):
         return lazy.permute_dims(arg, axis=axis)
+    fmt = ftype(arg)
+    if hasattr(fmt, "permute_dims"):
+        return fmt.permute_dims(arg, axis)
```
willow-ahrens, line 272: Check if permute_dims is defined on the argument rather than ftype

src/finchlite/interface/eager.py
```diff
@@ -819,6 +829,82 @@ def flatten(x):
     return compute(lazy.flatten(x))
 
 
+def reshape(x, shape: tuple, /, *, copy=None):
```
willow-ahrens, line 832: If argument has a reshape method then call it otherwise notimplementedError and move this logic to bufferized

willow-ahrens (approved):

src/finchlite/interface/eager.py
willow-ahrens: If argument is lazy compute it first, then check for reshape method

willow-ahrens: if arg was lazy return a lazy

willow-ahrens (changes requested):

willow-ahrens:
For the permute_dims and expand_dims calls, I think those would get fixed if bufferized was an EagerTensor. Let's make a PR for EagerTensor first, then check whether we still need to add those.

sreevickrant:
closing in favor of #400

PR #368, opened 2026-04-07
Rs/loader cache

### LINKED ISSUE #266: Nearest Neighbor stats caching Galley

### LINKED ISSUE #267: Boring Galley Caching
- need to be able to clear specific cache
- need to be able to clear all caches
- need to be able to turn off caching
- use the config machinery in utils
- add docs
- Need to tie file caches to a timestamp on the directory
-

### LINKED ISSUE #350: Refactor to unify handling of unions and joins in stats
TensorStats should implement mapjoin in terms of mapjoin_join (which takes an intersection of stats) and mapjoin union (which takes a union of stats). The TensorStats class should dispatch arguments to these cases based on whether operator fills are annihilators or not, and return
mapjoin_join(*join_args, mapjoin_union(*union_args))

You'll need to split all the child stats mapjoin implementations into mapjoin_join and mapjoin_union

Ruchika-2003:


src/finchlite/autoschedule/cache.py
```diff
@@ -0,0 +1,77 @@
...
+        bindings: dict[Alias, TensorFType],
+        stats: dict[Alias, TensorStats],
+        stats_factory: StatsFactory,
+    ):
+        key = (prgm, tuple(bindings.items()))
+
+        if key not in self.cache:
+            self.cache[key] = self.ctx(prgm, bindings, stats, stats_factory)
+
+        return self.cache[key]
```
willow-ahrens: This needs to be merged into executor so that we don't compute stats if there's already a kernel.

willow-ahrens:
@Ruchika-2003, could you merge main and get it to pass pre-commit?

src/finchlite/autoschedule/tensor_stats/blocked_stats.py
```diff
@@ -18,17 +18,22 @@
 class BlockedStatsFactory(StatsFactory["BlockedStats"]):
     def __init__(
         self,
-        blocks_per_dim: Mapping[Field, int],
         stats_factory: StatsFactory[NumericStats],
+        block_count: int = 5,
+        block_width: int = 2,
```
willow-ahrens: The default block_width could be a little higher

src/finchlite/autoschedule/tensor_stats/blocked_stats.py
```diff
@@ -18,17 +18,22 @@
...
+        self.block_width = block_width
         self.inner_factory = stats_factory
 
     def __call__(self, tensor: Any, fields: tuple[Field, ...]) -> BlockedStats:
         return BlockedStats.from_tensor(
             tensor,
             fields,
-            blocks_per_dim=self.blocks_per_dim,
+            blocks_per_dim={
+                f: int(min(self.block_count, n / self.block_width))
```
willow-ahrens: Should be n // blockwidth. look up differences between / and //

https://discuss.python.org/t/integer-division/23140

src/finchlite/autoschedule/tensor_stats/blocked_stats.py
```diff
@@ -266,3 +271,13 @@ def from_tensor(
...
         return float(sum(b.estimate_non_fill_values() for b in self.blocks.flat))
+
+    def get_embedding(self) -> np.ndarray:
+        total_elements = math.prod(self.tensordef.dim_sizes.values())
+        num_blocks = self.blocks.size
+        block_volume = total_elements / num_blocks
+        densities = [
+            b.estimate_non_fill_values() / block_volume for b in self.blocks.flat
+        ]
+        density_array = np.array(densities)
```
willow-ahrens, line 271: I think we might want to put the dimensions into this embedding

src/finchlite/autoschedule/tensor_stats/database_stats.py
```diff
@@ -244,3 +244,5 @@ def from_def(
 
     def estimate_non_fill_values(self) -> float:
         return self.nnz
+
+    def get_embedding(self): ...
```
willow-ahrens: You can see in the above method `from_def` that database stats has two data fields, nnz and V. Just use those, along with the tensor size, to make an embedding.

src/finchlite/autoschedule/tensor_stats/uniform_stats.py
```diff
@@ -149,3 +149,13 @@ def _get_volume(d: TensorDef) -> float:
...
     def estimate_non_fill_values(self) -> float:
         return self.nnz
+
+    def get_embedding(self) -> np.ndarray:
+        sizes = [float(self.dim_sizes[field]) for field in self.index_order]
+        volume = self._get_volume(self.tensordef)
+
+        prob = self.nnz / volume if volume > 0 else 0.0
+
+        vector = sizes + [prob]
```
willow-ahrens: minor point: In general it's almost never a good idea to name a variable "vector" Is there a more descriptive name you could use?

src/finchlite/autoschedule/executor.py
```diff
@@ -53,13 +54,16 @@ def __init__(
...
+        cache: bool = False,
     ):
         if ctx is None:
             ctx = DefaultLogicFormatter()
         if stats_factory is None:
             stats_factory = DenseStatsFactory()
         self.ctx: LogicLoader = ctx
         self.stats_factory = stats_factory
+        self.cache = cache
+        self.stats_cache: dict[tuple[Any, Any], OrderedDict] = {}
```
willow-ahrens: So the point of putting the cache in the executor is that we wanted the executor cache not to use stats, and instead to cache first. Can you swap the logic in here with the CacheFirst logic from cache.py?

src/finchlite/autoschedule/cache.py
willow-ahrens: You can also delete the CacheFirst method since it should only be implemented in the Executor.

src/finchlite/autoschedule/cache.py
```diff
@@ -0,0 +1,257 @@
...
+        if key not in self.cache:
+            logger.debug("CacheFirst MISS, compiling a new kernel")
+            self.cache[key] = self.ctx(prgm, bindings, stats, stats_factory)
+        else:
+            logger.debug("CacheFirst HIT, reusing kernel")
+
+        return self.cache[key]
+
+
+class LogicCacheLRU(LogicLoader):
```
willow-ahrens: We can delete the LogicCacheLRU in favor of the Embeddings version. We can also delete is_similar for all stats in favor of embeddings.

src/finchlite/autoschedule/cache.py
```diff
@@ -0,0 +1,257 @@
...
+        kernels[new_stats_key] = result
+        kernels.move_to_end(new_stats_key)
+
+        if len(kernels) > self.max_depth:
+            kernels.popitem(last=False)
+
+        return result
+
+
+class LogicCacheLRU_Embeddings(LogicLoader):
```
willow-ahrens: What's the difference between LRU_Embeddings and LRU_Embeddings_Norms?

willow-ahrens (changes requested):
A little bit of reorganization requested, but this is the right idea. I'm curious about the difference between the embeddings and the embeddings_norms caches.

Ruchika-2003:
> A little bit of reorganization requested, but this is the right idea. I'm curious about the difference between the embeddings and the embeddings_norms caches.

Norms is build up on top of the previous one, I forgot to remove it. I'll get rid of it.

src/finchlite/autoschedule/tensor_stats/database_stats.py
```diff
@@ -244,3 +234,17 @@ def from_def(
...
+
+    def get_embedding(self): 
+        sizes = [float(self.dim_sizes[field]) for field in self.index_order]
+        v_vals = [float(self.V.get(field,0.0)) for field in self.index_order]
+        total_elements = math.prod((self.tensordef.dim_sizes.values()))
+        density = self.nnz / total_elements
+
+        size_part = np.log2(np.array(sizes))
+        v_vals_part = np.log2(np.array(v_vals))
+        density_part = np.log2(np.array([density])+1)
```
willow-ahrens: can use raw nnz

src/finchlite/autoschedule/tensor_stats/dc_stats.py
```diff
@@ -519,3 +509,31 @@ def estimate_non_fill_values(self) -> float:
             if node.issuperset(idx):
                 min_weight = min(min_weight, weight)
         return min_weight
+
+    def get_embedding(self) -> np.ndarray:
+        all_fields = frozenset(self.tensordef.index_order)
```
willow-ahrens: dc_embedding = [dc.value for dc in sorted(dcs, by=lambda dc: (dc.x, dc.y))]
embedding = log2(np.ndarray([*sizes, *dc_embedding]))

src/finchlite/autoschedule/cache.py
willow-ahrens: need to merge lines 35-43 into logicexecutor

willow-ahrens: and then delete logicstatsfirst

src/finchlite/autoschedule/cache.py
```diff
@@ -0,0 +1,145 @@
...
+
+        prgm_key = (prgm, tuple(bindings.items()), stats_factory)
+        if prgm_key not in self.cache:
+            self.cache[prgm_key] = {
+                "cached_embeddings": None,
+                "kernels": [],
+            }  # embeddings : result (kernel)
+
+        entry = self.cache[prgm_key]  # fetching the cached vectors and kernels
+
```
willow-ahrens, line 55: generalize the normalization by doing:

factor = vector_norm(np.ones(len(embedding)), norm_ord)
embedding /= factor

src/finchlite/autoschedule/executor.py
```diff
@@ -53,13 +55,19 @@ def __init__(
...
-            stats_factory = DenseStatsFactory()
-        self.ctx: LogicLoader = ctx
-        self.stats_factory = stats_factory
+        loader = ctx if ctx is not None else DefaultLogicFormatter()
+        self.stats_factory: StatsFactory = (
+            stats_factory if stats_factory is not None else DenseStatsFactory()
+        )
+
+        if cache :
+            final_loader : lgc.LogicLoader = LogicCacheFirst(loader)
```
willow-ahrens: You should inline logiccachefirst here

willow-ahrens (changes requested):
looks good!

willow-ahrens (approved):

PR #371, opened 2026-04-09
Looplet simplify pass

mtsokol:
Should be reviewed after #344 is merged.

This PR introduces LoopletSimplify - a transform pass which takes care of e.g. x * 0 => 0 cases before unfurling respective levels and their looplets.

`tests/reference/test_asm_sparse_elemwise.txt` shows a newly generated plan for `A .* B` where `A` and `B` are CSRs with 0 fill value. You can see that only one `while` loop is generated where steppers overlap.


willow-ahrens:
I want to add a few updates to make function-call-based simplification unified across the repo. Since we already have a unified notion of functions and function calls, we could make a CallTerm, CallValue, and CallLiteral abstract classes that allow us to rewrite any terms which inherit from them the same way.

willow-ahrens:
I'll merge 344 first, then maybe take a go at this here, if you're okay with that.

tests/test_sparse_level.py
```diff
@@ -61,3 +66,62 @@ def test_selected_ops(dtype):
...
+            super().__init__(level=level)
+            self.records = []
+
+        def emit(self, record):
+            self.records.append(record)
+
+    handler = DummyHandler(logging.DEBUG)
+    log = logging.getLogger("finchlite.compile.lower")
+    log.addHandler(handler)
+    log.propagate = False
```
mtsokol, line 129: Something to discuss: Testing infrastructure: How we want fetch/collect compiler's outputs from different levels.

tests/test_sparse_level.py
```diff
@@ -61,3 +66,62 @@ def test_selected_ops(dtype):
...
+        DenseLevel(
+            SparseListLevel(
+                ElementLevel(element(dtype(0), dtype, np.intp, NumpyBufferFType), data),
+                np.intp(3),
+                ptr,
+                idx,
+            ),
+            np.intp(3),
+        )
+    )
```
mtsokol, line 98: We need constructors to get rid of these...

mtsokol:
> I want to add a few updates to make function-call-based simplification unified across the repo. Since we already have a unified notion of functions and function calls, we could make a CallTerm, CallValue, and CallLiteral abstract classes that allow us to rewrite any terms which inherit from them the same way.

I support this idea: for collapsing calls with an annihilator argument I wrote these simplifications at least twice (notation and assembly).

src/finchlite/finch_notation/simplification.py
```diff
@@ -0,0 +1,27 @@
...
+    def simplify(cls, term: ntn.NotationNode):
+        from ..compile import looplets as lplt
+        from ..interface.scalar import Scalar
+
+        match term:
+            case ntn.Call(ntn.Literal(_) as op, args):
+                for arg in args:
+                    match arg:
+                        case ntn.Unwrap(
+                            ntn.Access(lplt.Run() as tns, ntn.Read(), idxs)
```
willow-ahrens, line 23: Can we replace the run with its value first? We should be able to separately simplify any unwrap(access(run))) statements to unwrap(access(fill)), then unwrap(access(fill)) to 0

willow-ahrens (changes requested):

src/finchlite/finch_notation/simplification.py
mtsokol: It makes sense to separate but I think we can address it once we iron out Scalars.
If I do `unwrap(access(fill))`, `fill` is a Scalar which isn't a dataclass: Rewriting breaks with missing proper `__eq__` for Scalars for Fixpoint for example. I can add `__eq__` here for Scalar class but I think we could save it for separate "Scalar redo" work to get them as we want in the compiler properly.
I can add a TODO for that task - to separate steps for rewriting though.

src/finchlite/autoschedule/standardize.py
```diff
@@ -475,7 +474,14 @@ def rule(ex):
                 arg_3 = tuple(Rewrite(PostWalk(rule))(arg) for arg in args)
                 return Query(lhs, Reorder(MapJoin(op, arg_3), idxs))
 
-    return Rewrite(PostWalk(rule_2))(root)
+    def rule_2(stmt):
```
mtsokol, line 477: For now, to get rid of identity reorders that add_aggregates introduce. Once we refactor Optimizer and get rid of standardize.py this should be gone/somewhere else.

src/finchlite/finch_notation/simplification.py
willow-ahrens: Fill is a little different than a scalar, Fill is n-dimensional, but scalar is zero-dimensional.

willow-ahrens (approved):

Issue #72, opened 2025-06-03
annihilator properties (and others) should take the type of other arguments in the call


willow-ahrens:
we cannot determine whether e.g. False is an annihilator, we don't know if it promotes or not.

sreevickrant:
https://numpy.org/devdocs/reference/arrays.promotion.html

PR #375, opened 2026-04-09
Jx/annihilators

jxu44:
addresses #72 and #73


jxu44:
shortcircuit_or class implemented. currently is_annihilator changes only exist for Or_. is_annihilator should be changed to examine the types of the arguments passed into or_ (support nary)? test cases need to be updated as well.

PR #382, opened 2026-04-14
Fixed dummy stats and blocked stats

### LINKED ISSUE #380: Dummy stats not working

### LINKED ISSUE #381: BlockedStats Reorder bug
Ran into this error: ValueError: axes don't match array

yusheng036:
Fixed DummyStats not actually working. Fixed BlockedStats.reorder where axes_mapping didn't include all axes from old_order, causing np.transpose to fail with ValueError: axes don't match array when dropping size-1 dimensions

Fixes #380 Fixes #381


kylebd99 (dismissed):
Hey, these fixes look good, but it would be great to have a test in test_tensor_stats.py that demonstrates the bug fix (i.e. fails w/out the fix and passes w/the fix)!

willow-ahrens:
@yusheng036, do you have a reproducer for #380 and #381?

willow-ahrens (changes requested):
I'm with kyle that the reproducer should be added to tests

yusheng036:
Hi @willow-ahrens @kylebd99, I have added test cases for BlockedStats.reorder and DummyStats. For BlockedStats, I made one dimension being dim_size =  1 and two dimension being dim_size = 1 and apply reordering to them. The test would be True if they drop those dimensions after reordering happens. I have also written some DummyStats test to check whether it works as intended. I wrote two DummyStats.mapjoin test that test if the axes are the same and if they are broadcasted.

willow-ahrens (approved):

PR #383, opened 2026-04-17
add notation cfg builder

idkwhttsay:


src/finchlite/finch_notation/cfg_builder.py
```diff
@@ -0,0 +1,304 @@
...
+
+@dataclass(eq=True, frozen=True)
+class NumberedStatement(NotationStatement):
+    """Wrapper for NotationStatement that assigns a unique id to each statement."""
+
+    stmt: NotationStatement
+    sid: int
+
+    def __str__(self) -> str:
+        return str(self.stmt)
```
willow-ahrens: does this not need to add the number?

willow-ahrens (approved):
Looks great! Up next: could you add a live variable analysis or maybe const-prop?

src/finchlite/finch_notation/cfg_builder.py
idkwhttsay: I print the ids in the __str__ method of dataflow abstract class

BUT, i changed it to print them in the __str__ method of the class itself

PR #385, opened 2026-04-20
Hash consing

ricky122-5:
Adds hash consing to finch-lite.

Implements hash consing across all 5 IR node hierarchies, so structurally equal nodes now return the same object via a WeakValueDictionary intern table. Intercepts construction in __new__ so both direct calls and rewriter paths are covered. Fixed a few subtle bugs along the way: Literal(0) and Literal(0.0) colliding due to Python's hash(0) == hash(0.0), a GC bug from storing into a weak dict without a local strong ref, and a traversal test that assumed unique node identities.

### DIFF
--- a/src/finchlite/symbolic/term.py
+++ b/src/finchlite/symbolic/term.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import inspect
+import weakref
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from inspect import isbuiltin, isclass, isfunction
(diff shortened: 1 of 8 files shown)


src/finchlite/symbolic/term.py
```diff
@@ -45,6 +47,40 @@ def recurse(node: Term) -> Term:
...
+            else:
+                norm = args
+            key = (cls,) + tuple(_key_part(a) for a in norm)
+            obj = hash_cons_table.get(key)
+            if obj is None:
+                obj = object.__new__(cls)
+                hash_cons_table[key] = obj
+            return obj
+        except TypeError:
+            return object.__new__(cls)
```
willow-ahrens, line 81: Define eq and hash as id

src/finchlite/symbolic/term.py
```diff
@@ -45,6 +47,40 @@ def recurse(node: Term) -> Term:
...
+
+def _key_part(a: object) -> object:
+    if isinstance(a, Term):
+        return id(a)
+    if isinstance(a, tuple):
+        return tuple(_key_part(x) for x in a)
+    return (type(a), a)
+
+
+class HashCons:
```
willow-ahrens, line 62: Add an abstractmethod called __Hash_keys__ which defines which keys are used in the hashcons object

src/finchlite/symbolic/term.py
```diff
@@ -45,6 +47,40 @@ def recurse(node: Term) -> Term:
 """
 
 
+hash_cons_table: weakref.WeakValueDictionary = weakref.WeakValueDictionary()
+_sig_cache: dict = {}
+
+
+def _key_part(a: object) -> object:
```
willow-ahrens, line 54: This method would be disaggregated across child classes of HashCons

willow-ahrens (changes requested):
Term should inherit from hashCons, and TermTree should define __hash_key__

sreevickrant:
closed in favour of #642

PR #386, opened 2026-04-22
autoscheduler: efficiently support in-place updates in finch-logic

JoelMathewC:
This PR improves support for in-place updates within the autoscheduler. Before this PR, expressions denoting in-place updates were lowered to inefficient code. Consider the example below, we unnecessarily introduce the temporary `T1` despite the possibility to perform the operation in-place.
```
# Input
A[i,j] = A[i,j] + B[i,j]

# Output (before PR)
T1[i,j] = A[i,j] + B[i,j]
A[i,j] = T1[i,j]

# Output (after PR)
A[i,j] += B[i,j]
```

The above lowering is a result of the limitations imposed by the `LogicStandardizer` in an attempt to reduce the finch-logic program into canonical form. This PR makes the following changes
1. Expands the canonical form to now support _Queries that perform a Reorder of a map-join between query lhs and either of the other canonical forms._
    - Form 1: `Query(lhs, Reorder(MapJoin(op, Reorder(Table(lhs_1, idxs_1)), Reorder(Table(...))), idxs_2))` where `lhs == lhs_1` and `idxs_1 == idxs_2`
    - Form 2: `Query(lhs, Reorder(MapJoin(op, Reorder(Table(lhs_1, idxs_1)), Aggregate(op_1, ...))), idxs_2))` where `lhs == lhs_1` and `idxs_1 == idxs_2` and `op_1 in (op, ffuncs.overwrite)` _(If aggregation happens over the same operator as mapjoin or if the aggregate does not reduce over a dimension, i.e. ffunc.overwrite, then they are valid candidates for inplace)_
3. Updates several standardizer functions to recognize the new canonical form
4. Update the notation compiler to support lowering of the new canonical form to finch notation
5. Tests to verify the functionality


src/finchlite/autoschedule/standardize.py
```diff
@@ -65,13 +67,27 @@ def rule_1(ex):
...
+                for arg in node.children:
+                    yield from AggregateGuardedPostOrderDFS(
+                        arg, is_agg_seen or isinstance(arg, Aggregate)
+                    )
+
+            if is_agg_seen:
+                yield node
+
         match stmt:
             case Query(lhs, rhs):
```
willow-ahrens, line 189: Maybe instead of adding a new dfs, let's just add a separate match case for the query(lhs, mapjoin(op, (lhs, args...)) case.

src/finchlite/autoschedule/utils.py
```diff
@@ -0,0 +1,42 @@
...
+    Aggregate,
+    Alias,
+    Literal,
+    LogicExpression,
+    Reorder,
+    Table,
+)
+
+
+def is_inplace_expr(
```
willow-ahrens: Could we do this as a program transformation/rewrite, then we can just use a simple match pattern to detect inplace?

src/finchlite/autoschedule/standardize.py
```diff
@@ -84,6 +100,12 @@ def standardize_query_roots(root: LogicStatement, bindings) -> LogicStatement:
     )
 
     def rule(ex):
+        def wrap_tables_in_reorder(args):
+            return tuple(
+                Reorder(arg, arg.fields()) if isinstance(arg, Table) else arg
+                for arg in args
+            )
```
willow-ahrens: this makes me suspicious, I'm not sure precisely why it's needed.

src/finchlite/autoschedule/compiler.py
```diff
@@ -259,6 +280,45 @@ def __call__(self, prgm: lgc.LogicStatement) -> ntn.NotationStatement:
...
+                        ):
+                            body = self._lower_query_of_aggregate(
+                                lhs, op_1, arg_2, idxs_2
+                            )
+
+                if body is None:
+                    raise Exception(f"Unrecognized logic: {prgm}")
+
+                return ntn.Block(
+                    (
```
willow-ahrens, line 323: add thaw here

willow-ahrens (changes requested):
Great work, almost there!

src/finchlite/autoschedule/utils.py
willow-ahrens: Would it be possible to modify the standardizer to normalize in-place exprs to the form Query(lhs, Mapjoin(op, lhs, rhs))?

willow-ahrens (changes requested):
I'd like the standardizer to normalize in-place queries to the form `query(lhs, mapjoin(op, (lhs, rhs))`. Is this possible? Then we can check for in-place kernels with a match case `case query(lhs, mapjoin(op, (lhs_2, rhs)) if lhs == lhs2:`? Then we could avoid the use of a separate is_inplace_query helper, and keep the code in one place.

src/finchlite/autoschedule/utils.py
kylebd99: I don't believe that all expressions can be normalized to Query(lhs, Mapjoin(op, lhs, rhs)). For example, if we have Query(lhs, MapJoin(*, MapJoin(sigmoid, lhs), rhs)). Alternatively, we may have a non-commutative op.

kylebd99: Ah, wait, I think I see why my comment was dumb. The inplace update optimization itself requires the outermost mapjoin to have this form.

kylebd99: I do have one case that seems like it might be a bit tricky here, but may actually come up. In a lot of cases, we may want to only update some set of entries to a matrix based on a mask. This could potentially be expressed via something like the ternary ifelse(cond, A, lhs) operator which should still map to an inplace update where the aggregation function is the 'overwrite' function from the julia repo.

This may be out of scope for the current change or maybe it's covered here in a way that I'm not seeing.

kylebd99 (commented):
Overall, this looks great! Only two major comments:
1. It's unclear how to handle the case of inplace updates on some masked subset of a tensor. This is hard to express in the array API in the first place, so maybe we don't need to worry about this.
2. It would be great to have a benchmark file that demonstrates the speedup with a minimal extreme example.

Leaving the approval to @willow-ahrens

willow-ahrens (approved):
I'm happy with these changes now, just needs to pass tests!

src/finchlite/autoschedule/utils.py
JoelMathewC: Thanks for the review Kyle. The mask case is actually quite important to us since we wanted to use inplace updates for active sets. Willow and I have talked a bit about this, we can represent the mask case using the following logic program (however, this isnt generic enough since it wont work for all operators/fill values). I'll have to think about this a bit more as a part of my project. I'll try to have something up for that when I figure it out.

```
Query(lhs, MapJoin(+, Table(lhs), MapJoin(*, Table(mask), ... )))
```

Update: this will only work for semirings where + and * get replaced by corresponding semiring operators.

JoelMathewC:
@willow-ahrens this is ready for a merge now.

PR #391, opened 2026-04-28
LoopOrderingOptimizer

HoangNguyen0309:


src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,134 @@
...
+        Return the desired loop order.
+        """
+        ...
+
+    def __call__(
+        self,
+        prgm: LogicStatement,
+        bindings: dict[Alias, TensorFType],
+    ):
+        def reorder(node: LogicStatement) -> LogicStatement:
```
kylebd99: let's rename reorder

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,134 @@
...
+            match node:
+                case Plan(bodies):
+                    return Plan(tuple(reorder(body) for body in bodies))
+                case Query(lhs, rhs):
+                    loop_order = self.get_loop_order(node, bindings)
+                    return Query(lhs, Reorder(rhs, loop_order))
```
kylebd99: The correct place to put the loop order reorder is within the Aggregate if it exists. For example, the following would be a proper placement of reorder within the aggregate.
Query(Alias("C"), Aggregate(+, 0, Reorder(MapJoin(*, Table(Alias("A"), i, j), Table(Alias("B"), j,k)), (i,j,k)), (j,)))

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,134 @@
...
+                            clock += 1
+                        occurrences[idx] += 1
+                case Aggregate(_, _, arg, _):
+                    visit(arg)
+                case Reorder(arg, _):
+                    visit(arg)
+                case _:
+                    # MapJoin and other composites expose their children via
+                    # standard dataclass fields; walk them generically.
+                    for child in getattr(ex, "__dataclass_fields__", {}).keys():
```
kylebd99: use proper children interface instead of attr mashing

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,134 @@
...
+        occurrences: Counter[Field] = Counter()
+        first_seen: dict[Field, int] = {}
+        clock = 0
+
+        def visit(ex) -> None:
+            nonlocal clock
+            match ex:
+                case Table(_, idxs):
+                    for idx in idxs:
+                        if idx not in first_seen:
```
kylebd99: drop first_seen and clock

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,134 @@
...
+                                    visit(item)
+                        elif hasattr(val, "__dataclass_fields__"):
+                            visit(val)
+
+        match node:
+            case Query(_, rhs):
+                visit(rhs)
+
+        # Sort: highest occurrence first; break ties by first-seen order.
+        ordered = sorted(
```
kylebd99: add rev=True instead of negating the occurences

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+def _contains_aggregate_or_mapjoin(ex: LogicNode) -> bool:
+    """True if ``ex`` contains an ``Aggregate`` or ``MapJoin``."""
+    for node in PostOrderDFS(ex):
+        match node:
+            case Aggregate() | MapJoin():
+                return True
+    return False
+
+
+def _validate_query(query: Query, *, kind: str) -> None:
```
kylebd99: It doesn't look like "kind" gets used here. I think this is just validating the output grammar and will fail for valid inputs.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+        stats_factory: StatsFactory | None = None,
+    ):
+        validate(prgm, kind="input")
+        # NOTE: change when we have a proper stats factory
+        if stats is None:
+            stats = {}
+        if stats_factory is None:
+            stats_factory = DenseStatsFactory()
+        # End  NOTE
+
```
kylebd99: what do we mean by "proper stats factory" here?

src/finchlite/autoschedule/loop_ordering.py
kylebd99: I think that we should just have two separate functions 1) validate_input and 2) validate_output. Each of these functions should have a single walk rewrite that fully defines the grammar.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+
+# Validation func
+def _contains_aggregate_or_mapjoin(ex: LogicNode) -> bool:
+    """True if ``ex`` contains an ``Aggregate`` or ``MapJoin``."""
+    for node in PostOrderDFS(ex):
+        match node:
+            case Aggregate() | MapJoin():
+                return True
+    return False
```
kylebd99: This isn't used in the validation, it's used later on in the reorder func.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+        match node:
+            case MapJoin() as mj:
+                aligned = _transpose_tables(mj, loop_order)
+                return aligned if aligned is not mj else None
+            case _:
+                return None
+
+    return Rewrite(PostWalk(rule))(ex)
+
+
```
kylebd99, line 82: In this code, we're still using mapjoins in order to identify tables that are mis-ordered. This is messy. Instead, we'd like to do a single rewrite pass which matches on tables, checks whether that table is ordered correctly. If it's not, we then produces a new query that reorders it and rewrites the table to reference that alias.

src/finchlite/autoschedule/loop_ordering.py
kylebd99: still valid

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+    @abstractmethod
+    def get_loop_order(
+        self,
+        node: Query,
+        bindings: dict[Alias, TensorFType],
+    ) -> tuple[Field, ...]:
+        """
+        Return the desired loop order.
+        """
+        ...
```
kylebd99, line 111: This function should pass the stats dictionary as well because the galley version will require stats to do performance estimation.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+                            return Query(lhs, Reorder(rhs, loop_order))
+                case Produces(_):
+                    return node
+                case _:
+                    raise ValueError(
+                        f"Unsupported logic statement for loop ordering: {node}"
+                    )
+
+        prgm = reorder(prgm)
+        # mutate bindings_out
```
kylebd99: This comment is not self-explanatory.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+        bindings_out = dict(bindings)
+        prgm = concordize(prgm, bindings_out)
+        validate(prgm, kind="output")
```
kylebd99: Why are we validating twice?

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+        prgm = drop_reorders(prgm)
+        prgm = flatten_plans(prgm)
```
kylebd99, line 222: Is it possible to have nested plans at this stage?

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+                case Produces(_):
+                    return node
+                case _:
+                    raise ValueError(
+                        f"Unsupported logic statement for loop ordering: {node}"
+                    )
+
+        prgm = reorder(prgm)
+        # mutate bindings_out
+        bindings_out = dict(bindings)
```
kylebd99: Why are we making a new bindings dictionary? We shouldn't be changing the bindings dictionary anyways.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+            return (t, t.idxs)
+        case Reorder(Table(_, _) as t, idxs):
+            return (t, idxs)
+        case _:
+            return None
+
+
+def _transpose(t: Table, reordered: tuple[Field, ...]) -> Reorder:
+    """Logical view ``reordered`` of the same  tensor."""
+    return Reorder(Table(t.tns, t.idxs), reordered)
```
kylebd99: This could just be Reorder(t, reordered). Probably doesn't need to be a function anyways.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def _transpose_tables(mj: MapJoin, loop_order: tuple[Field, ...]) -> MapJoin:
+    """Reorder each all-table ``MapJoin`` arg to match
+    ``loop_order`` on shared indices."""
+    views = tuple(_get_operand_table_and_idxs(a) for a in mj.args)
+    if any(v is None for v in views):
+        return mj
```
kylebd99: Would you only reorder if all mapjoin args are tables? This doesn't make sense to me.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,336 @@
...
+                            return Query(
+                                lhs,
+                                Aggregate(
+                                    op,
+                                    init,
+                                    Reorder(inner, loop_order),
+                                    reduce_axes,
+                                ),
+                            )
+                        case Reorder(inner, _old) if not _contains_aggregate_or_mapjoin(
```
kylebd99: Remove this function and just match on Reorder(Table(), _old)

kylebd99 (changes requested):
Changes requested in comments

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+            )
+
+    rhs = query.rhs
+    n = sum(1 for node in PostOrderDFS(rhs) if isinstance(node, Aggregate))
+    if n > 1:
+        raise ValueError("Invalid loop ordering: at most one Aggregate per Query rhs")
+    walk(rhs, False)
+
+
+def _validate_output_query(query: Query, *, kind: str) -> None:
```
kylebd99: remove kind param

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+    new_ex = Rewrite(PostWalk(rule))(ex)
+    queries = tuple(
+        Query(alias, Reorder(Table(var, physical), desired))
+        for (var, physical, desired), alias in needed_swizzles.items()
+    )
+    return new_ex, queries
+
+
+# Validation func
+def _validate_input_query(query: Query, *, kind: str) -> None:
```
kylebd99: remove kind parameter

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+    n = sum(1 for node in PostOrderDFS(rhs) if isinstance(node, Aggregate))
+    if n > 1:
+        raise ValueError("Invalid loop ordering: at most one Aggregate per Query rhs")
+    walk(rhs, False)
+
+
+def _validate_output_query(query: Query, *, kind: str) -> None:
+    prefix = f"Invalid loop ordering {kind}:"
+
+    def walk(ex: LogicNode, inside_aggregate: bool) -> None:
```
kylebd99: remove inside_aggregate parameter and just set inside_agg to false

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+                case _:
+                    return None
+            return None
+
+        Rewrite(PreWalk(rule))(ex)
+
+    match query:
+        case Query(_, Aggregate(_, _, Reorder(_, _), _)):
+            pass
+        case Query(_, Reorder(inner, _)) if not isinstance(inner, Aggregate):
```
kylebd99: inner should be a table not just not an aggregate

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+                        )
+                case _:
+                    return None
+            return None
+
+        Rewrite(PreWalk(rule))(ex)
+
+    match query:
+        case Query(_, Aggregate(_, _, Reorder(_, _), _)):
+            pass
```
kylebd99: In this case, we should check whether every table agrees with the loop order. Can also be part of the overall walk rule.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+        case _:
+            raise ValueError(
+                f"{prefix} Query rhs must be "
+                "Reorder(...) or Aggregate(..., Reorder(...), ...)"
+            )
+
+    rhs = query.rhs
+    n = sum(1 for node in PostOrderDFS(rhs) if isinstance(node, Aggregate))
+    if n > 1:
+        raise ValueError("Invalid loop ordering: at most one Aggregate per Query rhs")
```
kylebd99: can check this using in_agg during the walk

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+        super().__init__()
+        if loader is None:
+            loader = MockLogicLoader()
+        self.loader = loader
+
+    @abstractmethod
+    def get_loop_order(
+        self,
+        node: Query,
+        bindings: dict[Alias, TensorFType],
```
kylebd99, line 104: should also pass stats and StatsFactory

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+        match node:
+            case Query(_, rhs):
+                visit(rhs)
+
+        ordered = sorted(
+            occurrences.keys(),
+            key=lambda idx: occurrences[idx],
+            reverse=True,
+        )
+        return tuple(ordered)
```
kylebd99: Take the loop ordering logic from standardize.py

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,328 @@
...
+        raise ValueError("Invalid loop ordering: at most one Aggregate per Query rhs")
+    walk(rhs, False)
+
+
+def validate(
+    prgm: LogicStatement,
+    *,
+    kind: Literal["input", "output"] = "input",
+) -> None:
+    """Reject programs outside the loop-ordering grammar."""
```
willow-ahrens: Could we write the grammar to validate in ebnf? in the docstring?

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,414 @@
...
+        validate(prgm, kind="input")
+        namespace = Namespace(prgm)
+
+        def apply_loop_order(node: LogicStatement) -> LogicStatement:
+            match node:
+                case Plan(bodies):
+                    return Plan(tuple(apply_loop_order(body) for body in bodies))
+                # if table and connected to bindings dont touch.
+                # added here to catch earlier
+                case Query(_, Reorder(Table(_, _), _)) if bindings:
```
kylebd99: should remove

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,414 @@
...
+) -> tuple[LogicExpression, tuple[Query, ...]]:
+    """Align each ``Table`` / ``Reorder(Table, ...)`` to ``loop_order``."""
+    needed_swizzles: dict[
+        tuple[Alias, tuple[Field, ...], tuple[Field, ...]], Alias
+    ] = {}
+    # vecdot test
+    # swizzles alias conflict
+    if namespace is None:
+        namespace = Namespace(ex)
+
```
kylebd99: remove this

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,414 @@
...
+        field_set = set(logical)
+        desired = tuple(f for f in loop_order if f in field_set)
+        if desired == logical:
+            return None
+        # ValueError: zip() argument 2 is longer/shorter than argument 1
+        # shape mismatch
+        # dont swizzle if not permuation of idxs
+        # and if bindings to preserve shape
+        if bindings and (
```
kylebd99: bindings should always exist

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,414 @@
...
+        bindings: dict[Alias, TensorFType],
+        stats: dict[Alias, TensorStats],
+        stats_factory: StatsFactory,
+    ) -> tuple[Field, ...]:
+        """
+        Return the desired loop order.
+        """
+        ...
+
+    def __call__(
```
kylebd99: should be lower()

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -0,0 +1,363 @@
...
+    A LoopOrderer determines the loop ordering for each query in a logic
+    program. Subclasses implement ``get_loop_order`` to swap in different
+    loop-ordering strategies.
+
+    The input program is expected to be in the standardized form produced by
+    ``LogicStandardizer``: every query is either a Reorder of a single
+    argument, or an Aggregate over a Reorder of a series of map-joins.
+    """
+
+    def __init__(self, loader: LogicLoader | None = None):
```
willow-ahrens, line 94: we should constrain the child pass type to be of whatever form is output by the orderer.

willow-ahrens (changes requested):
Could we put the code back into the original files, just so that we can see a simpler diff as we review? Then we can do code reorganization as a separate copy-paste only pr.

src/finchlite/autoschedule/loop_ordering.py
willow-ahrens: ignore sorry

willow-ahrens:
I think this is stale now right?

willow-ahrens:
Feel free to reopen if this is live, but I'm pretty sure this PR is stale after the standardizer cleanup from kyle

PR #394, opened 2026-04-30
BufferizedNDArray inherits from EagerTensor

### LINKED ISSUE #351: BufferizedNDArray Doesn't Support Matmul
Minimal Error Example:

```
import finchlite
import numpy as np

A = np.ones((2,2))
A = finchlite.asarray(A)
A@A
```

Unless I've missed something, it seems like this should pass. For example, the alternative version does not error:

```
import finchlite
import numpy as np

A = np.ones((2,2))
A = finchlite.asarray(A)
finchlite.matmul(A,A)
```

sreevickrant:
#351


willow-ahrens (changes requested):
Can bufferizedndarray get moved to be imported after interface? I think it should be imported around the same time we import FIberTensor

sreevickrant:
I was going to ask. I didn't know where to move it so I resorted to update in place.

willow-ahrens:
Why is count nonzeroes in this pr?

sreevickrant:
`tests/test_notation_interpreter.py::test_count_nonfill_vector` needs count_non_zeros in EagerTensor. I can revert it back.

willow-ahrens:
does it? It looks like it's just asking for np.count_nonzeros. The count_nonfill op is implemented in FinchNotation. I think all that's needed there is to convert `a` to a ndarray just before we call `np.count_nonzeros`, is that right?

sreevickrant:
> Can bufferizedndarray get moved to be imported after interface? I think it should be imported around the same time we import FIberTensor

I think the problem here is I need to fix `dc_stats.py` to get rid of the circular import because it uses BufferizedNDArray.

willow-ahrens:
We're trying to avoid dynamic imports in the library, but I'm not sure what to do about this one.

sreevickrant:
I tried of bunch of stuff but I think this was the cleanest fix. We can always force the linter to not sort init and we can avoid a bunch of the dynamic imports.

willow-ahrens:
@kylebd99 @mtsokol do you see a way to avoid this dynamic import? The summary is that DCStats uses bufferizedNDArray to accumulate results in finchnotation, but DCStats is defined in autoscheduling module, after the definition of interface. I see two easy fixes:
1. Embrace chaos and allow relative imports (strongly against but can be convinced)
2. Define a set of "advanced" autoschedulers after we define interface functions, so that they can reuse the interface functions. This is just moving complex statistics stuff to a different folder that gets loaded after interface gets loaded.

willow-ahrens:
Is option 2 above a possibility?

sreevickrant:
I agree with option 2. It would also fix the other dynamic import that I had to make: 

autoschedule/compiler.py — from ..compile.lower import make_extent inside _lower_query_of_reorder and _lower_query_of_aggregate.

src/finchlite/interface/lazy.py
```diff
@@ -463,6 +461,8 @@ def asarray(arg: Any, format: TensorFType | None = None) -> Any:
         from finchlite.interface.scalar import Scalar
 
         if isinstance(arg, np.ndarray):
+            from ..compile import BufferizedNDArray
```
willow-ahrens, line 464: it looks like there's still dynamic imports here.

willow-ahrens (changes requested):
I think the agreed order of imports is that we want Bufferized to inherit from eager, and to load it after interface, along with other tensors.

so it should be:
compile
autoscheduling (including simple stats)
interface 
tensors (including bufferized)
and then finally advanced stats (stats which use finch kernels)

sreevickrant:
> I think the agreed order of imports is that we want Bufferized to inherit from eager, and to load it after interface, along with other tensors.
> 
> so it should be: compile autoscheduling (including simple stats) interface tensors (including bufferized) and then finally advanced stats (stats which use finch kernels)

Can I clarify what you mean by "simple stats" and "advanced stats"?

willow-ahrens:
Sure! The distinction is just dependency-based. "Simple stats" are stats that only depend on numpy and perhaps a "count_non_fill" method. I'll try my hand today at refactoring.

willow-ahrens:
closing in favor of #399. I ended up allowing the eagertensor to load dynamically from finch-tensor-lite, which I think is the right place to put the dynamic load.

PR #399, opened 2026-05-15
Fix BufferizedNDArray issues

### LINKED ISSUE #351: BufferizedNDArray Doesn't Support Matmul

willow-ahrens:
fixes https://github.com/finch-tensor/finch-tensor-lite/issues/351, also adds absolute import requirement


willow-ahrens:
I think benchmark failures are okay here. I changed the interface and the old version is failing. I'm not used to working with ASV, so I'm not sure what to do with failing CI

kylebd99:
It looks like it's an import error within simple_benchmarks.py:
```
[9](https://github.com/finch-tensor/finch-tensor-lite/actions/runs/25938466033/job/76249766128?pr=399#step:7:630)
                 File "/home/runner/work/finch-tensor-lite/finch-tensor-lite/benchmarks/simple_benchmarks.py", line 11, in <module>
                   from finchlite.autoschedule import COMPILE_NUMBA, INTERPRET_NOTATION_GALLEY
               ImportError: cannot import name 'COMPILE_NUMBA' from 'finchlite.autoschedule' (/home/runner/work/finch-tensor-lite/finch-tensor-lite/.asv/env/606287e3db2922a1ea431fdbf3655ceb/lib/python3.12/site-packages/finchlite/autoschedule/__init__.py)
               asv: benchmark failed (exit status 1)
```

kylebd99:
Hmm, I think it's because it's trying to run that benchmark on the old code version. Yeah, we should just allow merge when asv fails I think. In general, we may want to make the schedulers imported from the top level to avoid this when we move them around internally.

Edit: I think this is what you were saying above, but.... apparently I felt the need to restate it.

sreevickrant (approved):

PR #400, opened 2026-05-15
Add array-api-tests

### LINKED ISSUE #359: add ArrayAPI tests
We're not passing them right now I think.

### LINKED ISSUE #432: reshape

sreevickrant:
fixes #359 fixes #432


src/finchlite/interface/lazy.py
```diff
@@ -362,6 +387,54 @@ def full(
     return broadcast_to(val, shape)
 
 
+def full_like(x, /, fill_value, *, dtype=None):
```
willow-ahrens, line 406: Is this broadcasting? Double check plz. thx!

src/finchlite/interface/lazy.py
```diff
@@ -1880,6 +1953,16 @@ def not_equal(x1, x2) -> LazyTensor:
     return elementwise(ffuncs.not_equal, lazy(x1), lazy(x2))
 
 
+def where(condition, x1, x2) -> LazyTensor:
+    condition = _compute(lazy(condition))
+    x1 = _compute(lazy(x1))
+    x2 = _compute(lazy(x2))
+    result = BufferizedNDArray.from_numpy(
+        np.where(np.asarray(condition), np.asarray(x1), np.asarray(x2))
```
willow-ahrens: Don't use np.where. use an ifelse from algebra.ffuncs. Make ifelse if it doesn't exist yet.

willow-ahrens: this should be elementwise

willow-ahrens (changes requested):
some initial tweaks from the readthrough

src/finchlite/algebra/ftypes.py
```diff
@@ -803,6 +804,8 @@ def ftype(x) -> FType:
         return x.ftype
     if isinstance(x, np.dtype):
         x = x.type
+    if isinstance(x, np.ndarray):
+        return ftype(x.dtype)
```
willow-ahrens: I'm not sure about this one. If we pass an array to ftype, it should probably error

src/finchlite/interface/lazy.py
```diff
@@ -356,13 +397,61 @@ def full(
 
     - out (array): an array where every element is equal to fill_value.
     """
-    val = lazy(np.full((), fill_value, dtype=dtype))
+    val = lazy(np.full((), fill_value, dtype=_np_dtype(dtype)))
```
willow-ahrens, line 400: Can we broadcast a scalar here instead?

src/finchlite/tensor/bufferized_ndarray.py
```diff
@@ -47,7 +47,7 @@ def to_numpy(self):
     def from_numpy(cls, arr: np.ndarray) -> "BufferizedNDArray":
         itemsize = arr.dtype.itemsize
         strides = tuple(np.intp(stride // itemsize) for stride in arr.strides)
-        shape = tuple(np.intp(s) for s in arr.shape)
+        shape = tuple(int(s) for s in arr.shape)
```
willow-ahrens: This may or may not be safe depending on the shape_type of bufferizedNDArray

src/finchlite/tensor/bufferized_ndarray.py
```diff
@@ -67,7 +67,7 @@ def ftype(self):
 
     @property
     def shape(self):
-        return self._shape
+        return tuple(int(s) for s in self._shape)
```
willow-ahrens, line 70: why do we want shape_type to by python int?

willow-ahrens (changes requested):
I'm debugging the compiler now, but I thought I'd give my thoughts on the PR content first. Only a few small changes, so once we debug and address comments we can merge.

willow-ahrens:
@sreevickrant Okay, I fixed the codegen error. Go ahead and make the requested changes. Once this is passing mypy, etc. we can merge. Be sure to add all the array api failing tests to the skips before we merge. Also, let's make sure reshape is actually working.

src/finchlite/interface/lazy.py
sreevickrant: https://data-apis.org/array-api/latest/API_specification/generated/array_api.full_like.html#full-like

Numpy and JAX broadcast if the fill_value is an array or if the fill value shape does not match.

src/finchlite/interface/lazy.py
sreevickrant: numpy does not broadcast a scalar but that should work

src/finchlite/tensor/bufferized_ndarray.py
sreevickrant: array_api threw an error for np.intp and required a tuple of ints for shape, i can revert back and check again

sreevickrant: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.shape.html#shape

src/finchlite/algebra/ftypes.py
sreevickrant: I'll move this over to asarray.

willow-ahrens (approved):
thank you!

lucascolley:
IMO it is not very nice that these are lumped in with the main test suite, as any failures are then contained to a subprocess in a single test. I think it would be preferable to just have a Pixi task to be invoked separately which can match e.g. what NumPy is doing in CI: https://github.com/numpy/numpy/blob/c09ba3a0331e6f93e83ea2a1a7fc984c3d70781e/.github/workflows/linux.yml#L294-L308

PR #405, opened 2026-05-22
Formalize Stages & Add Validation

### LINKED ISSUE #272: Formal validators on passes

kylebd99:


src/finchlite/symbolic/stage.py
```diff
@@ -2,5 +2,37 @@
...
+
     @abstractmethod
-    def __call__(self, *inputs): ...
+    def transform(self, *inputs) -> tuple: ...
+
+    @abstractmethod
+    def validate_outputs(self, *outputs): ...
+
+    def lower(self, *outputs):
+        return self.ctx(*outputs)
```
willow-ahrens: If we're gonna rely on the child having a ctx, I think we should make the ctx an abstract property. On the other hand, I think we might be able to get away with only ever validating inputs and return values, so we only need a "lower" function and we don't need to formalize the role of "ctx". Consider that some stages actually have two child contexts.

src/finchlite/autoschedule/einsum.py
```diff
@@ -77,14 +78,25 @@ def generate_einsum_expr(
...
+
+    def transform(
         self, prgm: LogicStatement, bindings: dict[lgc.Alias, TensorFType]
     ) -> tuple[ein.EinsumStatement, dict[ein.Alias, TensorFType]]:
         bindings_2 = {ein.Alias(var.name): val for var, val in bindings.items()}
         return (generate_einsum_stmt(prgm), bindings_2)
 
 
-class LogicEinsumLoader(LogicLoader):
+class LogicEinsumLoader(NoTransformStage, LogicLoader):
```
willow-ahrens: Whatever validator we're using, it feels like if we can't validate a logic to einsum lowering, we have sortof missed the point of validation. Shouldn't this be a regular transformation, where the output is validated einsum of some form?

src/finchlite/symbolic/stage.py
```diff
@@ -2,5 +2,37 @@
...
+
+    def __call__(self, *inputs):
+        self.validate_inputs(*inputs)
+        outputs = self.transform(*inputs)
+        self.validate_outputs(*outputs)
+        return self.lower(*outputs)
+
+
+class NoTransformStage:
```
willow-ahrens: I'm somewhat convinced this class shouldn't exist, or rather I'm convinced that every stage should be a NoTransform stage.

willow-ahrens (changes requested):
I have a suggestion here: We should not try to validate a program before it is passed to another stage. Instead, we should let the child stage validate it's own input. This would remove the need for a notransform stage, and it would allow for e.g. stages with more than one child lowerer (which do exist, like the einsum one). Another benefit is that we could define different abstract classes for each kind of input normal form, and then build stages where the child context must inherit from that stage type. 

For example, imagine we have a "SingleAggregate" normal form, where each query has at most one aggregate at the top level. Then we could define a "SingleAggregateStage" abstract class which enforces inputs are in single aggregate form, and all of the loop ordering stages can inherit from `SingleAggregateStage`. This would then allow us to write the `AggregateIsolator` with a child context that inherits from `SingleAggregateStage`:
```
class AggregateIsolator(AnyLogicStage):
   def __init__(self, ctx:SingleAggregateStage):
      self.ctx:SingleAggregateStage = ctx

   def lower(prgm):
      #we know here that prgm is in "AnyLogic" form
      #we do some stuff to prgm to make prgm2
      #we believe that prgm2 is in "SingleAggregate" form, so we should be able to lower it with any SingleAggregateStage in the next step
      return self.ctx(prgm2)
```

Thus, the contract of a stage is that any "LogicLowerer" should be able to lower a logic program completely, as long as the input is in some form, and any child stages verify their inputs recursively.

kylebd99:
@willow-ahrens This PR is fairly close to being finished, but I don't want to hold you up if you're busy. The current version is passing (on my system at least, currently regening regression tests), and it incorporates the proper verification at the DefaultLogicOptimizer, LogicCompiler and LogicFormatter level. It does not verify anything at LogicStandardizer (denoted by inheritance from UnverifiedForm)  because that class technically still handles arbitrary finchlogic. 

I have also not yet unified DefaultLogicOptimizer and GalleyLogicOptimizer under a single shared class. My hope is that this will be easy now... but I'll have to see about doing it tomorrow.

kylebd99:
Oh, it also formalizes the use of Reorder() to set the output order of every query.

kylebd99:
Just added an abstract class for each of the three optimization stages (fusion, loop ordering, and formatting) to stages.py.

willow-ahrens (approved):

PR #407, opened 2026-05-23
Exact Stats Implementation

yusheng036:
Hi everyone! 

I have implemented ExactStats that would compute the exact nnz in a tensor computation by using eager function. 

ExactStats handles any operator and also arbitrary matrix contraction. I have added different semirings to test ExactStats against: regular semiring (0, +, ×), tropical semiring(∞, min, +), max-plus semiring (-∞, max, +), max-min semiring (-∞, max, min), and boolean semiring (False, ∨, ∧).

Thank you for checking out my PR


src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,132 @@
...
+class ExactStatsFactory(BaseTensorStatsFactory["ExactStats"]):
+    def __init__(self):
+        super().__init__(ExactStats)
+
+    def copy_stats(self, stat: ExactStats) -> ExactStats:
+        if not isinstance(stat, ExactStats):
+            raise TypeError("copy_stats expected a ExactStats instance")
+        return ExactStats.from_tensor(stat.tensordef.copy(), stat.tensor)
+
+    def _merge(
```
willow-ahrens: Question: Can `ExactStats` just build a `FinchLogic` expression, so calling mapjoin just makes a new `ExactStats` that holds `Mapjoin`? Then when we evaluate `ExactStats` it would just call `compute` with the default scheduler?

willow-ahrens (changes requested):
I think this is more complicated than it needs to be.

yusheng036:
Hi @willow-ahrens, I have changed ExactStats as you requested. Each function in ExactStats call onto its Finch Logic Expression so mapjoin_union calling finch_logic.mapjoin and relabel calling finch_logic.relabel. The new version calls the default scheduler to compute the nnz from the finch logic expression instead.

src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def estimate_non_fill_values(self) -> float:
+        if self.expr is None:
+            return 0.0
+
+        result = get_default_scheduler()(self.expr)
+        if not isinstance(result, TableValue):
+            raise TypeError("estimate_non_fill_value expected a TableValue instance")
+
+        return float(np.count_nonzero(result.tns.to_numpy() != self.fill_value))
```
kylebd99: Converting to dense here is a potential issue. ExactStats will be slow inherently, but converting to dense will limit its usefulness a bit too much. You should be able to compute this value using interface functions (e.g. finchlite.sum(result.tns == self.fill_value)) which will make the computation sparse.

kylebd99 (approved):
I have one small change request, but otherwise LGTM. So, I'm going to go ahead and approve it, and you can merge after that change is made :)

yusheng036:
Hi @willow-ahrens @kylebd99 , I have added count_nonfill_value as an interface function and its respective test in test_interface.py. The function is used in ExactStats in estimate_non_fill_value.

src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,102 @@
...
+        if not isinstance(stat, ExactStats):
+            raise TypeError("copy_stats expected a ExactStats instance")
+        return ExactStats.from_def(stat.tensordef.copy(), stat.expr)
+
+    def _mapjoin_join(
+        self, new_def: TensorDef, op: FinchOperator, join_args: list[ExactStats]
+    ) -> ExactStats:
+        if len(join_args) == 0:
+            return ExactStats.from_def(new_def, None)
+        expr = MapJoin(Literal(op), tuple(s.expr for s in join_args))
```
willow-ahrens: op should be boolean and

src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,102 @@
...
+        return ExactStats.from_def(new_def, expr)
+
+    def _mapjoin_union(
+        self, new_def: TensorDef, op: FinchOperator, union_args: list[ExactStats]
+    ) -> ExactStats:
+        expr = MapJoin(Literal(op), tuple(s.expr for s in union_args))
```
willow-ahrens: op should be Or

src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def aggregate(
+        self,
+        op: FinchOperator,
+        init: Any | None,
+        reduce_indices: tuple[Field, ...],
+        stats: ExactStats,
+    ) -> ExactStats:
+        new_def = TensorDef.aggregate(op, init, reduce_indices, stats.tensordef)
+        expr = Aggregate(Literal(op), Literal(init), stats.expr, reduce_indices)
```
willow-ahrens: Op should be boolean or or boolean and depending on the fill value of the arg.

willow-ahrens: init should be false or true, depending on how init interacts with op

src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,102 @@
...
+        self, stats: ExactStats, reorder_indices: tuple[Field, ...]
+    ) -> ExactStats:
+        new_def = TensorDef.reorder(stats.tensordef, reorder_indices)
+        expr = Reorder(stats.expr, reorder_indices)
+        return ExactStats.from_def(new_def, expr)
+
+
+class ExactStats(NumericStats):
+    def __init__(self, tensor, fields):
+        self.tensordef = TensorDef.from_tensor(tensor, fields)
```
willow-ahrens, line 92: Get the pattern of the tensor here, so you should convert to `tensor != fill`

src/finchlite/autoschedule/tensor_stats/exact_stats.py
```diff
@@ -0,0 +1,102 @@
...
+        return obj
+
+    def estimate_non_fill_values(self) -> float:
+        if self.expr is None:
+            return 0.0
+
+        result = get_default_scheduler()(self.expr)
+        if not isinstance(result, TableValue):
+            raise TypeError("estimate_non_fill_value expected a TableValue instance")
+
```
willow-ahrens, line 111: Here, let's typecast the tensor to a Uint and then sum it

willow-ahrens (changes requested):
Let's construct the inner expression using booleans.

yusheng036:
Hi @willow-ahrens, I have added implementations to construct the inner expressions as booleans. I had to add another case for aggregate where the tensor is neither annihilator and identity.

kylebd99 (approved):
This LGTM! This is a fun stats type. I'm curious whether this could even be used in practice for certain problems, given the right optimizations (e.g. good boolean tensor support + DenseStats/NaiveStats Galley optimization).

willow-ahrens (approved):

PR #408, opened 2026-05-24
rm register property

### LINKED ISSUE #139: Commit the dependency diagram to the repo as documentation

### LINKED ISSUE #260: Enforce FType everywhere
Main change is to wrap numpy and python base types. Then we can start adding properties to some kind of "AbstractElementFtype".

https://data-apis.org/array-api/latest/API_specification/data_types.html

```
def ftype(x):
   if x isa float
      return FinchFloatFType()
   if x isa int32
      return FinchNPInt32FType()
   else
     return x.ftype()
```

need to handle python base types, numpy types, and python tuples. In the future, we can query the array api "is_dtype" and "__array_namespace_info__" for more compatibility stuff.

willow-ahrens:
fixes #260 fixes #139, removes all circular dependecies except one: Tensors have circular runtime dependecies on finch.interface to implement e.g. `__add__`


kylebd99 (approved):
This is very exciting! I think that it all makes sense to me except for the functions being removed from lazy.py. Were those always superfluous or has this change made them superfluous somehow?

willow-ahrens:
I believe that OverrideTensor has implementations for those methods that we were previously redefining. Now we're using OverrideTensor for those, and just setting the override_module to be lazy.py

willow-ahrens:
Note: One thing I found was that overridetensor was actually more complete than lazytensor, so deduplicating these overrides is a good thing for code quality.

PR #510, opened 2026-06-22
MLIR codegen

yusheng036:
- [x] Lowerings for function calls, control flow
- [x] Lowering for FiberTensor, BufferizedNDArray, DenseLevel, and ElementLevel classes
- [x] Generating a Kernel (as a Python string)
- [x] Compile a Kernel (access to MLIR compiler with package)
- [x] Test with dense matmul / dot product


willow-ahrens:
@yusheng036, could you maybe merge main in? I think main has diverged quite a bit since you started.

Edit: oh wait, I see you're already fairly up to date. thanks!

yusheng036:
Hi @willow-ahrens @mtsokol, I have finished my first draft of the mlir backend. I couldn’t find a version of mlir python bindings that was compatible with all the other dependencies, so I created a separate Pixi environment for the MLIR tests and excluded them from the rest of the test. Please check out the PR and let me know what changes is needed.

hameerabbasi:
So I looked into EUDSL, that's just a repo for `mlir-python-bindings` wheels; we don't really need it to merge this PR. We'd need to add to the README how to manually install it; similar to a C/C++ compiler.

mtsokol:
I think it's better to have mlir variant configured and at least one CI job (linux) that runs the tests. 
We will install `mlir-python-bindings` through eudsl registry.

pyproject.toml
```diff
@@ -132,6 +139,22 @@ clang = "*"
...
+[tool.pixi.feature.mlir.dependencies]
+pyyaml = "<=6.0.1"
+
+[tool.pixi.feature.mlir.pypi-options]
+index-url = "https://pypi.org/simple"
+find-links = [{ url = "https://llvm.github.io/eudsl" }]
+
+[tool.pixi.feature.mlir.pypi-dependencies]
+mlir-python-bindings = "*"
+
```
hameerabbasi, line 166: Can we merge/rebase with `main`, move this to `[project.optional-dependencies]`, and pin to a major version?

yusheng036: Hi Hameer, I have moved it to project.optional-dependencies and pinned to the latest version. I have also changed to direct wheel URLs from a single source to prevent supply chain security risk. I made independent links for each platforms.

pyproject.toml
```diff
@@ -132,6 +142,28 @@ clang = "*"
...
+[tool.pixi.feature.mlir.target.linux-aarch64.pypi-dependencies]
+mlir-python-bindings = { url = "https://github.com/llvm/eudsl/releases/download/mlir-python-bindings/mlir_python_bindings-20260715+3f6af3be1-cp312-cp312-manylinux_2_34_aarch64.whl" }
+
+[tool.pixi.feature.mlir.target.osx-arm64.pypi-dependencies]
+mlir-python-bindings = { url = "https://github.com/llvm/eudsl/releases/download/mlir-python-bindings/mlir_python_bindings-20260715+3f6af3be1-cp312-cp312-macosx_13_0_arm64.whl" }
+
+[tool.pixi.feature.mlir.target.win-64.pypi-dependencies]
+mlir-python-bindings = { url = "https://github.com/llvm/eudsl/releases/download/mlir-python-bindings/mlir_python_bindings-20260715+3f6af3be1-cp312-cp312-win_amd64.whl" }
+
```
hameerabbasi: I think the best thing to do here is to use extra index urls, instead of hardcoding by URL.

hameerabbasi (changes requested):

pyproject.toml
yusheng036: @mtsokol mentioned that both sources are combined for resolving packages without priority AFAIC and its a supply chain risk.

hameerabbasi: In `pip`; sure, they have priority but it's not enforced and may fall back. In `uv` and `pixi`; they have priority (unless configured otherwise or overridden on the cmdline). You can even do something like
```toml
[tool.pixi.feature.mlir.pypi-dependencies]
mlir-python-bindings = { version = "...", index = "..." }
```

inside `pypi-dependencies` to change the index for just one dependency.

mtsokol: `index = "..."` specified for the package is also good approach. So that `mlir-python-bindings` is always installed from predefined registry.

yusheng036: Hi @hameerabbasi @mtsokol, I cant make this work.
```
[tool.pixi.feature.mlir.pypi-dependencies]
mlir-python-bindings = { version = "...", index = "..." }
```
i tried:
```
[tool.pixi.feature.mlir.pypi-dependencies]
mlir-python-bindings = { version = "==20260715+3f6af3be1", index = "https://llvm.github.io/eudsl" }
```

but the pixi lock keeps on failing on win64 and osx-arm64. I think the problem is with the index link url. Any suggestions that I can do?

hameerabbasi: Try removing the part of the version after the +, and trying slightly older versions.

yusheng036: I have tried "==20260707" and "==20260616" and they both failed. pixi lock is failing them on every platform.

hameerabbasi: I see the issue; it's that they use `--find-links` which doesn't have a pixi equivalent. They don't have a package index. The URL will need to be updated every so often since old nightlies are deleted; and isn't a sustainable way to get wheels. The API of the nightlies is unstable, too making it a moving target.

I think https://github.com/finch-tensor/finch-tensor-lite/pull/510/commits/5907f97eda50fb4c1c9be4f8ea6bf1777d8475f8 is the best we can do for now.

src/finchlite/codegen/mlir_codegen/mlir.py
```diff
@@ -0,0 +1,973 @@
...
+                raise NotImplementedError(
+                    f"MLIR backend does not yet support floordiv on {arg}"
+                )
+            if MLIROperator.is_unsigned(arg):
+                return "arith.divui"
+            return "arith.floordivsi"
+        case ffuncs.mod:
+            if MLIROperator.is_unsigned(arg):
+                return "arith.remui"
+            raise NotImplementedError(f"MLIR backend does not yet support mod on {arg}")
```
mtsokol, line 289:
```
# arith.remsi/remf truncate, np.mod floors; they differ on
# negative operands, so only unsigned mod is safe to emit
```

src/finchlite/codegen/mlir_codegen/mlir.py
```diff
@@ -0,0 +1,973 @@
...
+            return "arith.muli"
+        case ffuncs.truediv:
+            if MLIROperator.is_float(arg):
+                return "arith.divf"
+            if MLIROperator.is_unsigned(arg):
+                return "arith.divui"
+            return "arith.divsi"
+        case ffuncs.floordiv:
+            if MLIROperator.is_float(arg):
+                raise NotImplementedError(
```
mtsokol, line 280:
```
# arith has no floored float division; divf alone truncates
```

src/finchlite/codegen/buffers/numpy_buffer.py
```diff
@@ -285,3 +297,74 @@ def construct_from_numba(self, numba_buffer):
...
+
+    def mlir_repack(self, ctx: "MLIRContext", lhs, obj):
+        pass
+
+    def serialize_to_mlir(self, obj):
+        from mlir.runtime import get_ranked_memref_descriptor
+
+        src = get_ranked_memref_descriptor(obj.arr)
+        desc = mlir_ctype(self.mlir_type())()
+        ctypes.memmove(ctypes.byref(desc), ctypes.byref(src), ctypes.sizeof(desc))
```
mtsokol, line 354: What does it do?

mtsokol (commented):
First review pass. I pushed changes:
- removed Resize for now - I think it's worth a second thought what exactly we're going to support in terms of MLIR/Python allocated buffers
- Let's stick with eudsl recommended way to install nightly bindings. I opened an issue asking if PyPI is planned https://github.com/llvm/eudsl/issues/480.

mtsokol:
I still want to fully understand the codegen part - otherwise it already LGTM.

pyproject.toml
mtsokol: I think the best long-term approach is to have it on PyPI (or the url becomes a registry) - I asked about it: https://github.com/llvm/eudsl/issues/480

Here we need nightly build - it has fixes we need to run most kernels. Let's stick with the eudsl recommended install for now.

src/finchlite/codegen/mlir_codegen/mlir.py
```diff
@@ -0,0 +1,973 @@
...
+            return "arith.xori -1"
+        case ffuncs.not_:
+            return "arith.xori 1"
+        case MLIROperator():
+            return op.mlir_name()
+        case _:
+            raise NotImplementedError(f"{op} has no MLIR representation.")
+
+
+def mlir_nary_function_call(mlir_name: str, ctx: Any, *args: Any) -> str:
```
mtsokol: I think we know what type of `ctx` is, right?

src/finchlite/codegen/mlir_codegen/mlir.py
```diff
@@ -0,0 +1,973 @@
...
+
+    for a in attrs:
+        if not isinstance(t, StructFType):
+            raise TypeError(f"Expected struct type, got: {t}")
+        if not t.struct_hasattr(a):
+            raise ValueError(f"trying to get missing attr {a!r}")
+        idxs.append(t.struct_fieldnames.index(a))
+        t = t.struct_attrtype(a)
+
+    key = f".getattr:{obj}[{','.join(map(str, idxs))}]"
```
mtsokol, line 451: Why this string format specifically? Do we use that string somewhere?

mtsokol (approved):

mtsokol (approved):

src/finchlite/codegen/buffers/numpy_buffer.py
```diff
@@ -285,3 +297,74 @@ def construct_from_numba(self, numba_buffer):
         Construct a NumpyBuffer from a Numba-compatible object.
         """
         return NumpyBuffer(numba_buffer[0])
+
+    def mlir_type(self):
+        return f"memref<?x{mlir_type(self.element_type)}>"
+
+    def mlir_length(self, ctx: "MLIRContext", buf: MLIRBufferFields):
```
hameerabbasi, line 304:
```suggestion
    def mlir_length(self, ctx: MLIRContext, buf: MLIRBufferFields):
```

And the same for all the other type hints. If there are issues due to recursive imports, use an `if TYPE_CHECKING` block or `from __future__ import annotations`.

hameerabbasi (approved):
LGTM other than the one change.

PR #512, opened 2026-06-23
Lazy implementation of outer product

Calculus0:
Lazy implementation of outer product similar to matmul.

### DIFF
--- a/src/finchlite/interface/lazy.py
+++ b/src/finchlite/interface/lazy.py
@@ -2059,3 +2059,18 @@ def get_lazy_tensor_stats(
         return last_stats[0]
 
     return last_stats
+    
+def outer(x1, x2):
+
+    x1 = lazy(x1)
+    x2 = lazy(x2)
+
+    if x1.ndim != 1:
+        raise ValueError(f"x1 must be a 1D array, got {x1.ndim}D array")
+    if x2.ndim != 1:
+        raise ValueError(f"x2 must be a 1D array, got {x2.ndim}D array")
+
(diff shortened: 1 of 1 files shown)


src/finchlite/interface/lazy.py
```diff
@@ -2059,3 +2059,18 @@ def get_lazy_tensor_stats(
...
+    x2 = lazy(x2)
+
+    if x1.ndim != 1:
+        raise ValueError(f"x1 must be a 1D array, got {x1.ndim}D array")
+    if x2.ndim != 1:
+        raise ValueError(f"x2 must be a 1D array, got {x2.ndim}D array")
+
+    # x1: (N,) -> (N, 1)
+    # x2: (M,) -> (1, M)
+    # multiply broadcasts elementwise to (N, M)
```
willow-ahrens, line 2075: use finch logic directly here instead of relying on other functions, to reduce overhead.

willow-ahrens (changes requested):
Write finch logic directly instead of using existing functions

willow-ahrens:
A few comments:
1. I went ahead and fixed the implementation to use Finch Logic
2. I added tests and exports, and unskipped the "outer product" tests
3. make a separate feature branch with a name other than "main" for new features.

willow-ahrens:
new PR lives here, I'll be rolling it into some other array api changes: https://github.com/finch-tensor/finch-tensor-lite/pull/519

willow-ahrens:
Also, if possible, it helps for you to allow maintainers to edit your PR branch.

PR #514, opened 2026-06-23
Remove Standardize

kylebd99:
This PR aims to remove the LogicStandardizer pass within the autoscheduler.

To do that it:
1. Simplifies `set_loop_order`
2. Adds functions to remove extraneous interior reorders
3. Fixes the output of `get_remainder_query` in AnnotatedQuery to be wrapped in an overwrite agg
5. Regenerates the sddm test file (seems like the simplest possible version though)
4. Changes the set_loop_order test to provide a full `Plan` instead of a `Query`


kylebd99:
Note: It's really easy to break ASV benchmark tests. When you change them to reflect new behavior/interfaces, they report as broken because they're broken on main. We may want to change them to optional.

For this reason, I left standardize.py in the code base, and I can remove it on a cleanup PR.

willow-ahrens (approved):

willow-ahrens:
How should we be handling benchmarks in the future? can we make it less brittle?

kylebd99:
I don't know what's best practice to be honest. @hameerabbasi thinks we should be using codspeed anyways ¯\(ツ)/¯

PR #515, opened 2026-06-23
Bug Fix: Lowering If's + Blank Transposes

kylebd99:
When using the full lowering to numba, transpose queries were leading to empty outputs. 

This seems to be due to the way if was being lowered in the notation->assembly compiler. The   ctx that held the body was never saved anywhere, so writes weren't happening.


kylebd99:
Note: Claude helped find/fix this one. But, I've taken a thorough look, and the fix seems sound.

willow-ahrens (approved):

PR #517, opened 2026-06-24
Cost calculation for formats

Ruchika-2003:


willow-ahrens (changes requested):
tweak: can we call it cost_storage_level?

PR #521, opened 2026-06-29
hn/loop-order-cost

HoangNguyen0309:
loop order cost port


src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -31,6 +33,49 @@
...
+    loop_prefix: tuple[Field, ...],
+    stats_factory: StatsFactory,
+    stats_bindings: dict[Alias, TensorStats],
+) -> float:
+    penalty = 0.0
+    for node in PostOrderDFS(expr):
+        match node:
+            case Table(Alias() as tns, idxs):
+                base = stats_bindings.get(tns)
+                st = stats_factory.relabel(base, tuple(idxs))
```
kylebd99: You probably don't need this relabel call.

src/finchlite/autoschedule/loop_ordering.py
```diff
@@ -31,6 +33,49 @@
...
+                if not is_subsequence(tuple(idxs), loop_prefix):
+                    penalty += st.estimate_non_fill_values()
+            case _:
+                pass
+    return penalty
+
+
+def loop_order_cost(
```
kylebd99: This function is unfortunately not quite right. I had a version that worked like this a while back in previous iterations of the julia code. However, it suffers from a specific problem:

Suppose you have the query: 
E_ijklm = A_ij * B_jk * C_kl * D_lm
Further, suppose that D is entirely zeros (i.e. it's empty). Then, the output will always be empty. This means that your full_stats object will represent an empty tensor, so when you project it down, it will still be empty. Because of this, the i->j->k->l->m and m->l->k->j->i order will look identical. However, in execution, the i->j->k->l->m order will be asymptotically worse because it will basically do the matmul between A and B in the first 3 loops, O(n^3), before realizing that the l,m loops are empty. The m->l->k->j->i order would simply exit immediately in O(1) time.

To fix this, you need to consider the sub-query that a prefix induces. When considering i-j-k, you should estimate the nnz of A_ij*B_jk*C_k. In the julia version, it does this in the function `get_loop_lookups` (https://github.com/finch-tensor/Finch.jl/blob/8b6554b2c73668670037bfb8413425035589d915/src/Galley/TensorStats/cost-estimates.jl#L15). However, it also relies on the function `get_conjuncts_and_disjuncts` (https://github.com/finch-tensor/Finch.jl/blob/8b6554b2c73668670037bfb8413425035589d915/src/Galley/PlanAST/plan.jl#L514) in order to figure out which input tensors in the expression are acting as 'conjuncts' (i.e. join-like arguments) and which ones are acting as 'disjuncts' (i.e. union-like arguments). 

So, to carry over the key logic, you'll have to port both of those functions over.

kylebd99 (changes requested):

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+# End transpose penalty funcs
+
+
+def needs_reformat(stats: TensorStats, prefix: tuple[Field, ...]) -> bool:
+    prefix_pos = {idx: pos for pos, idx in enumerate(prefix)}
+    current_loop = -1.0
+    reformat = False
+    for idx in reversed(stats.index_order):
```
kylebd99: I believe that arrays in finch-tensor-lite are row-major (i.e. an array A[i,j,k] should be accessed in i->j->k order), so this should no longer be reversed. This is a change from the julia version which was column major and was done to be compliant with the standards in the python ecosystem.

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+        idx_loop = float(prefix_pos.get(idx, float("inf")))
+        if idx_loop < current_loop:
+            reformat = True
+        current_loop = idx_loop
+    return reformat
+
+
+# NOTE: Julia version had per variable fraction of nnz.
```
kylebd99: Leave a TODO here. The estimate_non_fill_values function in julia was a bit more advanced and had a notion of "conditional indices" which would be leveraged for this. We may want to add that back soon, but it would be its own PR.

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+            rel_vars = set(stat.index_order) & prefix_set
+            approx_sparsity = _approx_axis_density(stat, rel_vars)
+            is_dense = approx_sparsity > 0.05
+            lookup_factor += SEQ_READ_COST / 5 if is_dense else SEQ_READ_COST
+            continue
+
+        # NOTE: stats format not present in Python
+        # everything below is dead code.
+
+        # index_formats = getattr(stat, "index_formats", None)
```
kylebd99: In the python version, we track the format of tensors using the bindings dictionary, so eventually we will replace this logic with that. 

For the moment, you should use the same _approx_axis_density function to check whether the arg is dense or sparse and use either SEQ_READ_COST/5 or SEQ_READ_COST, respectively.

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+        #    lookup_factor += SEQ_READ_COST
+        lookup_factor += SEQ_READ_COST
+
+    if output_vars is not None and new_var in output_vars:
+        output_list = list(output_vars)
+        new_var_idx = output_list.index(new_var)
+        prefix_positions = [
+            output_list.index(v) for v in new_prefix if v in output_list
+        ]
+        min_var_idx = min(prefix_positions)
```
kylebd99: Similar reasoning about the column vs row major order, we should make this max instead of min.

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+        is_rand_write = new_var_idx != min_var_idx
+        lookup_factor += RANDOM_WRITE_COST if is_rand_write else SEQ_WRITE_COST
+    else:
+        lookup_factor += SEQ_WRITE_COST
+
+    return lookups * lookup_factor
+
+
+# NOTE: remove for Julia version
```
kylebd99: Why is this labeled "remove for julia version"?

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+                if isinstance(base, NumericStats) and not is_subsequence(
+                    tuple(idxs), loop_prefix
+                ):
+                    penalty += base.estimate_non_fill_values()
+            case _:
+                pass
+    return penalty
+
+
+# NOTE: remove for Julia version.
```
kylebd99: Same question here?

tests/test_loop_order_cost.py
```diff
@@ -0,0 +1,150 @@
...
+            t0 = time.perf_counter()
+            expected = _ref_cost(A, B, names)
+            ref_time += time.perf_counter() - t0
+
+            assert got == pytest.approx(expected)
+
+    avg_cost_ms = cost_time / n_costs * 1000
+    avg_ref_ms = ref_time / n_costs * 1000
+    total_time = time.perf_counter() - test_start
+    print(
```
kylebd99: We shouldn't have prints in the test functions.

tests/test_loop_order_cost.py
```diff
@@ -0,0 +1,150 @@
...
+    forward_time = time.perf_counter() - t0
+
+    t0 = time.perf_counter()
+    reverse = loop_order_cost(expr, (m, l_, k, j, i), sf, bindings)
+    reverse_time = time.perf_counter() - t0
+
+    assert forward > reverse
+
+    total_time = time.perf_counter() - test_start
+    print(f"loop_order_cost forward: {forward_time:.4f}s")
```
kylebd99: Same here, no prints in test files.

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,295 @@
...
+    return projected.estimate_non_fill_values()
+
+
+# Transpose penalty funcs more like Julia version
+# TODO: used in bnb
+def cost_of_reformat(stats: TensorStats) -> float:
+    if not isinstance(stats, NumericStats):
+        return 0.0
+    nnz = stats.estimate_non_fill_values()
+    index_formats = getattr(stats, "index_formats", None)
```
kylebd99: Go ahead and just use the nnz  to check whether this is a dense matrix (e.g. >.1 sparsity) rather than calling index_formats. Calling a function that doesn't exist is confusing for the code base.

src/finchlite/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,260 @@
...
+        seen.append(stat)
+
+        if new_var not in set(stat.index_order):
+            continue
+
+        rel_vars = set(stat.index_order) & prefix_set
+        approx_sparsity = _approx_axis_density(stat, rel_vars)
+        is_dense = approx_sparsity > 0.05
+
+        if needs_reformat(stat, new_prefix):
```
kylebd99: Looking at this again, it looks like this branch is useless (i.e. it produces the same thing as the code that follows). When adding the cost of transpositions, we should just what input tensors need to be transposed due to this prefix and compute the total transpose cost once instead of per-index in this for loop. This cost should then be added in line 236.

kylebd99: This would also help to address the unused functions in this PR (e.g. get_reformat_set).

src/finch/autoschedule/loop_order_cost.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+    cost = 0.0
+    for j in range(1, len(loop_order) + 1):
+        cost += get_prefix_cost(
+            loop_order[:j],
+            conjunct_stats,
+            disjunct_stats,
+            stats_factory,
+            output_vars,
+        )
```
kylebd99, line 253: Transpose cost should be added here, not in line 237 to avoid duplicate costs.

willow-ahrens:
@kylebd99, it looks like you requested changes, so you'll need to review again for us to merge. Would you like my review on this one?

kylebd99:
> @kylebd99, it looks like you requested changes, so you'll need to review again for us to merge. Would you like my review on this one?

Comments are always welcome. I think the current plan is to push forward to get a greedy loop orderer in this PR, so that we can show actual speedups on a kernel.

HoangNguyen0309:
```
import time

import numpy as np

import finch as fl
from finch import ffuncs
from finch.autoschedule.compiler import LogicCompiler
from finch.autoschedule.executor import LogicExecutor
from finch.autoschedule.formatter import DefaultLogicFormatter
from finch.autoschedule.loop_order_cost import loop_order_cost
from finch.autoschedule.loop_ordering import (
    DefaultLoopOrderer,
    _heuristic_loop_order,
)
from finch.autoschedule.loop_order_greedy import greedy_loop_order
from finch.autoschedule.normalize import LogicNormalizer
from finch.autoschedule.optimize import DefaultLogicOptimizer
from finch.autoschedule.tensor_stats import DCStatsFactory
from finch.codegen import NumbaCompiler
from finch.compile import NotationCompiler
from finch.finch_assembly import AssemblySimplify, LowerPackedStructSlots
from finch.finch_logic import Alias, Field, Literal, MapJoin, Table


def _time(fn, repeats=5):
    fn()
    return min(_elapsed(fn)[0] for _ in range(repeats))


def _elapsed(fn):
    start = time.perf_counter()
    result = fn()
    return time.perf_counter() - start, result


def _pipeline(use_greedy: bool):
    after = DefaultLogicFormatter(
        LogicCompiler(
            NotationCompiler(
                NumbaCompiler(),
                ctx_transforms=(LowerPackedStructSlots(), AssemblySimplify()),
            )
        )
    )
    return LogicNormalizer(
        LogicExecutor(
            DefaultLogicOptimizer(
                DefaultLoopOrderer(
                    after,
                    loop_order=greedy_loop_order
                    if use_greedy
                    else _heuristic_loop_order,
                )
            ),
            stats_factory=DCStatsFactory(),
            cache=True,
        )
    )


def main():
    rng = np.random.default_rng(42)
    # Greedy time is always 0 despite n becuase D is empty
    n = 32
    a_np, b_np, c_np = (rng.random((n, n)) for _ in range(3))
    d_np = np.zeros((n, n))

    sf = DCStatsFactory()
    i, j, k, l, m = (Field(x) for x in "ijklm")
    A, B, C, D = (Alias(x) for x in "ABCD")
    tables = (Table(A, (i, j)), Table(B, (j, k)), Table(C, (k, l)), Table(D, (l, m)))
    logic = MapJoin(
        Literal(ffuncs.mul),
        tables,
    )
    stats = {
        alias: sf(fl.asarray(array), table.idxs)
        for alias, array, table in zip((A, B, C, D), (a_np, b_np, c_np, d_np), tables)
    }
    h_order = _heuristic_loop_order(logic, sf, stats)
    g_order = greedy_loop_order(logic, sf, stats)

    h_cost_t, h_cost = _elapsed(lambda: loop_order_cost(logic, h_order, sf, stats))
    g_cost_t, g_cost = _elapsed(lambda: loop_order_cost(logic, g_order, sf, stats))

    def _work(cost):
        n_iters = int(min(max(cost, 1.0), 20_000_000))
        return _elapsed(lambda: sum(range(n_iters)))[0]

    h_work_t, g_work_t = _work(h_cost), _work(g_cost)

    expr = fl.einsum(
        "ij,jk,kl,lm->im",
        *(fl.lazy(fl.asarray(array)) for array in (a_np, b_np, c_np, d_np)),
    )
    h_ctx, g_ctx = _pipeline(False), _pipeline(True)
    h_t = _time(lambda: fl.compute(expr, ctx=h_ctx))
    g_t = _time(lambda: fl.compute(expr, ctx=g_ctx))

    print(
        f"Timed work greedy={g_work_t:.4f}s heuristic={h_work_t:.4f}s "
        f"(compute greedy={g_t:.4f}s heuristic={h_t:.4f}s, "
    )


if __name__ == "__main__":
    main()
```

Code for performace of greedy vs heuristic

Matrix mutiplication of A(i,j) @ B(j,k) @ C(k,l) @ D(l,m), with D being empty.

Greedy is always near 0.00s (sees D is empty and orders the indices of D first) while heuristic does i, j, k, l, m

src/finch/autoschedule/loop_ordering.py
```diff
@@ -274,10 +290,16 @@ def rule_1(query):
 
 
 class DefaultLoopOrderer(LogicLoopOrderOptimizer):
```
kylebd99, line 275: I don't think that we want to add a "loop orderer" argument to this class. Instead, we should make a new class for each loop ordering algorithm.

tests/test_loop_order_cost.py
```diff
@@ -0,0 +1,146 @@
...
+            space *= mat.shape[axes[dim]]
+        density = np.count_nonzero(mat) / space if space else 0.0
+        is_dense = density > 0.05
+        lookup_factor += SEQ_READ_COST / 5 if is_dense else SEQ_READ_COST
+
+    lookup_factor += SEQ_WRITE_COST
+    return lookup_factor
+
+
+def _ref_cost(A, B, order, sf, expr, bindings):
```
kylebd99: This seems a bit complicated to me for a test. It seems like bad form to use the functionality being tested for the reference. 

I prefer the alternative approach used in test_empty_relation where you compare the cost of two orders and require that one is better.

kylebd99 (approved):
Had to make a couple of changes:
1) It was using get_loop_lookups not get_prefix_cost. 
2) It wasn't incorporating the cost of transposes.
3) It was raising an error on multi-query programs because of how insert_statistics was being used. 
Going to go ahead and merge with the changes.

PR #523, opened 2026-06-29
Implementing cost calculation for iterating levels

Ruchika-2003:


willow-ahrens (commented):
This is a good draft, but I think that we can probably do a better implementation of a dynamic programming formatter based on SmartFormatter.

PR #529, opened 2026-06-30
I think this was a leftover symlink from somewhere

willow-ahrens:


Ruchika-2003 (approved):
I think this was added by mistake in one of my very first pushes, I do remember us talking about removing it in the very start.

PR #540, opened 2026-06-30
Simplify TensorStats Inheritance

### LINKED ISSUE #307: Refactor TensorStats Inheritance hierarchy
The TensorDef stuff seems like it should live in a superclass TensorStats. child classes can access those values with super, or by doing e.g. self.shape for property getters. Also, not all stats need dims, but they do seem to all need fields and fills, maybe we need a slightly different inheritance hierarchy?

kylebd99:
This PR is a small cleanup within the TensorStats folder. The goal is to remove TensorDef and instead use inheritance to manage the shared functionality.


src/finchlite/autoschedule/tensor_stats/dummy_stats.py
```diff
@@ -1,43 +1,25 @@
...
-        same_axes = all(axes_set[0] == axes for axes in axes_set)
-
-        if not same_axes:
-            new_def.fill_value = 0.0
-
         return DummyStats.from_def(new_def)
 
     def _mapjoin_union(
-        self, new_def: TensorDef, op: FinchOperator, union_args: list[DummyStats]
+        self, new_def: BaseTensorStats, op: FinchOperator, union_args: list[DummyStats]
```
willow-ahrens: This part seems a little strange? I think we should just call super().mapjoin_union?

willow-ahrens (commented):
Some of these methods seem like they should be calling super(), but instead they're passing around the parent where the def used to be, which is a little bit of a mismatch in the programming pattern

src/finchlite/autoschedule/tensor_stats/dummy_stats.py
kylebd99: It's a bit tricky because the class is a factory not a statistic, so calling super().mapjoin creates a BaseTensorStats object. We essentially do this within the BaseTensorStatsFactory.mapjoin implementation, then pass it to the _mapjoin_union/_mapjoin_join functions in the child classes. We can instead do that call within every _mapjoin_union/_mapjoin_join implementation though (and I think that may be cleaner).

kylebd99: Just made that change

kylebd99:
@willow-ahrens I made some changes to clean this up. Lmk if it's more in the vein that you were thinking.

willow-ahrens:
@kylebd99 I think this is better, and I'm happy to merge, but I still wonder if more calls to super() would make sense/be more predictable, instead of passing around base classes. I'm not totally sure, but I'm happy to draft a PR to the PR?

kylebd99:
@willow-ahrens Given that I'm out Tu-Fri this week, I'm going to go ahead and enable auto-merge. I agree that there may still be a better way, but I'm not sure what it looks like due to the factory interface.

willow-ahrens (approved):

PR #545, opened 2026-06-30
Early Return Pipeline

### LINKED ISSUE #544: Add better tooling to run or print from isolated stages in the compiler
Perhaps we could have a "PrintLogic" stage or a "ReturnLogic" stage?

The minimal diff would be to add a stage that stores IR in mutable state, then throws a weird error for an outer harness to catch the intermediate state.

kylebd99:
This PR adds:

1) a no_compute option to LogicExecutor where it returns whatever its interior ctx produces
2) a NoOp stage which immediately returns its inputs

An early return pipeline can then look like, e.g. :
```
LogicNormalizer(
            LogicExecutor( GalleyLogicalOptimizer(DefaultLoopOrderer(NoOpStage()
                    )
                )

```

### DIFF
--- a/benchmarks/test_galley.py
+++ b/benchmarks/test_galley.py
@@ -27,6 +27,7 @@ from finchlite.autoschedule.tensor_stats import UniformStatsFactory
 from finchlite.finch_logic import Alias, Field, Plan, Produces, Query, Table
 from finchlite.finch_notation.interpreter import NotationInterpreter
 from finchlite.symbolic import gensym
+from finchlite.symbolic.stage import NoOpStage
 
 from .utils import patch_benchmark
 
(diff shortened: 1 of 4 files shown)


kylebd99:
Looking at benchmarks/test_galley.py, this PR solves half of the problem. It allows for early return, but it is still a bit inherently complicated to run the first few pipeline stages, intercept the middle, and then benchmark only the remainder. This is because LogicExecutor does things both before and after it calls its child ctx.

willow-ahrens (commented):
I guess I have a question about how no-op stage would interact with the type system? Stages typically have strict outputs so it feels like no-op stage would sometimes break that.

kylebd99:
@willow-ahrens Stages have strict inputs not outputs, so I think type-wise it should be okay. However, I'm not married to this PR in general, if we don't feel like it helps the code.

kylebd99:
I'm going to go ahead and retire this one for the moment.

PR #546, opened 2026-07-01
Use codspeed macro only on workflow dispatch

willow-ahrens:
We only get 600 free minutes so we can't run it on every commit...


hameerabbasi:
Alternative in #547, CodSpeed suggests `simulation` mode on algorithmic stuff anyway.

willow-ahrens:
@hameerabbasi, didn't simulation mode take a long time?

hameerabbasi:
It does; but we'll have unlimited minutes and a better measurement. It's a trade-off.

hameerabbasi (approved):
Feel free to merge either this, or #547, or even to cherry-pick some commits from each. I'm agnostic here.

PR #547, opened 2026-07-01
Compile Galley benchmarks with Numba and skip slow benchmarks in CI.

hameerabbasi:
See https://codspeed.io/docs/instruments


hameerabbasi:
@kylebd99 A lot of time is taken by `benchmarks/test_galley::test_galley_matmul_chain[dense_last-optimize]`.

(The case with `empty_last, metric = False, "optimize"`. It takes 1.87 seconds on my machine.

If we can figure out why optimizing dense stuff with Galley takes so long, we can probably do better.

I'd think you could calculate the stats of dense levels simply by counting the # of elements, you wouldn't need to mask/sum at all. Is that what's taking the time here?

Edit: By the way, the memory overhead is due to the Numba compilation, I'd assume. I can just add a warmup pass to get rid of it.

Edit 2: Full report (with warmup pass before the benchmark)

```
                                        Benchmark Results
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┓
┃                                       Benchmark ┃ Time (best) ┃ Rel. StdDev ┃ Run time ┃ Iters ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━┩
│   test_galley_matmul_chain[empty_last-optimize] │     101.0ms │        1.3% │    3.01s │    29 │
│ test_galley_matmul_chain[empty_last-downstream] │      66.2ms │        0.8% │    2.95s │    44 │
│   test_galley_matmul_chain[dense_last-optimize] │       1.87s │        0.0% │    1.87s │     1 │
│ test_galley_matmul_chain[dense_last-downstream] │      67.2ms │        2.5% │    2.98s │    43 │
│               test_f1_jit[compile_numba_galley] │     509.0ms │        0.2% │    2.55s │     5 │
│                      test_f1_jit[compile_numba] │     510.7ms │        0.7% │    2.58s │     5 │
│         test_jit_creation[compile_numba_galley] │     81.11µs │        2.1% │    3.02s │ 1,587 │
│                test_jit_creation[compile_numba] │     80.91µs │        0.9% │    3.00s │ 1,587 │
│    test_ops_binary[compile_numba_galley-matmul] │     156.4ms │        0.4% │    2.99s │    19 │
│       test_ops_binary[compile_numba_galley-add] │     141.6ms │        1.1% │    3.01s │    21 │
│  test_ops_binary[compile_numba_galley-multiply] │     141.6ms │        1.3% │    3.02s │    21 │
│           test_ops_binary[compile_numba-matmul] │     158.4ms │        0.9% │    2.89s │    18 │
│              test_ops_binary[compile_numba-add] │     145.3ms │        0.8% │    2.93s │    20 │
│         test_ops_binary[compile_numba-multiply] │     145.1ms │        0.5% │    2.92s │    20 │
│    test_ops_reduction[compile_numba_galley-sum] │      88.8ms │        1.7% │    3.01s │    33 │
│           test_ops_reduction[compile_numba-sum] │      89.6ms │        1.2% │    2.99s │    33 │
└─────────────────────────────────────────────────┴─────────────┴─────────────┴──────────┴───────┘
```

willow-ahrens:
@hameerabbasi is the TL;DR here that simulation wasn't working originally because we were running something big through the notation interpreter, but when we compile numba simulation is manageable? It's still taking 40+ minutes, right? How long did it take originally?

willow-ahrens:
A small suggestion, we probably only need one compile galley benchmark (matmul), rather than three. @kylebd99 is that right?

hameerabbasi:
> @hameerabbasi is the TL;DR here that simulation wasn't working originally because we were running something big through the notation interpreter, but when we compile numba simulation is manageable? It's still taking 40+ minutes, right? How long did it take originally?

Yeah; that's pretty accurate.

I think I let it run for 1h20m or so before giving up and switching to `walltime`.

kylebd99:
Yeah, we can go ahead and just do the empty-last one for these benchmarks. The reason that dense-last takes a long time is that every plan has equal cost, so we can't do any pruning during the plan search. This means we basically end up considering 2^10 plans. Eventually, we'll likely want to default Galley to a bounded runtime search algorithm (e.g. beam search or stochastic search).

hameerabbasi:
Down to 15m. I think this is ready, @willow-ahrens. All remaining benchmarks are under 200 ms on my machine.

willow-ahrens (approved):

PR #549, opened 2026-07-02
Concept for Concatenation

willow-ahrens:
what do we think?


src/finchlite/interface/eager.py
```diff
@@ -540,6 +540,26 @@ def sign(x):
 # https://data-apis.org/array-api/2024.12/API_specification/manipulation_functions.html
 
 
+def _flatten_for_concat(array):
+    array = lazy.asarray(array)
+    size = int(np.prod(array.shape, dtype=np.intp)) if array.shape else 1
```
hameerabbasi:
```suggestion
    size = functools.reduce(operator.mul, array.shape, 1)
```

Allows Python bigints to do stuff for values `>= 2 ** 64`; also is faster for plain integers than NumPy.

hameerabbasi (commented):
Is this a lazy `concat` that does a binary search to get an index into an underlying array? Nice.

src/finchlite/interface/eager.py
willow-ahrens: good catch!

willow-ahrens:
@hameerabbasi I'm rolling this into the larger PR here https://github.com/finch-tensor/finch-tensor-lite/pull/548

PR #553, opened 2026-07-05
Add Julia Backend

### LINKED ISSUE #543: Julia Backend
We should integrate finch-tensor as an optional pixi feature.

willow-ahrens:
fixes #543, largely copying @sreevickrant's implementation from https://github.com/finch-tensor/finch-tensor/pull/147.


sreevickrant:
Did we not need a `JLTensorFType` and a separate `asarray` construction pathway?

willow-ahrens:
No, I switched to reuse the existing level definitions in python, so we construct the julia tensor only at the call boundary when we call the kernel.

sreevickrant (approved):

PR #558, opened 2026-07-14
Remove standardize.py

kylebd99:
Does what it says on the tin.

1. Moves utility functions out of standardize.py and into loop_ordering.py and util.py
2. Deletes standardize.py
3. Fixes imports.


tests/test_scheduler.py
```diff
@@ -723,78 +717,3 @@ def test_scheduler_e2e_sddmm(file_regression):
     file_regression.check(
         str(plan_opt), extension=".txt", basename="test_scheduler_e2e_sddmm_plan"
     )
-
-
-def test_logic_standardizer_inplace(file_regression):
```
hameerabbasi, line 728: Did we move this test somewhere; or just delete it?

kylebd99 (commented):
> Did we move this test somewhere; or just delete it? 

Currently, deleted it because LogicStandardizer no longer exists. We may want to change it to something like test_pipeline_inplace. I'll go back and add a new version.

willow-ahrens (approved):

PR #563, opened 2026-07-15
LPBound Statistics

kylebd99:
Implementing LP frequency norms + the lp flow-bound as a new LPStats implementation of TensorStats. 
The new implementation should be provably as tight or tighter than the DCStats bound, but it will also be slower because it uses a linear program solver (scipy.optimize.linprog).


src/finchlite/autoschedule/tensor_stats/dc_stats.py
```diff
@@ -120,7 +111,18 @@ def aggregate(
 
     def relabel(self, stats: DCStats, relabel_indices: tuple[Field, ...]) -> DCStats:
```
kylebd99: Additionally caught a relabel issue where the DCs weren't being relabeled. Doesn't come up much because we always immediately remove relabel nodes.

src/finchlite/autoschedule/tensor_stats/lp_stats.py
```diff
@@ -0,0 +1,369 @@
...
+
+class LPStatsFactory(BaseTensorStatsFactory["LPStats"]):
+    def __init__(self, ps: Iterable[float] = DEFAULT_PS):
+        super().__init__(LPStats)
+        self.ps: tuple[float, ...] = tuple(ps)
+
+    def __call__(self, tensor: Any, fields: tuple[Field, ...]) -> LPStats:
+        return LPStats(tensor, fields, self.ps)
+
+    def _mapjoin_union(self, op: FinchOperator, *union_args: LPStats) -> LPStats:
```
kylebd99: The factory is roughly identical across DCStats and LPStats. We may even want to just merge these two into one class and distinguish them via constructor arguments.

willow-ahrens (approved):
This is great! Thanks! Super cool use of scipy stuff, which we already depend on. I have two suggestions you could solve now or we can make issues for:
1. the util.py file feels like it could be simplified using the default scheduler.
2. Do we think that maybe dc_stats and lp_stats should share a parent class?

kylebd99:
The degree_count_scan currently doesn't use FinchLogic in order to avoid doing multiple scans over the tensor. Naively, it would need to do a scan per-dimension, and the FinchLogic optimizer isn't smart enough to avoid this yet. We would need some basic loop fusion to make that work.

PR #570, opened 2026-07-22
Create docs action and bump versions of other actions.

hameerabbasi:
- **Docs publish and artifact actions.**
- **Update publish job.**
- **Update CI job.**
- **Update Benchmark job.**

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,17 +3,16 @@ on:
   workflow_dispatch:
 jobs:
   build:
-    #if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v7
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v7
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -31,13 +30,11 @@ jobs:
...
       - name: Build and publish to Prefix.dev
         run: pixi publish --target-channel https://prefix.dev/${{ secrets.PREFIX_CHANNEL_NAME }}
-      - name: Create Release
-        uses: actions/create-release@v1
-        if: success()
+      - name: Create a Release
+        uses: softprops/action-gh-release@v3
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
willow-ahrens, line 36: Did this name change?

willow-ahrens (approved):

.github/workflows/publish.yml
hameerabbasi: Yeah, the old action was archived. I updated it to a maintained one; and it has a different input name.

PR #573, opened 2026-07-23
Scipy.sparse constructor

### LINKED ISSUE #310: `scipy.sparse` constructors

yusheng036:
Fixes #310 

[EDIT ~Mateusz]
The PR adds `scipy.sparse` interop in two pieces:

1. `lazy.asarray` now accepts scipy sparse inputs: CSR and COO convert zero-copy into FiberTensor; other formats (LIL, DOK, DIA, BSR) are routed through `asformat("coo")`; CSC deliberately passes through unchanged so the linalg wrappers (det, lu, …) keep using scipy directly.
2. `FiberTensor.from_scipy_csr` / `from_scipy_coo` classmethods and a real `to_scipy` that pattern-matches DenseLevel→SparseListLevel→ElementLevel to csr_array and SparseCOOLevel→ElementLevel to coo_array, both zero-copy.


src/finch/interface/eager.py
```diff
@@ -502,7 +503,7 @@ def det(x, /):
     # https://stackoverflow.com/questions/19107617/how-to-compute-scipy-sparse-matrix-determinant-without-turning-it-to-dense
     x = _warn_compute(x, "det")
     try:
-        x_sp = to_scipy(lazy.asarray(x))
+        x_sp = x if scipy_sparse.issparse(x) else to_scipy(lazy.asarray(x))
```
willow-ahrens: this will never be true right? even if it was, we should be able to round-trip through lazy.asarray without copying.

src/finch/interface/eager.py
```diff
@@ -516,7 +517,8 @@ def det(x, /):
 def lu(x, /, *, permute_l=False, p_indices=False):
     x = _warn_compute(x, "lu")
     try:
-        return scipy_sparse_linalg.splu(to_scipy(lazy.asarray(x)).tocsc())
+        x_sp = x if scipy_sparse.issparse(x) else to_scipy(lazy.asarray(x))
```
willow-ahrens: same here

src/finch/interface/lazy.py
```diff
@@ -395,6 +397,28 @@ def asarray(
...
+                case "csr":
+                    pass
+                case "coo":
+                    raise NotImplementedError("SciPy COO format is not supported.")
+                case "csc":
+                    raise NotImplementedError("SciPy CSC format is not supported.")
+                case format_name:
+                    raise NotImplementedError(
+                        f"SciPy sparse format {format_name!r} is not supported."
+                    )
```
willow-ahrens, line 406: this is an awful lot of notimplemented. can we have these cases convert to csc and then call transpose?

src/finch/tensor/fiber_tensor.py
```diff
@@ -291,10 +294,31 @@ def to_numpy(self) -> np.ndarray:
...
+                    lvl=ElementLevel() as element,
+                    ptr=ptr,
+                    idx=idx,
+                )
+            ):
+                return scipy_sparse.csr_array(
+                    (element.val.arr, idx.arr, ptr.arr),
+                    shape=self.shape,
+                    copy=False,
+                )
```
willow-ahrens, line 143: can't we at least do coo and dense too?

src/finch/tensor/fiber_tensor.py
```diff
@@ -430,6 +454,44 @@ def from_numpy(self, arr: np.ndarray) -> FiberTensor:
             _device=self.device,
         )
 
+    @staticmethod
+    def from_scipy(obj, device=None):
```
willow-ahrens: I feel like this should mirror the finch asarray call.

willow-ahrens (changes requested):
Thanks for the help! This PR could implement a little more cases before it's ready.

src/finch/tensor/fiber_tensor.py
yusheng036: I wanted to do a format specific from_scipy function (e.g from_scipy_csr, from_scipy_csc) so I let asarray handle the copy handling, dtype handling and format then I hand them to format specific constructors.

src/finch/tensor/level/dense_level.py
```diff
@@ -170,6 +166,8 @@ def level_unfurl(
         proto,
         pos: asm.AssemblyExpression,
     ):
+        from finch.tensor.fiber_tensor import FiberTensorFType
```
willow-ahrens: Make an issue to fix this dynamic import

src/finch/tensor/level/level.py
```diff
@@ -0,0 +1,202 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+import numpy as np
+
+from finch.algebra import (
+    FType,
+    FTyped,
+)
+
```
willow-ahrens, line 10: Rename this file to level.py and put it inside the level directory

yusheng036:
Hi @willow-ahrens @mtsokol, the scipy constructor is ready to be reviewed and merged. 

I have untangled the circular dependency in FiberTensor and moved Level and LevelFType to level/level.py. The scipy conditions in lazy.asarray was followed from the julia version as seen here: https://github.com/finch-tensor/finch-jl-python/blob/main/src/finch/tensor.py with the inclusion of dtype conditions. The to_scipy supports Dense/Dense, Dense/SparseList and SparseCOO while the from_scipy in lazy.asarray supports CSC and CSR, while converting the rest to .tocsc(). I added a TransposedFiberTensor that inherits from OverrideTensor. It is a zero-copy transposed of FiberTensor. Its introduction was to fix the subscriptable nature of earlier FiberTensor.

willow-ahrens:
The issue with the tensor not being subscriptable was that you need to use the julia backend to transpose the tensor, right? can you try the julia backend to see if we can get an eager transpose working?

src/finch/tensor/fiber_tensor.py
```diff
@@ -291,16 +109,114 @@ def to_numpy(self) -> np.ndarray:
...
+                    ptr=ptr,
+                    idx=idx,
+                )
+            ):
+                return scipy_sparse.csr_array(
+                    (element.val.arr, idx.arr, ptr.arr),
+                    shape=self.shape,
+                    copy=False,
+                )
+            case DenseLevel(
```
mtsokol: Let's remove 2D dense from here (and fallback to `NotImplementedError`).

IMO `from/to_scipy` should only convert scipy COO/CSR to our representation and back. Converting 2D dense is also changing the format implicitly.

src/finch/tensor/fiber_tensor.py
mtsokol: Let's have COO and CSR in `to_scipy` for now. Dense would include changing the format which I believe shouldn't happen here.

src/finch/tensor/fiber_tensor.py
```diff
@@ -291,16 +109,114 @@ def to_numpy(self) -> np.ndarray:
...
+                    f"Finch format {self.ftype} is not supported by SciPy conversion."
+                )
 
     def item(self):
         if self.ndim != 0:
             raise ValueError("Cannot convert non-scalar tensor to Python scalar.")
         return self.to_numpy().item()
 
+    def transpose(self):
+        return TransposedFiberTensor(self)
```
mtsokol: For the first iteration, can we only have COO and CSR supported? For supporting CSC, which required `TransposedFiberTensor` here, I'd prefer to have more thorough discussion. But let's have this idea written down in an issue!

EDIT: Or can you use `matrix_transpose` instead that we already have instead for supporting CSC? `transpose` isn't present in [the Array API, but rather matrix_transpose](https://data-apis.org/array-api/latest/API_specification/generated/array_api.matrix_transpose.html) so let's use that.

src/finch/tensor/fiber_tensor.py
```diff
@@ -431,6 +347,53 @@ def from_numpy(self, arr: np.ndarray) -> FiberTensor:
         )
 
 
+@dataclass
+class TransposedFiberTensor(OverrideTensor):
```
mtsokol: Let's remove it. (Either use existing `matrix_transpose` or skip CSC for now?)

src/finch/interface/lazy.py
```diff
@@ -10,6 +10,7 @@
 from typing import Any, cast, overload
 
 import numpy as np
+import scipy.sparse as scipy_sparse
```
mtsokol: Let's use `import scipy.sparse as sps` convention in all places.

src/finch/tensor/level/level.py
mtsokol, line 1: Let's rename it to `abstract_level.py`.

src/finch/tensor/masks.py
```diff
@@ -7,7 +7,8 @@
 from finch.algebra import ImmutableStructFType, ffuncs
 from finch.compile import looplets as lplt
 
-from .fiber_tensor import FiberTensor, FiberTensorFType, Level, LevelFType
+from .fiber_tensor import FiberTensor, FiberTensorFType
+from .level.level import Level, LevelFType
```
mtsokol: I think `from .level import Level, LevelFType` should be enough, here and other places.

tests/test_tensor.py
```diff
@@ -551,3 +558,136 @@ def test_lazy_array_api_matrix_functions():
...
+            np.array([0, 2, 3], dtype=np.int32),
+        ),
+        shape=(2, 3),
+    )
+
+    tensor = asarray(csr)
+
+    assert isinstance(tensor, FiberTensor)
+    assert isinstance(tensor.lvl, DenseLevel)
+    assert isinstance(tensor.lvl.lvl, SparseListLevel)
```
mtsokol, line 577: Let's also assert that the buffers inside `SparseListLevel` are as expected.

src/finch/tensor/fiber_tensor.py
```diff
@@ -291,16 +109,114 @@ def to_numpy(self) -> np.ndarray:
         return np.reshape(self.lvl.val.arr, self.shape, copy=False)
 
     def to_scipy(self):
```
mtsokol, line 105: Let's have `to_scipy` and `from_scipy_scr` (and a new `from_scipy_coo`) here with the `from .level import ...` import as the first line. This way we can avoid circular import and revert:
```py
from finch.tensor.fiber_tensor import (
    FiberTensorFType
)
```
(Keeping Level classes in a separate file is a good idea).
Requiring FiberTensor to know about Levels breaks the logic here but I think we can have it with `def`-level import for now.

In the future we should to: renaming `FiberTensor` to some `CoreFiberTensor`, moving `to_scipy`, `from_scipy`, `to_numpy` etc. to mixins, e.g. `SciPyTensorMixin`, `NumPyTensorMixin`, and exposing to user only public API with `FiberTensor(CoreFiberTensor, SciPyTensorMixin, NumPyTensorMixin, ...)`.
This way we won't face it. @willow-ahrens WDYT?

src/finch/tensor/level/dense_level.py
```diff
@@ -9,11 +9,7 @@
 from finch.compile import AssemblyContext, LoopletContext
 from finch.compile import looplets as lplt
 from finch.compile.lower import SymbolicExtent
-from finch.tensor.fiber_tensor import (
-    FiberTensorFType,
```
mtsokol, line 13: Let's make sure we can import this item.

mtsokol (changes requested):
It's on track overall! What we still need:
- code structure: importing levels inside `to_scipy` so that we avoid circular import on importing FiberTensorFType in levels. 
- For CSC can we use `transpose_matrix` instead of a class or skip CSC support for this batch?

yusheng036:
Hi @willow-ahrens, I have tried calling asarray on a csc matrix with eager matrix transpose on a julia backend. It computes successfully but it would always return as a BufferizedNDArray as a dense tensor due to this function:
```
https://github.com/finch-tensor/finch-tensor/blob/56c0a4e3dfa61f41ae995ccd0495ef7aaa09ea23/src/finch/compile_jl/interop.py#L213

```

src/finch/tensor/fiber_tensor.py
willow-ahrens: @mtsokol, can you join the finch meeting so we can discuss a little on this one?

willow-ahrens (changes requested):
make an issue about transpose to make it work

src/finch/tensor/fiber_tensor.py
mtsokol: Sure! I'll be able to join first 30min of the meeting.

src/finch/tensor/level/dense_level.py
mtsokol: It's gone now.

src/finch/tensor/level/level.py
mtsokol: I thought that `abstract_level.py` but `level.py` can stay - reverted.

mtsokol (approved):
For now I reverted CSC support, now scipy object passes through as before. LGTM!

src/finch/tensor/fiber_tensor.py
mtsokol: For sanity I updated the description saying that here we provide the constructors for CSR and COO that we fully support. For CSC (and transpositions) let's have it in a separate PR after we decide how to go about it.

willow-ahrens (approved):

PR #579, opened 2026-07-24
skip mlir tests when `mlir` is not available

lucascolley:
https://github.com/finch-tensor/finch-tensor/issues/567#issuecomment-5066707979 @hameerabbasi

### DIFF
--- a/tests/test_codegen.py
+++ b/tests/test_codegen.py
@@ -1004,210 +1004,217 @@ def test_e2e_transpose_numba(a, dtype):
     finch_assert_equal(result, a.T)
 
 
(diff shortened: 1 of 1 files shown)


lucascolley:
maybe there is a way to roll this into `@pytest.mark.mlir_backend`, but I guess this is simple enough. Review with hide whitespace!

hameerabbasi:
> maybe there is a way to roll this into `@pytest.mark.mlir_backend`, but I guess this is simple enough. Review with hide whitespace!

Hehe, see #580, but you were faster.

PR #580, opened 2026-07-24
Make MLIR tests optional.

hameerabbasi:


tests/utils.py
```diff
@@ -0,0 +1,16 @@
...
+try:
+    import mlir  # noqa: F401
+
+    MLIR_AVAILABLE = True
+except ImportError:
+    MLIR_AVAILABLE = False
+
+skip_mlir = pytest.mark.skipif(
+    not MLIR_AVAILABLE, reason="MLIR Python bindings not installed."
+)
```
lucascolley: `pytest.importorskip("mlir")`

EDIT: or do we need the skip to be delayed such that that doesn't work?

hameerabbasi: That's a module-like-thing that skips on `getattr`, not a decorator, IIUC.

hameerabbasi: Yeah, doesn't work. See https://github.com/finch-tensor/finch-tensor/pull/580/commits/ae385239f8eae52372780151ab20e6d6fe4743ff for slightly less lines.

tests/utils.py
```diff
@@ -0,0 +1,14 @@
...
+except ImportError:
+    mlir = None
+
+skip_mlir = pytest.mark.skipif(
+    mlir is None, reason="MLIR Python bindings not installed."
+)
+
+
+def mlir_backend(func):
+    return pytest.mark.mlir_backend(skip_mlir(func))
```
lucascolley: Would this work?
```suggestion
def mlir_backend(func):
    _ = pytest.importorskip("mlir")
    return pytest.mark.mlir_backend(func)
```

tests/test_codegen.py
```diff
@@ -1019,7 +1020,7 @@ def test_e2e_transpose_numba(a, dtype):
         ),
     ],
 )
-@pytest.mark.mlir_backend
+@mlir_backend
```
lucascolley, line 1023: It would reduce duplication a bit to have a separate `TestMLIR` class

hameerabbasi: I feel like a topic for another PR; it's also spread out across files.

tests/utils.py
hameerabbasi: Nope, since `pytest.importorskip` runs at import-time, not at test-time. But a variant in https://github.com/finch-tensor/finch-tensor/pull/580/commits/9055e2abd85f6522d8b861c8e77ab28ef05c0e6e did the trick.

lucascolley (approved):

tests/utils.py
```diff
@@ -0,0 +1,12 @@
+from functools import wraps
+
+import pytest
+
+
+def mlir_backend(func):
+    @wraps(func)
+    def func_wrapper(*args, **kwargs):
+        _ = pytest.importorskip("mlir", reason="MLIR Python bindings not installed.")
+        return func(*args, **kwargs)
```
willow-ahrens, line 10: Other tests are optional too, but are just handled with inline @pytest.mark.mlir_backend. Is there a reason that couldn't work here?

willow-ahrens (approved):
Quick question, otherwise looks good. If you could make it look more like the other skippable test suites I would appreciate the uniformity. In general, I don't like adding helpers if they aren't saving many lines, because it increases the amount of finch-specific code people have to learn.

tests/utils.py
mtsokol: This PR removes all `@pytest.mark.mlir_backend` decorators from the code. So all MLIR tests are now executed only if MLIR is importable. Does it answer the question?

tests/utils.py
```diff
@@ -0,0 +1,12 @@
+from functools import wraps
+
+import pytest
+
+
+def mlir_backend(func):
```
mtsokol, line 6: @hameerabbasi Why is it in a `utils.py` file instead of `conftest.py`?

hameerabbasi: Personal preference. I prefer to put test utilities in a `utils.py` file, and PyTest configuration in the `conftest.py` (which is the original purpose).

mtsokol (approved):
LGTM, so I think after addressing the last question about uniformity it's ready.

PR #581, opened 2026-07-24
Blocked uniform implementation

Ruchika-2003:


willow-ahrens (commented):
Is this missing a file or something? I only see a constructor in blocked_uniform.py

Ruchika-2003:
I have separated the logic for creating blocks for blcoked_uniform, rest functions stay the same. I am adding test cases for this too

willow-ahrens:
The point of this was to make a dedicated stats object for blockuniform so that estimation could be faster, each operation like mapjoin, aggregate, etc. would need to be specialized for it to be faster. What do you believe the point of the current PR is?

Ruchika-2003:
The earlier blocks implementation was sub-optimal for uniform as it just needed one parameter per block so the entire block creation path of creating the stats object by wrapping it and then unwrapping it to get nnz was redundant so I avoided that in the current implementation, but yes I see what I missed. I will revise and push the updated code today.

willow-ahrens:
Okay, sure thing! If we're trying to make it faster, perhaps try to run a few performance tests on the implementation too to double-check performance improves.

src/finch/autoschedule/tensor_stats/blocked_stats.py
```diff
@@ -229,6 +231,9 @@ def build_grid(
         stats_factory: StatsFactory[NumericStats],
         data: Any,
     ) -> np.ndarray:
+
```
willow-ahrens, line 234: Let's not modify blocked_stats.py, if the user wants uniformstats they can request it directly.

willow-ahrens (changes requested):
@Ruchika-2003 This PR is on the right track. Would you please get this PR ready to merge and checkbox green? Also, please try to simplify blocked_uniform.py a little? remove tiny helpers, try to make better use of list comprehensions.

willow-ahrens (approved):

PR #584, opened 2026-07-28
DP cost formatter implementation

Ruchika-2003:


src/finch/autoschedule/smart_formatter.py
```diff
@@ -118,3 +119,54 @@ def get_tensor_ftype(
...
+
+class CostFormatter(SmartFormatter):
+    def __init__(self, loader: LogicLoader | None = None):
+        super().__init__(loader)
+        self._stats_factory = None
+
+    @abstractmethod
+    def cost_fn(self, name, num_pos, n_l, nnz_l, val_size, pos_size):
+        """Define cost using different methods like storage
+        based or iteration based cost"""
```
willow-ahrens, line 132: Let's document what all the args mean

willow-ahrens: Also, I think we probably want to define a type called like: LevelOption or something that has a cost_fn and also a constructor built in, so that the CostFormatter holds a list of LevelOption objects which can be used to represent each choice of level type and it's cost.

src/finch/autoschedule/smart_formatter.py
```diff
@@ -118,3 +119,54 @@ def get_tensor_ftype(
...
+            self._stats_factory,
+            fill_value,
+            self.cost_fn,
+        )
+        return fiber_tensor(lvl)
+
+
+class StorageCostFormatter(CostFormatter):
+    def cost_fn(self, name, num_pos, n_l, nnz_l, val_size, pos_size):
+        if name == "leaf":
```
willow-ahrens, line 159: Instead of the string "leaf", etc, I think it would be nicer to use objects here.

src/finch/autoschedule/stats_format_selector.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
```
willow-ahrens, line 1: Is this file redundant now in favor of smart_formatter.py?

willow-ahrens: Or should this merge into that file?

src/finch/autoschedule/stats_format_selector.py
```diff
@@ -0,0 +1,82 @@
...
+        if reduce_fields:
+            reduced = stats_factory.aggregate(ffuncs.or_, False, reduce_fields, stats)
+        else:
+            reduced = stats
+        return reduced.estimate_non_fill_values()
+
+    # memoizing (l,nnz) : (best_cost,best_fmt)
+    memo = {}
+
+    def rec(level, num_pos):
```
willow-ahrens, line 46: solid dynamic programming. Perhaps just double check the number of calls to rec to see that it's less than n^2

src/finch/tensor/level/sparse_list_level.py
```diff
@@ -93,6 +93,49 @@ def ptr_type(self):
...
+                ffuncs.or_, False, reduce_fields, stats
+            )
+        else:
+            reduced_stats = stats
+        nnz_prefix = reduced_stats.estimate_non_fill_values()
+
+        cost_sparse = num_pos + nnz_prefix
+        return cost_sparse + self.lvl_t.level_iter_cost(
+            fields, stats, stats_factory, nnz_prefix, lvl + 1
+        )
```
willow-ahrens, line 118: Okay, these cost functions should live in the smart_formatter file I think, rather than in the level source code.

src/finch/tensor/level/element_level.py
```diff
@@ -29,6 +29,17 @@ def struct_fields(self):
             ("val", self.buffer_type),
         ]
```
willow-ahrens, line 31: Perhaps we can just merge your "calculate size" Pr into this one, so that each level has a function to calculate size, and then we can use that to double-check the optimization algorithm against brute-force?

willow-ahrens (changes requested):
@Ruchika-2003 This PR is on the right track. Would you please get this PR ready to merge and checkbox green?

benchmarks/stats.py
```diff
@@ -0,0 +1,168 @@
+# Yu's code to generate the plots
```
willow-ahrens: This code is good to have, but belongs in a different repo

willow-ahrens: Or at the very least, if this code lives here we should define a codspeed benchmark to track the accuracy

src/finch/autoschedule/smart_formatter.py
```diff
@@ -117,3 +219,98 @@ def get_tensor_ftype(
...
+def build_dense(lvl, dim_type):
+    return dense(lvl, dim_type)
+
+
+def build_sparse_hash(lvl, dim_type):
+    return sparse_hash(lvl, dim_type, single_writer=False)
+
+
+CANDIDATES = (
+    LevelOption(
```
willow-ahrens: Could you make a class hierarchy for these instead of dataclasses with lambdas?

src/finch/autoschedule/smart_formatter.py
```diff
@@ -117,3 +219,98 @@ def get_tensor_ftype(
...
+        build=build_dense,
+        next_num_pos=dense_next_num_pos,
+        storage_cost_fn=lambda num_pos, n_l, nnz_l, val_size, pos_size: 0.0,
+        iter_cost_fn=lambda num_pos, n_l, nnz_l, val_size, pos_size: num_pos * n_l,
+    ),
+    LevelOption(
+        level_type=SparseHashLevelFType,
+        build=build_sparse_hash,
+        next_num_pos=sparse_hash_next_num_pos,
+        storage_cost_fn=(
```
willow-ahrens: instead of storage_cost_fn and iter_cost_fun, the parent class should implement cost_fn and then child classes implement different kinds of costs, so you could build a hierarchy

```
LevelOption:
   SparseLevelOption:
      IterCostSparseLevelOption
      StorageCostSparseLevelOption
   DenseLevelOption:
      IterCostDenseLevelOption
      StorageCostDenseLevelOption
```

willow-ahrens (changes requested):

willow-ahrens:
@Ruchika-2003 What happened to this PR?

Ruchika-2003:
@willow-ahrens This looks like all the comments are addressed and it passes the checks too, can you review and merge this.

willow-ahrens (approved):

PR #586, opened 2026-07-31
Recover sparse levels on Julia roundtrip

sreevickrant:


src/finch/tensor/fiber_tensor.py
```diff
@@ -287,14 +291,24 @@ def buffer_factory(self):
         return self.lvl.buffer_factory
 
     def to_numpy(self) -> np.ndarray:
```
sreevickrant, line 101: @willow-ahrens wasn't sure about this

src/finch/tensor/fiber_tensor.py
```diff
@@ -287,14 +291,24 @@ def buffer_factory(self):
...
-        # TODO: temporary for dense only. TBD in sparse_level PR
-        return np.reshape(self.lvl.val.arr, self.shape, copy=False)
+        out = np.full(self.shape, self.fill_value, dtype=self.lvl.val.arr.dtype)
+        for idx, val in self.lvl.iter_entries(int(self.pos)):
+            out[idx] = val
+        return out
 
     def to_scipy(self):
```
mtsokol, line 105: Hi @sreevickrant,
In https://github.com/finch-tensor/finch-tensor/pull/573 we implement constructing scipy's COO and CSR by pattern matching our FiberTensors' structures, which I think we would prefer over `iter_entries` interface. 
We will merge it soon so could you keep here the Julia -> Python `interop.py` only? WDYT?

mtsokol (changes requested):

src/finch/tensor/fiber_tensor.py
sreevickrant: I'll roll back the changes for `iter_entries`. It was a temporary workaround.

willow-ahrens (approved):

src/finch/tensor/fiber_tensor.py
willow-ahrens: Let's remove all of the scipy stuff from this PR

mtsokol (approved):

PR #589, opened 2026-07-31
Boeing matmul implementation

Ruchika-2003:


willow-ahrens:
Do not commit large files to the codebase. If you want to use a suitesparse matrix, use ssgetpy: https://github.com/drdarshan/ssgetpy

willow-ahrens:
Please remove the large file and rebase, when you get the chance. Thanks!

src/finch/tests/test_pipeline.py
```diff
@@ -0,0 +1,237 @@
...
+            stats_factory=FDStatsFactory(),
+        )
+    )
+
+
+@pytest.fixture
+def boeing_slice():
+    requires_boeing_matrix()
+    M_full = scipy.io.mmread(BOEING_PATH).tocsr()
+    return M_full[:BLOCK_SIZE, :BLOCK_SIZE].tocsr()
```
willow-ahrens, line 110: we should really be able to load the whole matrix.

willow-ahrens (changes requested):
We should be able to easily run the whole matrix.

willow-ahrens:
@sreevickrant This is stale right?

PR #593, opened 2026-08-03
Fix sparse stats construction

kylebd99:
Currently degree_count_scan relies on the NotationInterpreter which means that it crashes on sparse tensors. 

This PR shifts it to using the numba compile path and fixes a loop ordering bug along the way.


kylebd99 (commented):
```
import finch as ft
import numpy as np
from finch import FiberTensor, DenseLevel, SparseListLevel, ElementLevel, NumpyBuffer, NumpyBufferFType, element
from finch.autoschedule.compiler import LogicCompiler
from finch.autoschedule.executor import LogicExecutor
from finch.autoschedule.formatter import DefaultLogicFormatter
from finch.autoschedule.galley_optimize import GalleyLogicalOptimizer
from finch.autoschedule.galley_optimize import GalleyLogicalOptimizer
from finch.autoschedule.loop_ordering import DefaultLoopOrderer
from finch.autoschedule.normalize import LogicNormalizer
from finch.autoschedule.tensor_stats.bound_stats import DCStatsFactory
from finch.autoschedule.tensor_stats.dense_stat import DenseStatsFactory
from finch.autoschedule.tensor_stats.uniform_stats import UniformStatsFactory
from finch.codegen.numba_codegen.numba import NumbaCompiler
from finch.compile.lower import NotationCompiler
from finch.finch_assembly.simplification import AssemblySimplify
from finch.finch_assembly.simplification import AssemblySimplify
from finch.finch_assembly.struct_slots import LowerPackedStructSlots
from finch.finch_logic.nodes import MapJoin, Table, Literal, Field

def test_sparse_array():
    ptr = NumpyBuffer(np.array([0, 1, 3, 4], dtype=np.intp))
    idx = NumpyBuffer(np.array([0, 0, 1, 2], dtype=np.intp))
    data = NumpyBuffer(np.array([1, 1, 2, 1], dtype=np.int64))

    a = FiberTensor(
        DenseLevel(
            SparseListLevel(
                ElementLevel(
                    element(
                        np.int64(0), ft.ftype(np.int64), ft.ftype(np.intp), NumpyBufferFType
                    ),
                    data,
                ),
                np.intp(3),
                ptr,
                idx,
            ),
            np.intp(3),
        )
    )
    b = FiberTensor(
        DenseLevel(
            SparseListLevel(
                ElementLevel(
                    element(
                        np.int64(0), ft.ftype(np.int64), ft.ftype(np.intp), NumpyBufferFType
                    ),
                    data,
                ),
                np.intp(3),
                ptr,
                idx,
            ),
            np.intp(3),
        )
    )
    scheduler = LogicNormalizer(
    LogicExecutor(
        GalleyLogicalOptimizer(
            DefaultLoopOrderer(
                DefaultLogicFormatter(
                    LogicCompiler(
                        NotationCompiler(
                            NumbaCompiler(),
                            ctx_transforms=(
                                LowerPackedStructSlots(),
                                AssemblySimplify(),
                            ),
                        )
                    )
                )
            )
        ),
        stats_factory= DCStatsFactory(),
    )
)
    ft.set_default_scheduler(scheduler)
    la = a
    lb = b
    c = la + lb 
    computed_c = ft.compute(c)
    print(computed_c.to_numpy())

    expr = MapJoin(Literal(ft.algebra.ffuncs._Add()), (Table(Literal(la), (Field('i'), Field('j'))), Table(Literal(lb), (Field('i'), Field('j')))))
    stats = ft.autoschedule.galley.logical_optimizer.logic_to_stats.insert_statistics(UniformStatsFactory() ,expr, {}, False, {})
    print(stats)

test_sparse_array()

```

Example usage file.

kylebd99:
@willow-ahrens this is a small patch PR, but it is blocking sparse galley. 

The alternative route that we could follow is to use FinchLogic for degree_count_stats. The only issue is that this would require a scheduler that promises to never call degree_count_stats in order to avoid a loop.

kylebd99:
Working through some weirdness around Scalars...

kylebd99:
It looks like there were some places in the Assembly lowering processes that expected Literals to be scalars and some places that expected raw values. I've decided go to for raw values everywhere in Assembly and removing all Scalars from FinchAssembly. This got exposed due to the way we wrote the FinchNotation within degree_count_scan.

Should be good now though.

willow-ahrens (changes requested):
The changes to util.py look good, and the changes to simplification.py are something we should eventually fix, but maybe in a separate PR. The scalar.py changes can be removed.

kylebd99:
@willow-ahrens I went a slightly different direction with this one and decided to:
1) move stats computation to FinchLogic
2) specify a "non-recursive-scheduler" which essentially promises to not call a scheduler within it, in order to compute the new FinchLogic program
3) move COMPILE_JULIA to autoschedule.default_schedulers to avoid a circular dependency

src/finch/autoschedule/tensor_stats/util.py
```diff
@@ -5,153 +5,122 @@
...
-_INT64_VECTOR_FTYPE = BufferizedNDArray.from_numpy(np.zeros(1, dtype=np.int64)).ftype
 
+def _scalar(value) -> Table:
+    """A 0-d table holding ``value``, so it broadcasts against any MapJoin arg."""
+    return Table(Literal(BufferizedNDArray.from_numpy(np.asarray(value))), ())
 
-def _int_tuple_ftype(size: int):
-    return ftype(tuple(np.int64(0) for _ in range(size)))
 
+def get_lp_norms(
```
willow-ahrens, line 31: small request: Can this live in bound_stats.py, in the boundstats class?

src/finch/autoschedule/default_schedulers.py
```diff
@@ -121,6 +123,22 @@
...
+    LogicExecutor(
+        DefaultLogicOptimizer(
+            DefaultLoopOrderer(DefaultLogicFormatter(LogicCompiler(FinchJLCompiler())))
+        )
+    )
+)
+
+# This scheduler is used for computing tensor statistics.
+# Crucially, in order to avoid a circular dependency, this scheduler does not
+# cannot rely on a stats factory that itself calls the interface.
```
willow-ahrens, line 139: Can we use FDFormatter? Is the default Formatter already FDFormatter?

willow-ahrens: Is there some way to make the pixi dep for julia backend a default install group for now? I don't think interpret_notation will ever really be fast enough, it's almost a bug if we call it.

willow-ahrens (approved):
This is a cool PR. I've got two questions but overall it's cool to see some of the discussions about non-recursive schedulers, etc, getting materialized.

src/finch/autoschedule/tensor_stats/util.py
kylebd99: It's being used in uniform and VP stats as well, so it seemed like a utility function.

src/finch/autoschedule/default_schedulers.py
kylebd99: I just changed this to make the julia deps part of the core deps. Lmk what you think about this.

kylebd99:
Ah okay, there's a subtle julia bug that I uncovered when I turned on caching for the stats generation. Every call to julia compiles a new function named "main". When I cache that function generation, then something comes through and redefines it. The cache then silently references the new main function. To fix this, we should give the julia functions unique names. Plus, this might help with julia-side caching/avoiding rejitting.

src/finch/autoschedule/default_schedulers.py
willow-ahrens: solid.

PR #598, opened 2026-08-05
Fix docs warnings and enable strict mode

hameerabbasi:


hameerabbasi:
cc @willow-ahrens This should be an easy one.

mtsokol (approved):

PR #599, opened 2026-08-05
Test reorganization

### LINKED ISSUE #600: Python code confuses `lazy` module and function.
In the following line:

https://github.com/finch-tensor/finch-tensor/blob/d3119c7c834d30b7e9ef821bdd2dd071eab31f19/src/finch/interface/eager.py#L16

I believe we intend to import the `finch.interface.lazy` module, not the `finch.interface.lazy.lazy` function. This is evidenced by the fact that we do `lazy.LazyTensor` in a lot of places in that file.

The issue is caused by these lines:

https://github.com/finch-tensor/finch-tensor/blob/d3119c7c834d30b7e9ef821bdd2dd071eab31f19/src/finch/interface/__init__.py#L147-L151

As there's now a lazy module, and a lazy function there. Python tends to pick the latter.

hameerabbasi:
Fixes #600

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
             default
             ${{ matrix.suite.environment }}
       - name: Run tests
-        run: pixi run -e ${{ matrix.suite.environment }} ${{ matrix.suite.task }} --junit-xml=test-${{ matrix.os }}-Env-${{ matrix.suite.environment }}.xml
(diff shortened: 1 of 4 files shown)


hameerabbasi:
cc @willow-ahrens This is ready for a review. I've kept the commit history clean so we can use merge-commit instead of squash, and reviewing is also easier.

hameerabbasi:
Gonna open a new one -- conflicts are hard.

PR #601, opened 2026-08-05
Test Reorganization

hameerabbasi:


hameerabbasi:
Making a merge commit instead of a squash for this PR would be nice.

pyproject.toml
```diff
@@ -131,6 +147,13 @@ default-environment = "benchmark"
 cmd = "python -m mypy"
 default-environment = "test"
 
+[tool.pixi.feature.test.tasks.test-array-api]
```
willow-ahrens, line 150: Is there possibly a way to make the array api tests run with the toplevel pytest interface? Like maybe we forward args to it somehow after we clone?

willow-ahrens (approved):

pyproject.toml
hameerabbasi: We do already, see the deps of the regular test task. This is a dependency for that one, so it runs it before that one as well. `pixi run test` is all that's needed.

PR #602, opened 2026-08-05
Fix the confusion between the `lazy` module and function.

### LINKED ISSUE #600: Python code confuses `lazy` module and function.

hameerabbasi:
Fixes #600


willow-ahrens (changes requested):
For this change, would it be better to rename the "lazy" function to "defer" like in julia? I guess the intention behind naming the whole module "lazy" was that it would be nice to be able to call e.g. finch.lazy.add(A, B) and get a lazy addition of A and B out of it.

hameerabbasi:
Will discuss at the next meeting what the desired approach is here.

PR #605, opened 2026-08-07
Use accelerate on macos

mtsokol:
Apple Silicon: use Accelerate instead of the default OpenBLAS. Accelerate reaches the AMX matrix coprocessor, roughly doubling dense gemm throughput.


hameerabbasi (approved):

hameerabbasi:
Thanks! 🤩

PR #606, opened 2026-08-07
Dynamic Fill Values

### LINKED ISSUE #169: Taking scalars seriously
The `Literal` node in FinchLogic ultimately represents inputs to the program, but doesn't signify that the inputs are actually constant. In fact, the only constant is the type of the value in the node, and we expect subsequent invocations of the compiler to use the same compiled kernel, even though the Literals may actually have different values. We should likely use the `Value` node for this purpose.

This reflects a broader issue in all three Finch codebases, and issues have been raised in the other repos: https://github.com/finch-tensor/Finch.jl/issues/467 and https://github.com/finch-tensor/finch-tensor/issues/25.

At the interface level, it's impossible to tell the difference between `A + 1` and `A + (x:Int)`. We can't unilaterally decide whether scalars are constant or variable.

I believe the resolution proposed in https://github.com/finch-tensor/finch-tensor/issues/25 is sound. We pick a set of scalars (-Inf, -1, 0, 1, Inf) to automatically promote to constants, and another set of scalars to treat as variables, and provide users a way to override the decision either way. When we convert arguments to Finch Logic in interface.jl, we should ensure there is a difference between something assumed to be constant and compiled as part of the program AST itself, and something which is just an expression where we expect the type to be constant.

The only question is whether to introduce `StaticScalar` which puts the value it represents into the ftype, or to create a property called `is_static`, which defines whether values passed into the program should get wrapped in `Value` or `Literal`, or both. When users perform an operation like `A + 1`, we might ask them to do e.g. `A + var(1)` if they want the value to be treated like an input to the program, or `A + const(1)` if they want it to be treated like a constant.

kylebd99:
This PR does a few things:
1) it makes it so that elementwise functions actually use Literals instead of 0-d tensors when given scalars.
- see: lazy.py
2) It formalizes fill values into an AbstractFill class.
3) It introduces DynamicFill's as a subclass in order to denote the fill value being subject to change based on the inputs to the program. This is a way of avoiding recompilation when things change, e.g. when a scalar changes every iteration of some outer loop. 
4) It introduces ConstantScalar for users (or @jit) to denote when a scalar will not change and should be compiled against. This allows for important optimizations like constant folding. 
- see: a whole bunch of files, but mostly autoschedule/executor.py, algebra/fill.py, and tensor/scalar.py.

As an overview on the use of DynamicFill values: 
1) They are introduced by default when numbers are wrapped in Scalar, and they store the ftype of the number. They fail all algebraic property calls (e.g. is_annihilator). 
2) The dynamic fill values are carried into FinchNotation explicitly and without adjustment (e.g. they occur on the rhs of ntn.Declare operators). 
3) The FinchNotation->FinchAssembly step has to compile them carefully by computing the fill value based of the allocated output tensor at runtime.
4) The code to do this is produced via the lower_fill() function which basically just lowers to tns.lvl.lvl....lvl.value. 
5) The LogicExecutor computes the correct fill value for all allocated tensors at runtime based on the input tensors. This is not too complicated, but it requires the optimized logic program to be returned to the LogicExecutor so that we can call infer_fill_value on each output alias. 

Note: The Julia backend always compiles as if DynamicFill is equal to 0 which is a bit iffy. I am not sure if this will cause issues.

https://github.com/finch-tensor/finch-tensor/issues/169


kylebd99:
Small script showing the minimal compilation using dynamic fill values:

```
"""Show what a dynamic fill value buys, and what it costs.

A tensor's fill is the background value of a sparse tensor and the value a fresh
tensor is filled with. Its *ftype* records one more thing: whether a compiled
kernel may specialize on that value.

* `StaticFill` -- the value may be folded into generated code. Every distinct
  fill is a distinct ftype, so every distinct fill gets its own kernel.
* `DynamicFill` -- the value is known but must not be folded. It reaches the
  kernel as an argument at bind time, so one kernel serves every fill of that
  dtype.

Compilation is the expensive step, so the win is kernel reuse: a computation run
over many fill values compiles once instead of once per value. The cost is lost
constant folding, which the last section measures rather than glosses over.

    pixi run -e test python scripts/demo_dynamic_fill.py
    pixi run -e test python scripts/demo_dynamic_fill.py --values 25
    pixi run -e test python scripts/demo_dynamic_fill.py --n 4000000 --reps 50
"""

from __future__ import annotations

import argparse
import time
from collections.abc import Callable, Sequence

import numpy as np

import finch
from finch import ConstantScalar
from finch.algebra import is_dynamic
from finch.autoschedule import (
    DefaultLogicFormatter,
    DefaultLogicOptimizer,
    DefaultLoopOrderer,
    LogicCompiler,
    LogicExecutor,
    LogicNormalizer,
)
from finch.codegen import NumbaCompiler
from finch.compile import NotationCompiler
from finch.finch_assembly import AssemblySimplify, LowerPackedStructSlots
from finch.finch_notation.interpreter import NotationInterpreter


def fresh_scheduler() -> tuple[LogicExecutor, LogicNormalizer]:
    """A scheduler with an empty kernel cache, so counts start from zero."""
    executor = LogicExecutor(
        DefaultLogicOptimizer(
            DefaultLoopOrderer(
                DefaultLogicFormatter(LogicCompiler(NotationInterpreter()))
            )
        ),
        cache=True,
    )
    return executor, LogicNormalizer(executor)


def timed(work: Callable[[], object]) -> float:
    start = time.perf_counter()
    work()
    return time.perf_counter() - start


def rule(title: str) -> None:
    print(f"\n{title}\n{'-' * len(title)}")


# ───────────────────── 1. one kernel, or one per value ──────────────────────


def demo_kernel_reuse(values: Sequence[float]) -> None:
    rule(f"1. Adding a scalar to a vector, for {len(values)} different scalars")
    arr = np.arange(6.0)
    x = finch.asarray(arr)

    def run(wrap: Callable[[float], object], label: str) -> None:
        executor, ctx = fresh_scheduler()
        elapsed = 0.0
        for v in values:

            def once(v=v):
                return finch.compute(finch.lazy(x) + wrap(v), ctx=ctx)

            elapsed += timed(once)
            # The answers must agree, or "fewer kernels" would be meaningless.
            np.testing.assert_allclose(np.asarray(once().to_numpy()), arr + v)
        print(
            f"  {label:34} kernels compiled: {len(executor.cached_kernels):3d}"
            f"   total {elapsed * 1e3:7.1f}ms"
        )

    # A plain Scalar generalizes its fill: `argument_ftype` carries a DynamicFill.
    run(lambda v: finch.Scalar(np.float64(v)), "Scalar (dynamic fill)")
    # A ConstantScalar opts into specialization, so each value is a new ftype.
    run(lambda v: ConstantScalar(np.float64(v)), "ConstantScalar (static fill)")
    print("\n  Same results either way; the dynamic fill just stops the value")
    print("  from becoming part of the kernel's identity.")


# ──────────────── 2. why the *output* fill matters when chaining ─────────────


def demo_chaining(values: Sequence[float]) -> None:
    rule(f"2. Chaining a second computation off the result, for {len(values)} scalars")
    arr = np.arange(6.0)
    x = finch.asarray(arr)

    executor, ctx = fresh_scheduler()
    for v in values:
        step1 = finch.compute(finch.lazy(x) + v, ctx=ctx)
        assert is_dynamic(step1.ftype.fill_value), "expected a dynamic output fill"
        step2 = finch.compute(finch.lazy(step1) * 2.0, ctx=ctx)
        np.testing.assert_allclose(np.asarray(step2.to_numpy()), (arr + v) * 2.0)
    dynamic_kernels = len(executor.cached_kernels)

    # The counterfactual: materialize the intermediate with a *static* fill, as
    # would happen if the executor resolved a dynamic output fill to a concrete
    # value. Its fill then becomes part of the second kernel's identity.
    executor2, ctx2 = fresh_scheduler()
    for v in values:
        step1 = finch.compute(finch.lazy(x) + finch.ConstantScalar(v), ctx=ctx2)
        assert not is_dynamic(step1.ftype.fill_value), "expected a static output fill"
        step2 = finch.compute(finch.lazy(step1) * 2.0, ctx=ctx2)
        np.testing.assert_allclose(np.asarray(step2.to_numpy()), (arr + v) * 2.0)
    static_kernels = len(executor2.cached_kernels)

    print(f"  dynamic output fill (kept)      kernels compiled: {dynamic_kernels:3d}")
    print(f"  output fill pinned to a value   kernels compiled: {static_kernels:3d}")
    print("\n  An output fill that was never determined by the static inputs has")
    print("  no business entering the next kernel's identity. Keeping it dynamic")
    print("  is what stops the recompiles from compounding down a pipeline.")


# ───────────────────────── 3. what it costs ─────────────────────────────────


def numba_scheduler() -> tuple[LogicExecutor, LogicNormalizer]:
    """A cached scheduler that compiles to numba, so folding shows up in code."""
    executor = LogicExecutor(
        DefaultLogicOptimizer(
            DefaultLoopOrderer(
                DefaultLogicFormatter(
                    LogicCompiler(
                        NotationCompiler(
                            NumbaCompiler(),
                            ctx_transforms=(
                                LowerPackedStructSlots(),
                                AssemblySimplify(),
                            ),
                        )
                    )
                )
            )
        ),
        cache=True,
    )
    return executor, LogicNormalizer(executor)



def main(argv: Sequence[str] | None = None) -> int:
    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
    parser.add_argument(
        "--values",
        type=int,
        default=8,
        help="how many distinct fill values to sweep (default: 8)",
    )
    parser.add_argument(
        "--n",
        type=int,
        default=1_000_000,
        help="elements in the steady-state timing of section 3 (default: 1e6)",
    )
    parser.add_argument(
        "--reps",
        type=int,
        default=20,
        help="timed executions per configuration in section 3 (default: 20)",
    )
    args = parser.parse_args(argv)
    values = [float(i) + 1.0 for i in range(args.values)]

    demo_kernel_reuse(values)
    demo_chaining(values)
    return 0


if __name__ == "__main__":
    raise SystemExit(main())

```

willow-ahrens (approved):
Thanks Kyle!

I think that to make things align more with the current runtime/compile-time distinctions, we may want to have .fill_value return an object which is either StaticFill or DynamicFill, and then define .fill_value on the ftype to return the type of the fill object to be returned, StaticFillFtype or DynamicFillFType. This is likely how we'll handle dimensions too, as StaticDim or DynamicDim objects. Still, it feels like material for another PR.

PR #607, opened 2026-08-07
Add IR-agnostic algebraic simplification pass

### LINKED ISSUE #336: Simplify pass for FinchNotation
We already have simplification pass for Finch Assembly - there should be one for Notation as well, with appropriate rules.

### LINKED ISSUE #534: Abstract call simplification interface
We should add an abstract interface for function-call-like IR nodes that allows us to simplify/const-prop/annihilate accross different IRs with the same set of simplifications.

kylebd99:
### Rules

| Rule | Rewrite | Requires |
|---|---|---|
| `fold_literals` | evaluates literal arguments at compile time | see below |
| `annihilate` | `f(…, z, …)` → `z` | `is_annihilator` |
| `canonicalize_associative` | reassociates and floats literals to top if possible| associative|
| `dedup_idempotent` | drops duplicate arguments | idempotent + abelian |
| `drop_identities` | `f(…, e, …)` → `f(…, …)` | `is_identity` + associative |
| `unwrap_singleton` | `f(x)` → `x` | associative |

### Supporting changes in `algebra`

- **`arity` on `FinchOperator`** — a positive integer, or `math.inf` for an operator that accepts any number of arguments. Associativity permits *regrouping*, not a wider call: `logical_and` is associative but has arity 2, so flattening a nest of them builds a term that cannot be typed or evaluated. `flatten_associative` therefore admits a regrouping only when `len(flat) <= arity(op)`, which also lets a ternary associative operator absorb a nested binary call. Declared on all four operator base classes and on each of the 24 operators that subclass `FinchOperator` directly, so the class default is never relied on; each declaration was audited against the operator's `__call__` (or, for the two `*args` forwarders, its `return_type`). Seven associative operators turn out to be fixed-arity and are no longer flattened.
- **`is_identity` / `is_annihilator` answer `False` for tensor arguments.** `Scalar` is a `Tensor`, so `Scalar == 0` inside `_Add.is_identity` was building and running a whole tensor program to answer a question about the algebra. Callers that understand the wrapper still unwrap and ask about the value inside, as `AssemblySimplify` does.

### Wiring

`AssemblySimplify` and `LoopletSimplify` now compose the generic rules with their own:

```python
return simplify_terms(term, [*simplify_rules(), self.simplify])
```
This is important because it should all happen as a fixpoint operation.

### Codegen effect

The `test_asm_sparse_elemwise` regression baseline is updated. The change is an improvement: `add(0, mul(stride, i))` → `mul(stride, i)` removes three additions from the outer loop body, and nested `min(min(a,b),c)` / `max(max(a,b),c)` are flattened.

### Testing

New tests in `src/finch/tests/test_simplification.py`.

🤖  Generated with [Claude Code](https://claude.com/claude-code)
^ Edited thoroughly by Kyle


kylebd99:
Slowdown is due to the new simplify fixpoint, but this seems like a reasonable tradeoff.

src/finch/symbolic/simplification.py
```diff
@@ -0,0 +1,252 @@
...
+        fold_literals,
+        annihilate,
+        canonicalize_associative,
+        dedup_idempotent,
+        drop_identities,
+        unwrap_singleton,
+    ]
+
+
+def simplify(root: T, rules: Sequence[RwCallable] | None = None) -> T:
```
kylebd99: Let each IR handle the rewrite.

src/finch/symbolic/simplification.py
```diff
@@ -0,0 +1,252 @@
...
+`is_annihilator`, and `arity`. Rules which need to know about specific operators,
+or about IR nodes other than calls, belong in that IR's own simplification pass.
+Rules are applied until fixpoint.
+"""
+
+import math
+from collections.abc import Sequence
+from typing import TypeVar
+
+from finch.algebra import (
```
kylebd99, line 13: Change the dep diagram

willow-ahrens (commented):
Comment: let's have each IR own the fixpoint, and the symbolic.py module just defines a standard library of rewrites anyone can use.

src/finch/codegen/mlir_codegen/mlir.py
```diff
@@ -391,8 +393,19 @@ def mlir_function_name(op, arg: FType) -> str:
             raise NotImplementedError(f"{op} has no MLIR representation.")
 
 
+def _operand_type(args: Sequence[Any]) -> Any:
```
willow-ahrens: This seems like we can just apply promote_type directly, does that not work?

src/finch/algebra/algebra.py
```diff
@@ -11,6 +11,7 @@ class FinchOperator(ABC):
     is_associative: bool = False
     is_commutative: bool = False
     is_idempotent: bool = False
+    arity: int | float = 2
```
willow-ahrens, line 14: What do we do with arity?

src/finch/symbolic/simplification.py
```diff
@@ -0,0 +1,224 @@
...
+            if (
+                not math.isinf(arity(op.val))
+                or len(literals) < 2
+                or not is_associative(op.val)
+            ):
+                return None
+            if is_commutative(op.val):
+                rest = [arg for arg in args if not isinstance(arg, LiteralTerm)]
+                return _call_like(node, [_evaluate(op, literals), *rest])
+            start, stop = _run_of_literals(args)
```
willow-ahrens: We don't need _run_of_literals, we can just repeatedly pairwise reduce neighboring literals. Some of these rules are a little complicated for what they are.

src/finch/symbolic/simplification.py
```diff
@@ -0,0 +1,224 @@
+"""
```
willow-ahrens, line 1: Can we try to ask claude to try to simplify the syntax in these functions? I suspect there's simpler ways to write some of these with pattern matching, comprehensions, fixpoint, etc.

willow-ahrens (changes requested):
Overall, this looks great!

willow-ahrens (approved):
I'm approving because I trust whatever cleanups happen next, but do see my previous comments

src/finch/codegen/mlir_codegen/mlir.py
kylebd99: Mostly, this just to grab the element type from scalar/buffer types & make promote_type n-ary. Renamed it to make it clearer though.

src/finch/algebra/algebra.py
kylebd99: We need arity to check whether we can absorb child CallTerm's. NAryFinchOperator may have been sufficient here, but it felt like `arity()` should be a proper algebra function.

PR #609, opened 2026-08-09
Annotated Query Paths + Literals Fix

kylebd99:
This PR does a couple of things.
1) It makes Literals LogicExpressions in FinchLogic. This allows them to be used as arguments to Aggregates and MapJoins without wrapping them in Scalars. In a later PR, this will allow us to use them during simplification passes because they are semantically code not data. It also means that they can be compiled to ntn.Literal nodes.
- see: finch_logic/nodes.py, autoschedule/compiler.py
2) It updates the stats interpreter, insert_stats function, finchlogic interpreter, and finchlogic compiler to handle literals properly. 
- see: stats_interpreter.py, logic_to_stats.py, finchlogic/interpreter.py, and autoscheduler/compile.py 
3) It fixes several bugs in AnnotatedQuery by switching from node-based rewrites on the point_expr to path-based rewrites. The bugs were due to the deletion/replacement of nodes in different parts of the expression tree. This was surfaced by the addition of literals which are more likely to be erroneously matched on. In addition, it moves the handling of "repeat operators" (e.g. \sum_i c where c is a scalar) to be lazier.
- see: autoschedule/galley/logic_optimizer/annotated_query.py

### DIFF
--- a/src/finch/autoschedule/compiler.py
+++ b/src/finch/autoschedule/compiler.py
@@ -54,6 +54,8 @@ class PointwiseContext:
                         tuple(loops[idx] for idx in idxs),
                     )
                 )
+            case lgc.Literal(val):
+                return ntn.Literal(val)
             case lgc.Relabel(arg, idxs):
                 return self(
                     arg,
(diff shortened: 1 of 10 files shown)


willow-ahrens (approved):

willow-ahrens:
This looks good, I only checked the finchlogic and finchNotation stuff, annotatedquery is a little outside my wheelhouse

PR #613, opened 2026-08-11
loop order big fix (output outerproducts)

Ruchika-2003:


sreevickrant:
@kylebd99 need your help with codspeed, I'm not sure if I rolled back your bug fixes from earlier.

src/finch/autoschedule/loop_ordering.py
```diff
@@ -38,9 +39,29 @@ def concordize(
 ) -> LogicStatement:
     needed_swizzles: dict[Alias, dict[tuple[int, ...], Alias]] = {}
     namespace = Namespace(root)
+    field_orders: dict[Alias, tuple[Field, ...]] = {}
```
kylebd99: Remove this code

willow-ahrens: @kylebd99, if it's clear which code to remove, I think it would be fastest if you took this PR over the finish line. Would that be alright with you?

PR #614, opened 2026-08-11
Set loop order bug fix

HoangNguyen0309:
set_loop_order made loop order an output field order so in concordize Table case rule_0 disagreed with the order and inserted a transpose 

fix: set_loop_order now wraps each reordered query in an outer Reorder so the query’s original output index order is preserved after the inner loop order is changed.


kylebd99 (approved):
I went ahead and made a couple of small changes for things that I had missed earlier. In particular, I removed the broken case from rule_0 of concordize. I made an abstract class for loop ordering and made both loop orderers inherit from it instead of having Greedy inherit from Default.

PR #618, opened 2026-08-13
Automatically Insert ConstantScalar in Lazy

kylebd99:
This PR adds automatic ConstantScalar conversion to lazy for any raw numeric inputs that might be useful for optimization. Specifically, it defines the notion of "specializable" values (0, 1, inf, -inf). This is used in defer() to determine whether a scalar should be a ConstantScalar or Scalar.

In addition, it uses this notion in apply_fill to determine the correct output fill value from a program. We make output tensors StaticFill if their fill can be statically determined AND it is specializable. In all other cases, we make the output DynamicFill. This avoids recompilation when a tensor's fill is updated to a new value every loop iteration.

Lastly, it contains now unrelated code to implement "with_fill" on tensors. This was needed in an earlier iteration, but it still seems useful. So, I decided to keep it. It just allows us to change the fill value of a tensor after its construction.


JoelMathewC (approved):
Looks good to me. Rebasing with main should fix CI (Willow fixed them in #645)

src/finch/algebra/algebra.py
```diff
@@ -76,6 +77,20 @@ def is_annihilator(op: FinchOperator, val: Any) -> bool:
     return ok and op.is_annihilator(value)
 
 
+SPECIALIZABLE_VALUES = (0, 1, math.inf, -math.inf)
```
JoelMathewC, line 80: What about `float('inf')` or `xp.inf`?

JoelMathewC: Okay I think this is still fine

```
>>> import math
>>> import numpy as np
>>>
>>> SPECIALIZABLE_VALUES = (0, 1, math.inf, -math.inf)
>>> float('inf') in SPECIALIZABLE_VALUES
True
>>> np.inf in SPECIALIZABLE_VALUES
True
```

src/finch/interface/fuse.py
```diff
@@ -123,9 +135,12 @@ def compute(arg, ctx=None):
                 len(res[lazy_idx].shape) == 0
             ):  # if the result is a scalar, extract the value and turn it into a
                 # finch `Scalar`
-                outputs[out_idx] = asarray(res[lazy_idx][()], device=device)
+                out = asarray(res[lazy_idx][()], device=device)
             else:
-                outputs[out_idx] = asarray(res[lazy_idx], device=device)
+                out = asarray(res[lazy_idx], device=device)
+            if not is_specializable_value(out.fill_value):
```
willow-ahrens: do this check in defer, rather than in fused

willow-ahrens (changes requested):
Slightly dissapointing request from willow

kylebd99:
I think this PR should be good to go now. Another minor change, ConstantScalar still gets input as a Table node, but it gets immediately unwrapped into a bare literal in normalize.py. This simplified the interface logic a bit and felt more compiler-ey.

willow-ahrens (approved):
Awesome! It's great to see scalars getting proper treatment finally.

PR #626, opened 2026-08-19
Stricter type checking

hameerabbasi:


willow-ahrens:
Because this is gonna affect everyone's style for writing code, let's ask what people think at the next meeting

hameerabbasi:
One of the reasons I marked it as draft -- besides the fact that it isn't finished. 😅

src/finch/algebra/ftypes.py
```diff
@@ -961,7 +967,34 @@ def isdtype(dtype, kind):
...
+
+@overload
+def ftype(x: FT) -> FT: ...
+
+
+@overload
+def ftype(x: Any) -> FType: ...
+
+
+def ftype(x: Any) -> FType:
```
willow-ahrens, line 1005: I find the overloads here to be a little strange just because this was a method meant to be overloaded later. can the new type checker infer return types if we overload NewFTypeableClass.ftype? Would it make sense to call the ftype method directly when we want the type to be inferrable?

src/finch/algebra/ftypes.py
```diff
@@ -1038,3 +1071,25 @@ def ftype(x) -> FType:
...
     raise NotImplementedError
+
+
+@overload
+def np_dtype(dtype: None) -> None: ...
+@overload
+def np_dtype(dtype: type[np.generic] | np.dtype | FType) -> np.dtype: ...
+
+
+def np_dtype(dtype) -> np.dtype | None:
```
willow-ahrens: I'm in support of adding a method like this, but I'd like to make it so that the ftype input is constrained to be something of type FDType. It's a sign of code smell when we need to define methods that work on both dtype and ftype

src/finch/finch_logic/nodes.py
```diff
@@ -165,7 +168,8 @@ def children(self) -> list[LogicNode]:  # type: ignore[override]
         ...
 
 
-T = TypeVar("T")
+T = TypeVar("T", bound=LogicNode)
+TBind = TypeVar("TBind")
```
willow-ahrens: was this typevar necessary? I thought that the old types were clearer here.

src/finch/interface/lazy.py
```diff
@@ -1995,18 +2038,15 @@ def diagonal(x, /, *, offset: int = 0) -> LazyTensor:
...
-        LazyTensor,
-        defer(
-            np.ascontiguousarray(
-                np.diagonal(
-                    compute(x).to_numpy(),
-                    offset=offset,
-                    axis1=-2,
-                    axis2=-1,
-                )
+    return defer(
```
willow-ahrens, line 2035: This looks cleaner, yay!

src/finch/tensor/level/dense_level.py
```diff
@@ -249,9 +248,7 @@ def stride(self) -> np.integer:
 
     @property
     def ftype(self) -> DenseLevelFType:
-        # mypy does not understand that dataclasses generate __hash__ and __eq__
-        # https://github.com/python/mypy/issues/19799
-        return DenseLevelFType(self.lvl.ftype, ftype(self.dimension))  # type: ignore[abstract]
+        return DenseLevelFType(self.lvl.ftype, ftype(self.dimension))
```
willow-ahrens, line 251: Nice.

src/finch/tensor/level/level.py
```diff
@@ -228,3 +229,30 @@ def buffer_factory(self):
...
+class MultiDimensionLevel(Level):
+    lvl: Level
+
+    def with_fill(self, fill_value: AbstractFill) -> Self:
+        """Rebuild this level with the leaf fill value replaced."""
+        return replace(self, lvl=self.lvl.with_fill(fill_value))  # type: ignore[type-var, attr-defined]
+
+
+class SingleDimensionLevelFType(LevelFType):
+    dimension_type: ftypes.FDTypeInteger = ftypes.intp
```
willow-ahrens, line 254: It's not always the case that levels will have intp dimensions. This is something I've had to push for throughout the development of the repo because it's tempting to simplify, but I think it will benefit us later to have kept it on the radar from the beginning.

src/finch/tests/test_assembly_parser.py
```diff
@@ -86,23 +99,23 @@ def test_for_loop():
...
+                        Literal(val=np.intp(1)),
                     ),
                 ),
             ),
             Call(
                 op=Literal(val=np.resize),
                 args=(
-                    Variable(name="lvl_idx", type=np.ndarray),
-                    Variable(name="qos_stop", type=int),
+                    Variable(name="lvl_idx", type=ndarray_ftype),
```
willow-ahrens, line 117: oh, I think the type was wrong here to start, this should have been a buffer. I know we're in tests now though

willow-ahrens (changes requested):
Overall, just a few small tweaks requested from me, mostly about resolving things in different ways.

src/finch/tensor/level/level.py
willow-ahrens: Ah, I see that this is just a default.

hameerabbasi: Indeed, it's just a default.

kylebd99: Agreed, we're going to want int16/int8/etc

src/finch/finch_logic/nodes.py
hameerabbasi: 1. Generics make nicer typing overall possible; i.e., this says that `T` must be a *specific* `LogicNode` type, not just a mishmash of any `LogicNode` types.
2. `TBind` is needed as the bindings have *some* type, but the bound isn't the same.
3. There were type failures without the type vars; we needed to add a LOT of asserts. Better typing is always a cleaner solution than asserting.

src/finch/algebra/ftypes.py
hameerabbasi: It can, but that needs more generics, type-vars and overloads.

src/finch/algebra/ftypes.py
hameerabbasi: We can split this into multiple functions; if that helps. I'm not 100% sure we can though.

src/finch/algebra/ftypes.py
kylebd99: In general, I prefer ftype as a member function of an object. However, that may be a big change throughout the code base.  I don't think they the type checker will be able to follow the type logic if we just call .ftype within ftype().

In addition, it may be worth following the numpy convention of a standalone dtype/ftype function for scalar types at least.

src/finch/tests/test_assembly_parser.py
hameerabbasi: Yep, a lot of the tests were doing funky typing stuff.

src/finch/finch_logic/nodes.py
kylebd99: I'm a bit unclear on how TBind works here as well (partly because I don't entirely understand how TypeVar interacts with the typechecker). There's a trickiness to how dimmap and valmap are defined, and we may just want to use (a tuple of) the Any type for the val type of their bindings dict.

src/finch/algebra/ftypes.py
hameerabbasi: I opened #666 for this. I can address it in a future PR.

src/finch/finch_logic/nodes.py
hameerabbasi: `TBind` says "it's a specific type, but just one type". `Any` can be a mix of types.

hameerabbasi: It also "turns off" parts of the checking which is why I'm rather unwilling to accept it.

willow-ahrens: The previous types were much clearer though, they used parametrically typed dictionaries.

hameerabbasi: Okay; there are two separate issues; One is the bound for `T`; that's necessary to get things working. The TBind one is nicer for type checking in general; and avoiding bugs.

willow-ahrens (approved):

PR #631, opened 2026-08-24
fix codegen for elementwise ops

sreevickrant:


willow-ahrens (approved):
@sreevickrant Notice how I have made the code much simpler, rather than adding additional cases. AI has a tendency to grow the codebase, but if we can, let's use it to make things clearer and more concise. A few of the simplifications i made:
1. don't you think ops_to_ignore was confusing? That's kindof a weird name for a special_case lowering path for tuples
2. there's a function called _ops_for? that just resolves names from ffunc? let's just write ffunc directly as keys in the dict
3. If we're going to write infix ops for & and |, it would be prettier to just have a list of infix ops for +, -, etc.

Overall, I think the code is clearer than when we started now.

PR #634, opened 2026-08-28
FusedAlias implementation

### LINKED ISSUE #633: RFC for FusedAlias
# RFC for FusedAlias Node

## Summary

We plan to introduce a new Finch Logic node, `FusedAlias(alias: Alias,n: int)` for representing a workspace, reusing existing Alias node with n representing the fusion degree. Once the loop order is known wherever the alias is used, the outermost n dimensions of that order are fused. This is mainly to create the grounds for implementing Gustavsons algorithm, and allowing consumers share one computed intermediate without forcing it to be materialized as an individual output tensor.

## Motivation 
Using only SparseHashLevels directly for each level to accumulate the output ends up forcing the structure to start fresh along with hashing and regrowing the entire size of the output instead of a small workspace and copied into sequentially built output. We currently do not have a general way to express a workspace that we can reuse across consumers using Finch Logic IR, hence we plan to introduce FusedAlias.

### Benefits of having a workspace : 

- It minimizes scatter by avoiding out-of-order writes into the output structure.
- It introduces the benefit of cache hierarchy by reusing the same workspace throughout several consumers 

## Mechanism 
`FusedAlias(alias: Alias,n : int)`
Is a thin wrapper around `Alias` along with an integer $n$ which is free : 0<= n <= total_dims.  Where n=0 is a pure scalar that resets every iteration and n=total_dims is fully materialized which loops through regularly.

Fused dimension are always the outermost loops, therefore $n$ represents the first $n$ dimensions that would be fused out of all the dimensions that are going to be looped over.

## Rules for using FusedAlias 

- FusedAlias can never appear as input to the program or output from the program.
- Bindings should not contain any references to the FusedAlias.
- It can be used in the lhs and rhs of a Query. As a FusedAlias in lhs and as Table with FusedAlias in the rhs the same way in which Alias is read.

FusedAlias is meant to be a scratch space local to the computation, hence it should not exist outside the scope of the loops in which it is used. 

## Implementation 

- Step 1 : Adding it as a node to Finch Logic 
- Step 2 : Making sure it can be interpreted just like Alias 
- Step 3 : Making sure it compiles to a proper FinchNotation 

## Worked Examples 

1. Scalar workspace 
```text
B_1[i,j] = A[i,j]*1
B_2[i,j] = A[i,j]*2
B_2[i,j] = A[i,j]*3
```
Using workspace :
FusedAlias(HardAlias("C"),2) = A[i,j]

Lowers to : 
```text
for i :
      for j:
            C.=0
            C = A[i,j]
            B_1[i,j] = C*1
            B_2[i,j] = C*2
            B_3[i,j] = C*3
```

2. Vector Workspace : Gustavsons 
The goal is to compute C[I,j] = sum over k A[i,k] * B[k,j] without materializing a dense output. For each row `i` we will accumulate the contribution from A[i,k] into a scratch workspace and then assign it to the output.

The scratch workspace here is defined over `(i,j)` same as the output `C[i,j]` as  `FusedAlias(HardAlias("scratch"), 1)`.
Here i is the fused index, which leaves j to be materialized.

Lowering:
```text
for i :
      scratch .= 0
      for k in non-zeros of row_i of A
            for j in non-zeros of row_k of B:
                  scratch[j] += A[i,k]*B[k,j]
      for j:
            C[i,j] = scratch[j]
  
```

## Blast Radius
```text
src/finch/autoschedule/compiler.py
src/finch/autoschedule/stages.py
src/finch/finch_logic/__init__.py
src/finch/finch_logic/interpreter.py
src/finch/finch_logic/nodes.py
```

# Implementation using Loop node in Finch Logic
We will continue using `FusedAlias(HardAlias, int)` for referring to a workspace for fusion, but the way we work through the backend is by introducing a Loop node in Finch Logic. Currently the compiler simply processes one Query at a time and thus we have no way to see if sibling queries in a Plan are related and can be fused with the producer consumer logic. 

The compiler gets the loop order passed down by the loop orderer for each query unaware of the effect of sibling queries on each other, which can help us decide if any of the loops can be fused for our purpose. So we should have a way in which the loop orderer explicitly indicates what dimension of certain portions of queries belonging to a plan can be fused together in specific order.

## Idea :
Add `Loop(idx: Field, body : Plan)` as a new FinchLogic node. The interface doesn't deal with this node at all, it is solely meant to be used to enable loop fusion in our pipeline. It will get produced by a compiler pass and will be consumed in the loop ordering or lowering pipeline. We should implement the formation of the Loop node before deciding the output order so fusion can directly control the loops produced making sure the fusion is taken into account. 

`Loop(idx: Field, body : Plan)` is basically where `idx` represents the dimensions to be fused, and `Plan` is what we scan over to check which `Queries` can be potentially fused.

## Logic : 
Building the Loop node involves scanning over the entire Plan and,
1. Grouping : We perform connected component analysis over Queries, where two Queries would be marked connected if they share a `FusedAlias`.
2. Topological Sorting of connected components : We take the connected components and topologically sort them so that the producer consumer order isn't violated while performing fusion.
3. Takes into account the dependency between queries as a whole not just ones that shares `FusedAlias` so that we use this information to order the queries properly and not break the order in which they are performed.
4. Rewriting each component into Plan wrapping relevant `Loop(idx, Plan(sorted_queries))` 

5. Now that we have Queries grouped into specific order we can go ahead and have loop orderer decide the order for looping over remaining dimensions of the queries like the loop_orderer did previously, basically we pin the `idx` dimensions to be fused such that it isn't changed by the loop orderer. 

6. Now that we have `idx` specified and loop order defined for each queries we can have a separate block in `NotationContext` in the compiler looping over this such that we have `idx` outer loops and then inside those loops the queries declare workspace and have the loops generated. 

This ensures the outer loops are fused for sure as the `Loop` node had specified and inner loops loop over the remaining dimension and produce relevant Notation block. 

### Worked Example : 

Ex 1 :

```
Plan(
                          Query(FusedAlias(w, 1), Mapjoin(*,Table(A, i, j), Table(B, i, j))),
			Query(FusedAlias(z, 1),(Mapjoin(*,Table(A, i, j), Table(B, i, j)), j, i)),
			Query(HardAlias(C), Mapjoin(*, Table(FusedAlias(w, 1), i,j) , 2))
                          Query(HardAlias(D), Mapjoin(*, Table(FusedAlias(z, 1), i, j), 2))
)
```
After introducing Loop node would be 
```
Plan(Loop(i,
Plan(
			Query(FusedAlias(w, 1), Mapjoin(*,Table(A, i, j), Table(B, i, j))),
			Query(FusedAlias(z, 1), Reorder(Mapjoin(*,Table(A, i, j), Table(B, i, j)),
			Query(HardAlias(C), Mapjoin(*, Table(FusedAlias(w, 1), i,j) , 2))
Query(HardAlias(D), Mapjoin(*, Table(FusedAlias(z, 1), i, j), 2))
))
```

Which would be lowered to : 
```
For i
	For j
		W[j] = A[i, j] * B[i, j]
	For j
		Z[j] = A[i, j] * B[i, j]
	For j
		C[i, j] = W[j]
	For j
		D[i, j] = Z[j]
```

Ex 2 : 
```
Plan(
                          Query(FusedAlias(w, 1), Mapjoin(*,Table(A, i, j), Table(B, i, j))),
			Query(FusedAlias(z, 1),Reorder((Mapjoin(*,Table(A, i, j), Table(B, i, j)), j, i)),
			Query(HardAlias(C), Mapjoin(*, Table(FusedAlias(w, 1), i,j) , 2))
                          Query(HardAlias(D), Reorder(Mapjoin(*, Table(FusedAlias(z, 1), i, j), 2,(i,j))))
)
```
After introducing Loop node would be 
```
Plan(Loop(i, Plan(  Query(FusedAlias(w, 1), Mapjoin(*,Table(A, i, j), Table(B, i, j))), 
Query(HardAlias(C), Mapjoin(*, Table(FusedAlias(w, 1), i,j) , 2))),
Loop(j, Plan(Query(FusedAlias(z, 1),Reorder((Mapjoin(*,Table(A, i, j), Table(B, i, j)), j, i)),,
Query(HardAlias(D), Reorder(Mapjoin(*, Table(FusedAlias(z, 1), i, j), 2,(i,j)))))
```

Lowers to : 
```
For i
	For j
		W[j] = A[i, j] * B[i, j]
	For j
		C[i, j] = W[j]

For j :
        For i:
              Z[i] = A[j, i] * B[j,i]
        For i:
		D[j, i] = Z[i]
```

Ex 3 : 
Gustavsons : 

Plan((Query(

Ruchika-2003:
https://github.com/finch-tensor/finch-tensor/issues/633


src/finch/finch_logic/nodes.py
```diff
@@ -430,7 +452,7 @@ class Table(LogicTree, LogicExpression):
         idxs: The fields indexing the tensor.
     """
 
-    tns: Literal | Value | Alias
+    tns: Literal | Value | Alias | FusedAlias
```
willow-ahrens, line 460: FusedAlias would be redundant here.

src/finch/finch_logic/nodes.py
```diff
@@ -679,7 +704,7 @@ class Query(LogicTree, LogicStatement):
         rhs: The right-hand side to evaluate.
     """
 
-    lhs: Alias
+    lhs: Alias | FusedAlias
```
willow-ahrens, line 707: I think we agreed alias should be a superclass of FusedAlias and HardAlias.

src/finch/autoschedule/compiler.py
```diff
@@ -261,24 +271,30 @@ def __call__(self, prgm: lgc.LogicStatement) -> ntn.NotationStatement:
...
-            case lgc.Query(lhs, lgc.Reorder(lgc.Table(lgc.Alias(), _) as arg, idxs_2)):
-                body = self._lower_query_of_reorder(lhs, ffuncs.overwrite, arg, idxs_2)
-                match self.bindings[lhs].fill_value:
+            case lgc.Query(
+                lhs,
+                lgc.Reorder(
+                    lgc.Table(lgc.Alias() | lgc.FusedAlias(), _) as arg, idxs_2
+                ),
+            ):
+                key = lhs.alias if isinstance(lhs, FusedAlias) else lhs
```
willow-ahrens, line 280: This is almost certainly wrong unless we have a fuse node in this PR, since we aren't looping over fused dimension properly.

willow-ahrens (changes requested):
This PR also needs Fuse, and I'd like to see the PR that implements Table on the lhs first.

PR #642, opened 2026-09-01
Hash consing 2

### LINKED ISSUE #49: hash consing
Eventually, we want https://en.wikipedia.org/wiki/Hash_consing, the node constructor can return unique instances of the node so that eq and hash can be implemented with pointer equality.

For now, we may want to consider fixing the current implementation of __hash__, as setting `eq=true frozen=true` is overriding it.

At first, I thought we should add the line __hash__ = Term.__hash__ to each LogicNode class definition, as described in this StackOverflow: https://stackoverflow.com/questions/53518981/inheritance-hash-sets-to-none-in-a-subclass.

However, I think I would rather just set eq=false. from https://docs.python.org/3/library/dataclasses.html:

>    If eq is false, hash() will be left untouched meaning the hash() method of the superclass will be used (if the superclass is [object](https://docs.python.org/3/library/functions.html#object), this means it will fall back to id-based hashing).

At a minimum: I think we should move our optimized __hash__ and __eq__ to TermTree, set eq=false for TermTree subclasses, and leave eq=true for TermTree and Leaf nodes like immediate, deferred, field, and alias. 

https://arxiv.org/abs/2509.20534

weakvaluedict would be the right place to start: https://docs.python.org/3/library/weakref.html

MaksimKrylykov:


src/finch/tests/test_assembly_parser.py
```diff
@@ -42,21 +42,27 @@ def test_for_loop():
                     Variable(name="lvl_ptr", type=np.ndarray),
                     Call(
                         op=Literal(val=ffuncs.add),
-                        args=(Variable(name="pos_stop", type=int), Literal(val=1)),
+                        args=(
+                            Variable(name="pos_stop", type=int),
+                            Literal(val=np.intp(1)),
```
willow-ahrens, line 47: why did normal python int not work?

willow-ahrens (changes requested):
Let's try to find performance improvements, try compiling matmul with all of the default schedulers and see if we can find either a runtime or memory improvement

PR #645, opened 2026-09-04
Add bellman-ford benchmark and keyword support for fused finch

JoelMathewC:
This PR adds bellman-ford as a benchmark program for the julia backend so that we can track improvements and regressions associated with finch kernel calls within loops. Additionally, the PR extends the fused framework to support keyword arguments within function calls (CC - @kylebd99)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Run Julia benchmarks
         uses: CodSpeedHQ/action@v4
         with:
-          mode: simulation,memory
+          mode: walltime
           run: |
             pixi run benchmark-julia -vvv -m "not slow"
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -21,6 +21,27 @@ jobs:
           run: |
             pixi run benchmark -vvv -m "not slow"
 
+  benchmark-julia:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # required for actions/checkout
+      id-token: write # required for OIDC authentication with CodSpeed
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+      - uses: prefix-dev/setup-pixi@v0.10.0
+        with:
+          pixi-version: v0.71.3
+          environments: benchmark-julia
+      - name: Run Julia benchmarks
+        uses: CodSpeedHQ/action@v4
+        with:
+          mode: simulation,memory
+          run: |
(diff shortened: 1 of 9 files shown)


JoelMathewC:
Is the `mlir on ubuntu` and `docs` known failures?

kylebd99 (approved):
This looks good to me! Bellman-ford is a great, minimal benchmark for iterative expressions. Does it currently avoid recompilation correctly?

kylebd99:
The MLIR test failure appears unrelated, so I wouldn't block this based on that.

JoelMathewC:
> Does it currently avoid recompilation correctly?

When I was testing it earlier this week, it wasn't avoiding recompilation. I need to take a proper look again to verify if I was doing the right thing. Willow suggested to get the benchmark merged out as a start and then subsequent PRs should show improvements on codspeed.

kylebd99:
Sounds good, happy to merge and improve perf later.

willow-ahrens:
Yeah, Joel found some good stuff to fix.

willow-ahrens:
both failures look like totally unrelated fetch failures

willow-ahrens:
@yusheng036 any idea the mlir failure?

willow-ahrens (approved):

willow-ahrens:
I'm gonna try to fix, but we can also merge I'm not worried about these fails

willow-ahrens:
@JoelMathewC Please make a new PR to add the fixes you found. You don't need separate PRs for each fix,

JoelMathewC:
@willow-ahrens I actually don't see the test popping up on codspeed, I think I have to add some more configuration for it to be recognized

kylebd99:
@JoelMathewC Are the fixes related to Static/Dynamic fills? If so, I'm cleaning up my jit constants pr right now, and you could pick it up from there and/or use it for reference. If you're already too deep in, then don't worry about it though.

JoelMathewC:
I don't think they are related to Static/Dynamic fills. The fixes I had were mostly around marshalling data between julia and python and also caching at the logic executor (although I need to verify if this was because I was using it incorrectly). Then I had a seperate fix for `.where` lowering in the julia backend.

kylebd99:
Perfect, I'll do my fix separately then.

JoelMathewC:
Looks like julia tests were being completely ignored. Added in the config for that to reflect on codspeed.

willow-ahrens:
@JoelMathewC are we good to merge now?

JoelMathewC:
Yupp we're good to merge now. I had to switch to walltime to have the julia tests complete in reasonable time. But codspeed reports the following

image

PR #646, opened 2026-09-04
Improving memory management within the Julia Backend

JoelMathewC:
This PR introduces a number of improvements for the Julia backend.
1. **Logic Executor Caching:** Enable logic executor caching for `COMPILE_JULIA`. I verified that this results is cache hits and helps avoid redundant recompilation.
2. **Avoiding memory transfers between Julia <-> Python**: The proposed change enforces an ownership model where julia owns all memory. This implies the following
    - `JuliaBufferContext`: In the first invocation of each kernel, memory is copied over from python to julia. After the first invocation the buffer context keeps these buffers alive. All future kernel invocations pick the corresponding julia buffers from cache and reuse them without any copying. 
    - `Index Buffers`: This PR introduces a `MinusOneBuffer` which bridges the 0/1 indexing mismatch between python and julia. This wrapper allows julia to maintain ownership.
    - Since julia does own the memory it can continue to call functions that modify memory allocation like `resize!` / `push!`. However, python on the other hand can only mutate values in memory but cannot modify the memory allocation itself without performing a copy.
4. **Functional Improvements**: 
    - introduces operator mappings for vanilla non-reduction min/max functions, first_arg and compare operators.
    - dynamic fill handling for jl backend
    - support fill tensor lowering as dense-dense


willow-ahrens:
This PR doesn't seem to improve performance on the walltime runner

JoelMathewC:
The PR is still a work in progress, I should be done soon

JoelMathewC:
I've been racking my brain trying to figure out why I'm unable to reproduce the performance improvement anymore, but I finally figured it out. The reason the performance overhead is not prominent here is because the benchmark converts the scipy matrix to a finch tensor using `ft.asarray` (materializing only non-zeros and hence relatively small) whereas the test I was doing last week was passing in a dense numpy tensor. That said the fix here is still a valid concern because `_ndarray_to_jl_tensor` which is called on every kernel invocation performs a tensor copy instead of passing a pointer. Here are the benchmarks from my numpy dense (4kx4k matrix) running for varying number of iterations

| Num Iterations | Tensor Copy (in ms) | Pointer Passing (in ms) |
|-|-|-|
| 10 | 271 | 24 |
| 20 | 320 | 47 |
| 30 | 392 | 70 |
| 40 |  539 | 93 |

PS - We should ignore the codspeed result above, that performance improvement is just because we got allocated a different CPU

src/finch/compile_jl/interop.py
willow-ahrens: Joel, this looks like it makes a copy, or at the very least should become a MinusOneBuffer rather than an eager update.

willow-ahrens (changes requested):
Should we also avoid the copy on the way back? I think that to fully get this right, we would need to be careful about who owns which arrays.

willow-ahrens:
I've made a PR that sketches an approach: https://github.com/finch-tensor/finch-tensor/pull/647. It's a codex PR, so we should vet it a bit or draft our own, but it's roughly what I think we need to do.

JoelMathewC:
@willow-ahrens this ready for another look now. I inverted the ownership model so that julia has more flexibility in managing memory. The description explain the new flow. I used the `MinusOneBuffer` to deal with index arrays.

On a side note, I had to set `single_writer=True` in the smart formatter so that it agrees with the default on SparseHashLevelFType. Without this change there is a conflict between the julia kernel prototype and the runtime input passed in

willow-ahrens (approved):

Issue #288, opened 2026-01-13
Implement Global Value Numbering


idkwhttsay:
number statements instead of variables
then, implement GVN

paper on GVN: 
- https://www.cs.tufts.edu/~nr/cs257/archive/keith-cooper/value-numbering.pdf

Bril lectures:
- https://www.cs.cornell.edu/courses/cs6120/2025sp/blog/global-value-numbering/
- https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/global-value-numbering/

Bril LVN and SSA implementation (https://github.com/sampsyo/bril/tree/main/examples):
- examples/lvn.py
- exampels/to_ssa.py

Some random compiler's GVN implementation:
- https://deepwiki.com/yizhiluobo622/shade_that_compiler/5.3-global-value-numbering-and-code-motion
- https://github.com/yizhiluobo622/shade_that_compiler/blob/master/src/toolkit/gvn.rs

PR #653, opened 2026-09-09
Add local value-numbering analysis for Finch Logic

tarundevi:
Track query definitions and equivalent values in sequential Logic plans, with conservative invalidation for tensor writes and unknown effects. Add focused regression and frontend tests. Related to #288; no CSE rewrites or global control-flow analysis yet.

### DIFF
--- a/src/finch/finch_logic/__init__.py
+++ b/src/finch/finch_logic/__init__.py
@@ -4,6 +4,11 @@ from .interpreter import (
     MockLogicLibrary,
     MockLogicLoader,
 )
+from .local_value_numbering import (
+    LogicDefinition,
+    LogicLocalValueNumberingResult,
+    logic_local_value_numbering,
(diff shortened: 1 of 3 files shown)


willow-ahrens (commented):
I wonder if the way that you intern objects might be made more simple if you were able to use a small hashconsed programming language, like:
class NumberedValue(HashCons):
   def number(self):
      return self.id

class LiteralValue(NumberedValue):
   def __init__(self, val):
      self.val = val
   def __eq__(self, other):
      return self.val == other.val and type(self.val) == type(other.val)

class DerivedValue(NumberedValue):
   def __init__(self, op, args):
      self.op = op
      self.args = args

   def __eq__(self, other):
      return self.op == other.op and self.args == other.args(edited)
[11:14 AM]^would this work?

PR #654, opened 2026-09-09
Buffer reuse in the julia backend

JoelMathewC:
This PR supplements the work in #646. It improves performance in the julia backend through the following
1. **Cross-kernel buffer reuse**: Previously buffers were reused only for a single kernel, this PR allows buffers to be reused across multiple kernels by maintaining a process-wide julia buffer context. To verify the performance I compiled the following program which generates two kernels within the loop and the performance is shown in the table below
   ```python
    def bellman_ford_activeset(G, n, max_iter, xp, source=0):
        D = xp.full(n, math.inf)
        D = xp.where(xp.equal(xp.arange(n), source), 0.0, D)
        active = xp.zeros(n, dtype=bool)
        active = xp.where(xp.equal(xp.arange(n), source), True, active)
        t = 0
        while t < max_iter and xp.any(active):
            D_active = xp.where(active, D, math.inf)
            new_dists = xp.min(xp.expand_dims(D_active, 1) + G, axis=0)
            new_active = new_dists < D
            D = xp.minimum(D, new_dists)
            active = new_active
            t += 1
        return D
   ```

    | Baseline | With Cross kernel buffer reuse |
    |-|-|
    | 1464.48ms | 14.92ms |

2. Misc Changes
    - Got rid of `CompiledJLKernel` since it was a wrapper over `FinchJLKernel` without any distinctive features.

### DIFF
--- a/src/finch/autoschedule/smart_formatter.py
+++ b/src/finch/autoschedule/smart_formatter.py
@@ -209,10 +209,10 @@ class FDFormatter(SmartFormatter):
                 required_fields.issubset(dense_fields)
                 for dense_fields in stats.dense_props
             )
-            # Temporary Julia-backend experiment: materialize compiler-created
-            # tensors densely rather than selecting SparseHash for dimensions
-            # whose density cannot be proven by the FD analysis.
-            lvl = dense(lvl, shape_type[dim])
+            if is_dense:
+                lvl = dense(lvl, shape_type[dim])
+            else:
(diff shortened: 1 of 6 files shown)


willow-ahrens:
Why are buffers so expensive?

JoelMathewC:
Closing this PR out because some of the work here needs information beyond kernel boundaries. I'm going to have to raise those optimizations into the GLAM layer. I'll raise a new PR with the fixes that can stay within julia.

PR #655, opened 2026-09-11
Wma/fast stats

willow-ahrens:
make stats go fast!

Adds support for julia masks, and links them to patterntensors in python. Just don't iterate the masks out of order and everything will be okay. Also don't try to compute stats on masks if you can help it haha.

https://github.com/finch-tensor/Finch.jl/pull/816 has the new masks needed. This was a fun PR. We can probably increase performance on some of the other stats, but this is a good start.

This PR gives up on blockstats. Somehow, with the current infra, individual kernel call overhead is a little high to be running 100 separate kernels just to compute stats.


kylebd99:
When we move over to the python backend, will these masks need "level_lower_unfurl" implementations?

willow-ahrens:
That's exactly right, they will

Ruchika-2003 (approved):

willow-ahrens:
pre-commit.ci autofix

src/finch/autoschedule/tensor_stats/sampling_stats.py
```diff
@@ -289,25 +289,23 @@ def __init__(self, sample_prob: float = 0.5, estimator: str = "uj1"):
...
 
-    def _get_mask(self, field: Field, size: int) -> np.ndarray:
-        "Returns mask for dimension that already exists or creates a new one"
+    def _get_mask(self, field: Field, size: int) -> RandomMaskTensor:
         mask_key = (field, size)
         if mask_key not in self._masks:
-            self._masks[mask_key] = (self._rng.random(size) < self.sample_prob).astype(
-                float
+            self._masks[mask_key] = RandomMaskTensor(
+                size, self.sample_prob, rng=self._rng, dtype=np.intp
```
willow-ahrens, line 299: Should have been different on each mask OOPS

willow-ahrens: need a different rng for each side

willow-ahrens (commented):
oops

willow-ahrens:
I checked, this formulation is actually fine. the rng is used to make a new seed on each construction

PR #658, opened 2026-09-11
Allowing queries to have tables on the lhs

Ruchika-2003:


kylebd99:
Hey, this PR is currently a few different things (Table in queries and the FusedAlias changes). It would be nice to have the table change separated out because I think it can be done independently. 

Second, I would push for having Table be the ONLY thing allow on the LHS of a query. It's awkward to always have to check which one it is, and we can always find the correct fields for the LHS by calling .fields() on the RHS. We'll also want to update the relevant interface functions like permute_dims to update the LHS table indices.

Ruchika-2003:
> Hey, this PR is currently a few different things (Table in queries and the FusedAlias changes). It would be nice to have the table change separated out because I think it can be done independently.
> 
> Second, I would push for having Table be the ONLY thing allow on the LHS of a query. It's awkward to always have to check which one it is, and we can always find the correct fields for the LHS by calling .fields() on the RHS. We'll also want to update the relevant interface functions like permute_dims to update the LHS table indices.

Should I just delete this PR, I branched it off of fused_alias since we are going to tie its usage to this

willow-ahrens:
Typically, when we have to do a bunch of dependent features, like feature A and then a feature B that depends on A, we introduce a PR that just does feature A off of main, and then we do a PR off of feature A branch, which we retarget to main once A is merged.

willow-ahrens:
@Ruchika-2003 What's the status of this PR? Should we close it for now and open a PR that's just the tables on lhs?

Ruchika-2003:
> @Ruchika-2003 What's the status of this PR? Should we close it for now and open a PR that's just the tables on lhs?

Yes Willow, I forgot to close this one. closing it now. Also I am branching the lhs table one off of main correct me if I am wrong, based on Kyle's previous message.

PR #665, opened 2026-09-14
Julia Runtime Improvements through buffer reuse

JoelMathewC:
## Performance Improvements
1. **Leasing Buffers from pool**: While previous changes (#646) allowed us to map input tensors to their corresponding julia buffers, this PR now allows us to maintain a pool of julia buffers that can be leased when tensors are being written afresh and returned to the pool when killed within the python runtime. This follows a similar idea to how CUDA buffers are pooled and reused within pytorch.
2. **Cross-kernel buffer reuse**: Previously buffers were reused only for a single kernel, this PR allows buffers to be reused across multiple kernels by maintaining a process-wide julia buffer context.

## Structural Improvements
The current julia backend implements the runtime alongside the compiler. This limits consumers of the julia backend from extending the runtime to exploit opportunities that are otherwise undecidable from looking at a single kernel. This PR introduces the following structural changes to achieve this goal
1. **FinchJLRuntime**: Abstract class for runtime.
2. **FinchJLDefaultRuntime**: Class that implements FinchJLRuntime. The runtime caches kernels and handles kernel calls. This class enforces the memory ownership model introduced in #646.
3. `FinchJLCompiler` now accepts a runtime. Schedules are updated to pass in `FinchJLDefaultRuntime`.
4. `FinchJLKernel` defers `__call__` to the runtimes `kernel_call` method giving the runtime complete control of how kernels are called.
5. Misc Changes
    - Got rid of `CompiledJLKernel` since it was a wrapper over `FinchJLKernel` without any distinctive features. 
    - Avoid using expensive `jl.isa` calls in the `kernel_call` path.


JoelMathewC:
We talked about this in the meet today. I'm going to add a sparse reuse here so that we can verify that everything works

JoelMathewC:
@willow-ahrens I removed you as a reviewer since I wanted to use this PR to test out some of the lazy graph caching stuff as well. I'll add you back once I clean stuff up. If the lazy graph caching works I'll move it into a separate PR

PR #671, opened 2026-09-18
Stats Benchmark and refactoring test_tensor_stats.py

yusheng036:
I have addressed the issues raised in the previous meeting. My PR is quite big.

First of all, I have changed the the benchmark test to test_stats_estimation.py. I have added my test as part of the test-julia test suite. The ssgetsy matrix are loaded into the /test_data suite and used for the computing of actual nnz and the estimated nnz of the kernels. I have adopted the data regression approach and generated a separate regression file for each factory. When changes are made to a specific factory, only that specific test would fail instead of the entire test. To regenerate the data regression test, you would have to add --force-regen to the run command. I have added instructions at the top of the test file.

Next, I have cut over 900 lines of code from test_tensor_stats.py. The following test are removed or combined:

Refactoring of test_tensor_stats

```
combined disjunction DC Card
combined DC_Card
combined reduce DC_Card
combined exact_stats semiring test
cut deduplication exact test
cut dc_stats_4d
cut test_lpdc_is_hashable_frozen
cut test that has no asserts
```


kylebd99:
move to tests/ and make it a file regression test

src/finch/tests/test_stats_estimation.py
```diff
@@ -0,0 +1,181 @@
...
+def actual_nnz(original_datadir):
+    with (original_datadir / "stats_actual_nnz.csv").open(newline="") as input_file:
+        return {
+            (row["matrix"], row["group"], row["kernel"]): int(row["actual_nnz"])
+            for row in csv.DictReader(input_file)
+        }
+
+
+@functools.cache
+def load(name, group):
```
kylebd99: Commit the matrices being used for the test in /test_data, then just load them from disk instead. Just make sure that they're reasonably small e.g. <=50MB.

.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
             default
             test-julia
       - name: Run Julia backend tests
-        run: pixi run test-julia -- --junit-xml=test-ubuntu-latest-Env-test-julia.xml
+        run: pixi run test-julia 0 -- --junit-xml=test-ubuntu-latest-Env-test-julia.xml
```
kylebd99: Remove this serialization, no need to slow down the whole test suite.

src/finch/tests/test_stats_estimation.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+pytestmark = pytest.mark.skipif(
+    not julia_available() or ssgetpy is None,
+    reason="Julia backend (juliacall/juliapkg) or ssgetpy not installed",
+)
+
+
+@pytest.fixture(scope="session")
+def actual_nnz(original_datadir):
```
kylebd99: This is an understandable workaround, but we should debug why computing the actual nnz is not fast. With relatively small matrices, all of these operations should be <1s.

src/finch/tests/test_tensor_stats.py
kylebd99, line 1: love clearing out some of these tests

kylebd99 (changes requested):
In general, this PR is an improvement. However, I don't think that it helps us understand why getting the actual nnz was so slow in the first place. We talked about this in the meeting, and you mentioned that you would do some debugging to figure out where the slowdown was coming from. So, let's do that debugging & see if we can simplify the test a bit (e.g. remove the csv). If there's a fundamental reason, then we can keep it this way.

src/finch/tests/test_stats_estimation.py
yusheng036: I have added 5 matrices to the /test_data suite. Their file sizes are:

```
ct20stif.mtx: 34 MB
bcsstk39.mtx: 20 MB
web-NotreDame.mtx: 19 MB
ca-HepTh.mtx: 256 KB
ca-GrQc.mtx: 140 KB
```

.github/workflows/ci.yml
yusheng036: I have removed the serilization and opted for loading the matrices from the /test_data suite.

src/finch/tests/test_stats_estimation.py
yusheng036: I have tried to find optimization for computing the actual nnz of the matrics. I performed a speedscope benchmarking test to test the compilation of the actual kernel. I realised that the problem was the count_nonzero function that was performing additional sum_duplicates and sort_indices. I have since changed to using .nnz to compute the non-zeroes instead.

src/finch/tests/test_tensor_stats.py
yusheng036: I enjoyed clearing the test too!

.github/workflows/ci.yml
willow-ahrens: Okay, but ssget uses file locking right?

PR #674, opened 2026-09-18
Refactor Autoscheduler + Galley Scheduler

### LINKED ISSUE #673: Add Galley Scheduler
Compile_JULIA_GALLEY

kylebd99:
This PR does some cleanup within autoschedule and creates a default scheduler called COMPILE_JULIA_GALLEY  which does all of the Galley stuff except for the formatting.


src/finch/tests/test_scheduler.py
```diff
@@ -815,7 +818,7 @@ def test_scheduler_e2e_sddmm(file_regression):
     )
 
     capture = LogicCapture()
-    scheduler = DefaultLogicOptimizer(
+    scheduler = DefaultLogicFactorizer(
```
JoelMathewC, line 821: Here and everywhere else - why do we want this to be `factorizer` instead of `optimizer`? I don't really have an opinion for or against, just curious.

JoelMathewC (approved):
LGTM. I have one comment around naming

kylebd99:
My thinking is that optimizer is a bit overused, and I wanted something more specific. The loop orderer and the formatter are both also optimizers, for example.

